Issue #2, opened 2015-09-27
[HttpPlatformHandler]: Forward common headers from reverse proxy


Tratcher:
_From @davidfowl on September 11, 2015 16:11_

This bug is to track the work that needs to happen so that items forwarded from httpplatformhandler show up in the user's application.

There are several parts to this:

- Windows Auth (coming soon)
- Client certs (not yet available)
- Https scheme

We could do some of it in middleware (like WindowsAuth) but that would mean it wouldn't work out of the box. It could also be in the template by default so that we don't couple to the HttpPlatformHandler in any of our core components.

_Copied from original issue: aspnet/Hosting#349_

Tratcher:
Also of interest are the remote ip & port and the reverse proxy public IP & port. I'd recommend doing it all in middleware.

Eventually we'll need the TLS token binding.

Tratcher:
_From @davidfowl on September 11, 2015 16:41_

I think scheme is the only one we'd want to support by default.

Tratcher:
HttpPlatformHandler needs to forward them all, it's up to the app how many forwarders they check.

Tratcher:
_From @davidfowl on September 11, 2015 17:13_

We'll let the user experience drive that. This is our new ootb experience for IIS and it can't be degraded in obvious ways. We'll discuss it in the design meeting

Tratcher:
A few other interop questions:
100-continue support
http/2 termination (should just work, it happens in http.sys)
client disconnects

Tratcher:
_From @davidfowl on September 15, 2015 3:43_

We decided to make this a separate middleware `UseHttpPlatformHandler()`. The middleware may reuse the forwarding middleware as the Http platform handler uses some of the standard headers `X-Forward-For` and `X-Forward-Proto`.

This middleware also needs to handle windows auth via the `X-IIS-WindowsAuthToken` header. The value is a hex value that represents the primary user token (see https://msdn.microsoft.com/en-us/library/ms692659(v=vs.90).aspx). We have to close the handle on the way out since it's duped per request. The token is represented in the header by a hex value.

Tratcher:
_From @rubenprins on September 16, 2015 17:39_

What about nested applications/child paths? Is that still something that will be supported out of the box? Or will we need to configure the application's base path (a.k.a. `~/`) manually?

Tratcher:
_From @davidfowl on September 16, 2015 17:51_

[...]

Issue #1, opened 2015-09-26
Implement a full authentication handler


Tratcher:
To enable mixed anonymous and windows auth the PlatformHandlerMiddleware must implement the IAuthenticationHandler contracts for Authenticate, Challenge, and Forbidden. Open questions:
- Should it be 'automatic', always processing the incoming identity and always responding to Challenges? This may need to be configurable.
- How can it know what auth is enabled in IIS? This may need to be configured too.
- When there is a WindowsIdentity, how can it tell what type of auth was used to put in the AuthScheme field? This will also be needed for converting challenges to forbidden.

PR #3, opened 2015-09-27
Platform handler middleware, sample, tests, functional tests.

Tratcher:
#2 @davidfowl @blowdart @muratg 
For the record, I'm not fond of the middleware/extension name UsePlatformHandler. I'd rather call it UseIISForwarders. Same for the package name, AspNet.IISForwarders, or just AspNet.IIS.

Only simple WindowsAuth is supported, advanced scenarios will be implemented via #1.


Tratcher:
Note that everything works find forwarding to Kestrel. I was able to make WebListener work is full IIS, but not in IIS Express for some reason. Will debug later.

davidfowl:
@Tratcher is the build and ci all setup for this repo?

Tratcher:
No. And we need to get the msi installed on the agents.

src/Microsoft.AspNet.PlatformHandler/HttpPlatformHandlerMiddleware.cs
```diff
@@ -0,0 +1,85 @@
...
+            int hexHandle;
+            if (!StringValues.IsNullOrEmpty(xIISWindowsAuthToken)
+                && int.TryParse(xIISWindowsAuthToken, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out hexHandle))
+            {
+                var handle = new IntPtr(hexHandle);
+                var winIdentity = new WindowsIdentity(handle);
+                // WindowsIdentity just duplicated the handle so we need to close the original.
+                NativeMethods.CloseHandle(handle);
+
+                httpContext.Response.RegisterForDispose(winIdentity);
```
davidfowl: Why do this instead of handle it on the way out (via try finally)? It means that middleware registered before this will see the WindowsIdentity on the way out. Isn't that weird?

src/Microsoft.AspNet.PlatformHandler/HttpPlatformHandlerMiddleware.cs
```diff
@@ -0,0 +1,85 @@
...
+                var winPrincipal = new WindowsPrincipal(winIdentity);
+
+                var existingPrincipal = httpContext.User;
+                if (existingPrincipal != null)
+                {
+                    httpContext.User = SecurityHelper.MergeUserPrincipal(existingPrincipal, winPrincipal);
```
davidfowl: I believe this thing handles nulls? https://github.com/aspnet/Common/blob/dev/src/Microsoft.Framework.SecurityHelper.Sources/SecurityHelper.cs#L21

So no need to check for null. Unless you're trying to avoid the ClaimsPrincipal wrapping the WindowsIdentity?

src/Microsoft.AspNet.PlatformHandler/HttpPlatformHandlerMiddleware.cs
Tratcher: We want it to be valid as long as possible so it can be used in odd places like OnStarting. None of our components attempt to remove HttpContext.User.

src/Microsoft.AspNet.PlatformHandler/HttpPlatformHandlerMiddleware.cs
Tratcher: Right, there's no reason for it to take the WindowsPrincipal and replace it with a ClaimsPrincipal.

Tratcher:
Updated. Renamed to IISPlatformHandler

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HelloWorldTest.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNet.Server.Testing;
+using Microsoft.AspNet.Testing.xunit;
+using Microsoft.Framework.Logging;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.AspNet.Loader.IIS.FunctionalTests
+{
+    // Uses ports ranging 5061 - 5069.
```
troydai: I'm sure this had been discussed before. But shall we find the free port in a more dynamic fashion?
https://github.com/aspnet/dnx/blob/dev/test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/Infrastructure/DthTestServer.cs#L135

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HelloWorldTest.cs
```diff
@@ -0,0 +1,94 @@
...
+    public class HelloWorldTests
+    {
+        [ConditionalTheory]
+        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
+        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x86, "http://localhost:5061/", ServerType.Kestrel)]
+        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, "http://localhost:5062/", ServerType.Kestrel)]
+        [InlineData(RuntimeFlavor.CoreClr, RuntimeArchitecture.x86, "http://localhost:5063/", ServerType.Kestrel)]
+        [InlineData(RuntimeFlavor.CoreClr, RuntimeArchitecture.x64, "http://localhost:5064/", ServerType.Kestrel)]
+        public Task HelloWorld_IISExpress(RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl, ServerType delegateServer)
```
troydai: You may want to use the new test infrastructure to iterate through the dnx flavors.

troydai: Non-blocking.

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HelloWorldTest.cs
troydai: Non-blocking.

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HelloWorldTest.cs
```diff
@@ -0,0 +1,94 @@
...
+        public Task HelloWorld_IIS(RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl, ServerType delegateServer)
+        {
+            return HelloWorld(ServerType.IIS, runtimeFlavor, architecture, applicationBaseUrl, delegateServer);
+        }
+
+        public async Task HelloWorld(ServerType serverType, RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl, ServerType delegateServer, bool noSource = false)
+        {
+            var logger = new LoggerFactory()
+                            .AddConsole()
+                            .CreateLogger(string.Format("HelloWorld:{0}:{1}:{2}:{3}", serverType, runtimeFlavor, architecture, delegateServer));
```
troydai: string interploation

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HelloWorldTest.cs
troydai: Just realize that this is not at the sample repo as __DNX__ so the test infrastructure is out of reach.

It would be worth to consider to pull our test infrastructure to a independent package so as to be consumed across repo. @davidfowl @JunTaoLuo

JunTaoLuo: Potentially https://github.com/aspnet/Testing?

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HelloWorldTest.cs
Tratcher: https://github.com/aspnet/Hosting/issues/378

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HttpsTest.cs
```diff
@@ -0,0 +1,186 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNet.Server.Testing;
+using Microsoft.AspNet.Testing.xunit;
+using Microsoft.Framework.Logging;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.AspNet.Loader.IIS.FunctionalTests
+{
+    // IisExpress preregisteres 44300-44399 ports.
```
troydai: Same as above. Non-blocking though,

Tratcher:
Updated.

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HttpsTest.cs
```diff
@@ -0,0 +1,186 @@
...
+        [ConditionalTheory]
+        [OSSkipCondition(OperatingSystems.MacOSX | OperatingSystems.Linux)]
+        [InlineData(RuntimeFlavor.CoreClr, RuntimeArchitecture.x86, "https://localhost:44399/")]
+        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, "https://localhost:44398/")]
+        public Task Https_HelloWorld(RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl)
+        {
+            return HttpsHelloWorld(ServerType.IISExpress, runtimeFlavor, architecture, applicationBaseUrl);
+        }
+
+        public async Task HttpsHelloWorld(ServerType serverType, RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl)
```
troydai: Why extract this? It is used in only one test method.

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HttpsTest.cs
```diff
@@ -0,0 +1,186 @@
...
+        public Task Https_HelloWorld_ClientCert(RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl)
+        {
+            return HttpsHelloWorldCerts(ServerType.IISExpress, runtimeFlavor, architecture, applicationBaseUrl, sendClientCert: true);
+        }
+
+        public async Task HttpsHelloWorldCerts(ServerType serverType, RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl, bool sendClientCert)
+        {
+            var logger = new LoggerFactory()
+                            .AddConsole()
+                            .CreateLogger(string.Format("HttpsHelloWorld:{0}:{1}:{2}", serverType, runtimeFlavor, architecture));
```
troydai: string interpolation

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/NtlmAuthentationTest.cs
```diff
@@ -0,0 +1,115 @@
...
+
+                        response = await httpClient.GetAsync("/RestrictedNTLM");
+                        Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+                        Assert.Contains("NTLM", response.Headers.WwwAuthenticate.ToString());
+                        // Note we can't restrict a challenge to a specific auth type, the native auth modules always add themselves.
+                        Assert.Contains("Negotiate", response.Headers.WwwAuthenticate.ToString());
+
+                        response = await httpClient.GetAsync("/Forbidden");
+                        Assert.Equal(HttpStatusCode.Forbidden, response.StatusCode);
+                        */
```
troydai: Any reason to keep these commented code?

test/Microsoft.AspNet.PlatformHandler.FunctionalTests/HttpsTest.cs
Tratcher: Consistency with the other tests, and if we want to add more tests.

troydai:
:shipit:

PR #13, opened 2015-10-12
Implement a full authentication handler.

Tratcher:
#1 @HaoK @davidfowl @muratg 
This integrates IISPlatformHandlerMiddleware with the authentication infrastructure to:
1. Authenticate the user and set HttpContext.User, as well as respond to Authenticate calls.
2. Accept Challenges and set 401s. We don't set the headers because IIS will.

Note there's no way for us to detect what auth is actually enabled in IIS so we assume it's Negotiate and NTLM by default.

### DIFF
--- a/src/Microsoft.AspNet.IISPlatformHandler/AuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.IISPlatformHandler/AuthenticationHandler.cs
@@ -1,0 +1,138 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features.Authentication;
+
+namespace Microsoft.AspNet.IISPlatformHandler
+{
+    internal class AuthenticationHandler : IAuthenticationHandler
+    {
+        internal AuthenticationHandler(HttpContext httpContext, IISPlatformHandlerOptions options, ClaimsPrincipal user)
+        {
+            HttpContext = httpContext;
+            User = user;
+            Options = options;
+        }
+
+        internal HttpContext HttpContext { get; }
+
+        internal IISPlatformHandlerOptions Options { get; }
+
+        internal ClaimsPrincipal User { get; }
+
+        internal IAuthenticationHandler PriorHandler { get; set; }
+
+        public Task AuthenticateAsync(AuthenticateContext context)
+        {
+            if (ShouldHandleScheme(context.AuthenticationScheme))
+            {
+                if (User != null)
+                {
+                    context.Authenticated(User, properties: null,
+                        description: Options.AuthenticationDescriptions.Where(descrip =>
+                            string.Equals(User.Identity.AuthenticationType, descrip.AuthenticationScheme, StringComparison.Ordinal)).FirstOrDefault()?.Items);
+                }
+                else
+                {
+                    context.NotAuthenticated();
+                }
+            }
+
+            if (PriorHandler != null)
+            {
+                return PriorHandler.AuthenticateAsync(context);
+            }
+            return Task.FromResult(0);
+        }
+
+        public Task ChallengeAsync(ChallengeContext context)
+        {
+            bool handled = false;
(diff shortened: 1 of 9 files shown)


blowdart:
No way to query the metabase? I'm a bit concerned about the default choice, basic is also valid.

Tratcher:
Not on coreclr. Basic is valid, but not usually used with windows auth. Users can always add it on the options.

src/Microsoft.AspNet.IISPlatformHandler/IISPlatformHandlerMiddleware.cs
```diff
@@ -57,29 +72,77 @@ public Task Invoke(HttpContext httpContext)
...
             int hexHandle;
+            WindowsPrincipal winPrincipal = null;
             if (!StringValues.IsNullOrEmpty(xIISWindowsAuthToken)
                 && int.TryParse(xIISWindowsAuthToken, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out hexHandle))
             {
+                // Always create the identity if the handle exists, we need to dispose it so it does not leak.
                 var handle = new IntPtr(hexHandle);
-                var winIdentity = new WindowsIdentity(handle);
+
+                string authType = _options.AuthenticationDescriptions.First().AuthenticationScheme;
```
davidfowl: Why is this ok?

davidfowl: Can't this be cleared by the user?

Tratcher:
Updated.

davidfowl:
Why are there merge conflicts

src/Microsoft.AspNet.IISPlatformHandler/IISPlatformHandlerMiddleware.cs
```diff
@@ -57,29 +72,66 @@ public Task Invoke(HttpContext httpContext)
...
+            {
+                await _next(httpContext);
+            }
+            finally
+            {
+                DetachAuthenticationhandler(handler);
+            }
+        }
+
+        private void HookUpAuthenticationHandler(AuthenticationHandler handler)
```
davidfowl: Attach/Detach - Hookup must be a polycheck violation

Tratcher:
Rebased, updated.

src/Microsoft.AspNet.IISPlatformHandler/AuthenticationHandler.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        internal IAuthenticationHandler PriorHandler { get; set; }
+
+        public Task AuthenticateAsync(AuthenticateContext context)
+        {
+            if (ShouldHandleScheme(context.AuthenticationScheme))
+            {
+                if (User != null)
+                {
+                    context.Authenticated(User, null,
```
davidfowl: What's the null parameter? Use optional parameter syntax there

src/Microsoft.AspNet.IISPlatformHandler/IISPlatformHandlerMiddleware.cs
```diff
@@ -22,13 +27,44 @@ public class IISPlatformHandlerMiddleware
...
+                throw new ArgumentNullException(nameof(options));
+            }
             _next = next;
+            _options = options;
         }
 
-        public Task Invoke(HttpContext httpContext)
+        public async Task Invoke(HttpContext httpContext)
+        {
+            CheckScheme(httpContext);
```
davidfowl: Why Check? UpdateScheme, UpdateRemoteIp? TryUpdate. This operation potentially mutates the HttpContext so Check is a misnomer

Tratcher:
Updated

HaoK:
Looks reasonable to me, I always have to ask, is there anything we can push to the base AuthHandler stuff to make it a candidate as the base class for this to eliminate some of the code duplication?  Since any future changes we make to the base stuff will need to be manually ported here...

davidfowl:
@HaoK is the auth handler pretty stand alone? It would be a shame if this pulled in a bunch of dependencies.

davidfowl:
Yeahhhh https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/project.json#L11-L18

Tratcher:
Discussed offline. This one uses the same patterns but has some key differences such as multiple auth types. There's no plan to re-use the security components here.

davidfowl:
:shipit:

PR #23, opened 2015-10-27
Lower integration test timeouts.

Tratcher:
Under some failure conditions test requests can take 100s to timeout, and then they are retried 60 times for a total of 1.3 hours per test failure. This can cause the entire test pass to time out. Lowering this to a maximum of 5s per request, and 30 retires, for a maximum of 2.5 minutes per test failure.

@muratg


muratg:
:shipit:

Issue #17, opened 2015-10-16
HttpPlatformHandler/Kestrel: HttpContext.Connection.RemoteIpAddress not working


GuardRex:
Following: https://github.com/aspnet/IISIntegration/issues/16

I'm using Google RECAPTCHA on this contact form, and I think I used to have ...
```
string gIp = Context.Connection.RemoteIpAddress.ToString();
```
... in the beta7 version of this app. For the beta8 upgrade, that `Context` went away, but it seemed cool with `HttpContext` ...
```
string gIp = HttpContext.Connection.RemoteIpAddress.ToString();

```
... but that's throwing a null reference exception. Is this a "Luke not handling the contexts correctly" problem or a "HttpPlatformHandler/Kestrel" problem?

Tratcher:
` app.UseIISPlatformHandler();` should set RemoteIpAddress.

GuardRex:
I have the `app.UseIISPlatformHandler();` in the app with the package in place. Is there anything else I rundown on this for you?

Tratcher:
You could show your startup class.

GuardRex:
```c#
namespace MyApp
{
    public class Startup
    {
        public IConfiguration Configuration { get; set; }

        public Startup(IHostingEnvironment env)
        {
            var configurationBuilder = new ConfigurationBuilder().AddEnvironmentVariables();
            Configuration = configurationBuilder.Build();
            if(machineName.StartsWith("~~~~~~", StringComparison.OrdinalIgnoreCase))
            {
                Configuration["ASPNET_ENV"] = "Production";
                ... other config stuff here ...
            }
            else if (machineName.StartsWith("~~~~~~", StringComparison.OrdinalIgnoreCase))
            {
                Configuration["ASPNET_ENV"] = "Production";
                ... other config stuff here ...
            }
            else
            {
                Configuration["ASPNET_ENV"] = "Development";
            }
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddAntiforgery();
            services.AddSingleton(_ => Configuration);
            services.AddMvc(options => 
            {
                string machineName = Environment.GetEnvironmentVariable("COMPUTERNAME");
                if (machineName.StartsWith("~~~~~~", StringComparison.OrdinalIgnoreCase))
                {
                    options.CacheProfiles.Add("PublicCache", new CacheProfile()
                    {
                        Duration = 2592000,
                        Location = ResponseCacheLocation.Any
                    });
                }
                else
                {
[...]

PR #37, opened 2015-11-17
Split the IPAddress and Port

JonghoL:


davidfowl:
Add some tests

src/Microsoft.AspNet.IISPlatformHandler/IISPlatformHandlerMiddleware.cs
```diff
@@ -91,6 +93,11 @@ private static void UpdateRemoteIp(HttpContext httpContext)
                         httpContext.Request.Headers[XOriginalIPName] = remoteIPString;
                     }
                     httpContext.Connection.RemoteIpAddress = ipFromHeader;
+                    int remotePortFromHeader;
+                    if (int.TryParse(ipString.Substring(portStart+1, ipString.Length - (portStart+1)), out remotePortFromHeader))
```
davidfowl: Split this into multiple lines

Tratcher:
Test input:
`127.0.0.1`
`127.0.0.1:80` - Currently broken
`::1` - Regressed by this PR.
`[::1]` - Regressed by this PR.
`[::1]:80` - Currently broken

pakrym:
Fixed in https://github.com/aspnet/IISIntegration/pull/38/files

GuardRex:
@pakrym This might be the wrong spot to ask (I'll move it back to https://github.com/aspnet/IISIntegration/issues/17 if that's best):

1. Forwarded is the RFC now (https://tools.ietf.org/html/rfc7239), that's coming at some point down the road?
2. This all looks to be IPv4, what's the story on IPv6 support? Does the `TryParse` handle it ok?
3. The headers list ... I mean I found a lot more when I looked: X-FORWARDED-FOR, REMOTE_ADDR, HTTP_X_FORWARDED_FOR, HTTP_CLIENT_IP, HTTP_X_FORWARDED, "HTTP_X_CLUSTER_CLIENT_IP, HTTP_FORWARDED_FOR, HTTP_FORWARDED, X_FORWARDED_FOR, CLIENT_IP, X_FORWARDED, X_CLUSTER_CLIENT_IP, FORWARDED_FOR, FORWARDED. I was just wondering if some of these other non-standard headers that we might come up against in the wild should be handed by us ... not by the package?

Tratcher:
@GuardRex For 1 and 3, this middleware only handles headers set by the HttpPlatformHandler. I've asked them about 1 and they said they'd look into it.
For 2, this was PR replaced by https://github.com/aspnet/IISIntegration/pull/38 where IPv6 works fine.

GuardRex:
@Tratcher Gotcha. I saw @pakrym closed https://github.com/aspnet/IISIntegration/issues/17, so that's why I asked.

For #3 though, we should handle funky non-standard headers for IP ourselves. I just want to make sure I know what I can count on with `Connection.RemoteIpAddress` and `Connection.RemotePort`.

PR #43, opened 2015-12-04
Remove x-fowarded-for and x-forwarded-proto.

Tratcher:
@muratg @blowdart @davidfowl 
We're going to consolidate this logic over in https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs and make it more configurable.


davidfowl:
What?

davidfowl:
@Tratcher Do you have a bug on the Templates repository to add the forwarding middleware?

Tratcher:
https://github.com/aspnet/Templates/issues/343

muratg:
:shipit:

GuardRex:
https://tools.ietf.org/html/rfc7239 ... You took a look at this earlier, but nothing came back "yeah" or "nay" ... just curious what the outcome of the internal discussion was.

Tratcher:
@GuardRex HttpPlatformHandler work will pick up again in Jan.

Issue #42, opened 2015-12-02
Remove caption from NTLM so it doesn't show up on the sign in page


davidfowl:

BrennanConroy:
41cd7bf45ecbe5c310b8f627d987c0bb913f692b

PR #48, opened 2015-12-10
Remove display name for Negotiate and Ntlm

BrennanConroy:
#42 
@Tratcher 

Have 6 tests failing locally, but they were failing before this too.


Tratcher:
:shipit:

Issue #33, opened 2015-11-11
HttpPlatformHandler forwards the path un-escaped, breaks for `?`


Tratcher:
When HttpPlatformHandler forwards the request path and query, it uses the un-escaped path and the original escaped query. This causes a few issues:

1. Query - If there is an escaped `?` (%3F) in the original path it is un-escaped and forwarded. The back-end server (kestrel, weblistener) then interprets this as the query separator, breaking both path and query. There is no workaround.
2. Backslash - %5C is unescaped to `\` and then converted to `/`, confusing routing
3. Double un-escaping - This would be a problem if you sent `/%25%34%31`, except IIS's request filtering logic blocks %25 by default.

Tratcher:
FYI @davidfowl @halter73

Tratcher:
Fixed by the next version of HttpPlatformHandler (v1.3?)

Tratcher:
This as been fixed in AspNetCoreModule v0.8: https://github.com/aspnet/IISIntegration/issues/105.

PR #59, opened 2016-01-11
Add question mark escaping test

pakrym:
@Tratcher


Tratcher:
:+1: Too bad we can't check it in for a few weeks.
https://github.com/aspnet/IISIntegration/issues/33

Tratcher:
Merged with #96

Issue #52, opened 2016-01-07
Add UseIISPlatformHandler extension method to IWebApplicationBuilder


davidfowl:
See https://github.com/aspnet/Hosting/pull/553. Here's what the entry point would look like:

```C#
using System.IO;
using Microsoft.AspNet.Hosting;

namespace HelloMvc
{
    public class Program
    {
        public static void Main(string[] args)
        {
            var hostingConfiguration = WebApplicationConfiguration.GetDefault(args);
            
            var application = new WebApplicationBuilder()
                        .UseApplicationBasePath(Directory.GetCurrentDirectory())
                        .UseServerFactory("Microsoft.AspNet.Server.Kestrel")
                        .UseConfiguration(hostingConfiguration)
                        .UseUrls("http://localhost:5000")
                        .UseIISPlatformHandler()
                        .UseStartup<Startup>()
                        .Build();

            application.Run();
        }
    }
}
```

UseIISPlatformHandler sets the appropriate server.urls entry for the port and base path. These are baked into hosting today.

glennc:
I assume this is part of the outcome from yesterdays hours of discussion?

davidfowl:
Yap

Tratcher:
If you call it UseIISPlatformHandler for both IWebApplicationBuilder and IApplicationBuilder it's going to be very confusing to document. How about UseIISPlatformHandlerUrl?

GuardRex:
@Tratcher Is it safe to remain on a current runtime (`-rc2-16357` at MyGet nightly) for this change with `Hosting` package(s) declared `-rc2-*`? In addition to the extension in `Main`, what needs to be done on the runtime/packaging dependencies side (if anything)?

Tratcher:
Microsoft.AspNet.Hosting rc2-16219 and later. Microsoft.AspNet.IISIntegration rc2-16132 and later. You'll need to update both packages at the same time as there is a conflict between the new and old behavior.
It's hard to say what build of DNX works with specific versions of other packages without testing.

GuardRex:
@Tratcher Thx. Apparently I've been ok with runtime `dnx-coreclr-win-x64.1.0.0-rc2-16343`, but I'll upgrade that later today. I've always had my packages set to `-rc2-*`, so I'm :cake: on that.

PR #60, opened 2016-01-12
Added extension method to handle the HTTP_PLATFORM_PORT

davidfowl:
/cc @JunTaoLuo @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.IISPlatformHandler/IISWebApplicationBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.IISPlatformHandler/IISWebApplicationBuilderExtensions.cs
@@ -1,0 +1,19 @@
+﻿using System;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public static class IISWebApplicationBuilderExtensions
+    {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.IISPlatformHandler/IISWebApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public static class IISWebApplicationBuilderExtensions
+    {
+        public static IWebApplicationBuilder UseIISPlatformHandler(this IWebApplicationBuilder webApplicationBuilder)
```
Tratcher: IWebApplicationBuilder.UseIISPlatformHandler needs to be searchably distinct from IApplicationBuilder.UseIISPlatformHandler. UseIISPlatformHandlerUrl?

Tratcher:
I was also testing this code path yesterday and found that it fails unless we first remove the same logic from Hosting. Otherwise the address is double registered and kestrel crashes.

Tratcher:
https://github.com/aspnet/IISIntegration/issues/52

davidfowl:
Closing

davidfowl:
:shipit:

PR #62, opened 2016-01-12
Microsoft.AspNet.Tools.PublishIIS -> dotnet-publish-iis

moozzyk:
Renaming Microsoft.AspNet.Tools.PublishIIS to dotnet-publish-iis as this is the convention dotnet tools must follow


moozzyk:
@davidfowl - cli seems to be using  `"NETStandard.Library": "1.0.0-rc2-23704",` I used this version in my original PR but was told not to use exact versions and therefore I am using "NETStandard.Library": "1.0.0-*". (I also think Coherence verifies if a project is using non `x.0.0-*` versions) 

I guess the question is - do we want to constantly update this project so that we are in sync with cli or only if things break because of incompatibilities? Or is there a different concern here I am not aware of? 

/cc @Tratcher

Tratcher:
:shipit:

PR #66, opened 2016-01-13
Rename windows auth forwarder header to MS-PLATFORM-HANDLER-WINAUTHTOKEN.

Tratcher:
This is changing with the next version of HttpPlatformHandler so all the custome headers use the same naming scheme.

Do not merge yet or the tests will fail.
@davidfowl @pakrym

### DIFF
--- a/src/Microsoft.AspNet.IISPlatformHandler/IISPlatformHandlerMiddleware.cs
+++ b/src/Microsoft.AspNet.IISPlatformHandler/IISPlatformHandlerMiddleware.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Globalization;
-using System.Security.Cryptography.X509Certificates;
 using System.Security.Principal;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.IISPlatformHandler/IISPlatformHandlerMiddleware.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.AspNet.IISPlatformHandler
 {
     public class IISPlatformHandlerMiddleware
     {
-        private const string XIISWindowsAuthToken = "X-IIS-WindowsAuthToken";
+        private const string MSPlatformHandlerWinAuthTOken = "MS-PLATFORM-HANDLER-WINAUTHTOKEN";
```
davidfowl: Need back compat or we'll break everyone at the same time. Read both headers for now.

Tratcher:
Updated with temporary back compat. This can now be merged without updating HttpPlatformHandler.

davidfowl:
:shipit:

PR #68, opened 2016-01-14
dotnet-publish-iis refactoring

moozzyk:
- moving command logic to a dedicated class
- adding logic for reading webroot value from a parameter, hosting.json file and folder structure
- reading application name from project.json instead of a parameter

### DIFF
--- a/src/dotnet-publish-iis/Program.cs
+++ b/src/dotnet-publish-iis/Program.cs
@@ -2,9 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.IO;
-using System.Xml;
-using System.Xml.Linq;
 using Microsoft.Extensions.CommandLineUtils;
 
 namespace Microsoft.AspNet.Tools.PublishIIS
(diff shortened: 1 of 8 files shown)


NuGet.config
```diff
@@ -3,5 +3,7 @@
   <packageSources>
     <add key="AspNetVNext" value="https://www.myget.org/F/aspnetcidev/api/v2" />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="dotnet-cli" value="https://www.myget.org/F/dotnet-cli/api/v3/index.json" />
```
Tratcher: I'm pretty sure we don't want these feeds here. @pranavkm?

src/dotnet-publish-iis/Program.cs
```diff
@@ -15,46 +12,27 @@ public static int Main(string[] args)
...
+                Name = "dotnet publish-iis",
                 FullName = "Asp.Net IIS Publisher",
                 Description = "IIS Publisher for the Asp.Net web applications",
             };
             app.HelpOption("-h|--help");
 
-            var publishFolderOption = app.Option("--publish-folder", "The path to the publish output folder", CommandOptionType.SingleValue);
-            var appNameOption = app.Option("--application-name", "The name of the application", CommandOptionType.SingleValue);
+            var publishFolderOption = app.Option("--publish-folder|-p", "The path to the publish output folder", CommandOptionType.SingleValue);
+            var webRootOption = app.Option("--webroot|-w", "The name of webroot folder", CommandOptionType.SingleValue);
```
Tratcher: If not the default wwwroot.

NuGet.config
pranavkm: Yeah - it messes up the gating.

moozzyk:
I need to double check this - yesterday I was not able to build even dev branch without adding this feed. Maybe it was just a temporary problem?

src/dotnet-publish-iis/Program.cs
moozzyk: Not necessarily - webroot can also be read from hosting.json

NuGet.config
moozzyk: @pranavkm - not sure what it means apart from "remove this line"

Tratcher: Remove both of these lines. The CI pulls these packages and pushes them to our own feeds.

Tratcher:
The CoreClr and CLI builds have been broken on and off for weeks. That's why we have the CI pull them in and validate them.

NuGet.config
pranavkm: :+1:

moozzyk: I may not be able to - I added `dotnet-core` to be able to get `"Microsoft.DotNet.ProjectModel": "1.0.0-*"`

moozzyk:
Looks like today things work without additional feeds so I removed them from the NuGet.config file.

NuGet.config
moozzyk: Removed both entries

Tratcher:
:shipit:

src/dotnet-publish-iis/Program.cs
moozzyk: @Tratcher : how about: 

The name of webroot folder. Overwrites webroot from hosting.json file or the default "wwwroot"

?

Tratcher: To wordy for the command line. You can leave it as is.

moozzyk: Yes. It's hard to something both explanatory and short. will leave as is then.

PR #72, opened 2016-01-21
Avoid merging with anonymous users.

Tratcher:
https://github.com/aspnet/Common/issues/85
@blowdart @muratg 
This avoids the issue most of the time. We'll still want to make principal merging smarter at some point.

### DIFF
--- a/src/Microsoft.AspNet.IISPlatformHandler/IISPlatformHandlerMiddleware.cs
+++ b/src/Microsoft.AspNet.IISPlatformHandler/IISPlatformHandlerMiddleware.cs
@@ -105,7 +105,8 @@ namespace Microsoft.AspNet.IISPlatformHandler
 
                 if (_options.AutomaticAuthentication)
                 {
-                    var existingPrincipal = httpContext.User;
+                    // Don't get it from httpContext.User, that always returns a non-null anonymous user by default.
+                    var existingPrincipal = httpContext.Features.Get<IHttpAuthenticationFeature>()?.User;
                     if (existingPrincipal != null)
                     {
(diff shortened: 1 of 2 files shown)


muratg:
:shipit:

mrsheepuk:
Hi guys - quick question on this - I'm using the code below to set up some policies to authorise users against membership in certain AD groups - I'm finding that it's only working if I use SIDs as the group names when hosting in IIS - Kestrel (it works perfectly using weblistener outside IIS) on RC1-Update1 - am I right in thinking this change will address the problem I'm seeing, such that this should work against group names correctly in RC2? 

Is using the SIDs a sensible, safe workaround until RC2 is released?

Thanks!!

```
var viewerGroups = Configuration.GetSection("InternalGroups:Viewer").Value.Split(',');
var adminGroups = Configuration.GetSection("InternalGroups:Admin").Value.Split(',');
var namedGroups = Configuration.GetSection("InternalGroups:Named").Value.Split(',');
services.AddAuthorization(auth =>
{
    auth.AddPolicy("read", policy =>
    {
        policy.RequireRole(viewerGroups);
    });
    auth.AddPolicy("admin", policy =>
    {
        policy.RequireRole(adminGroups);
    });
    auth.AddPolicy("named", policy =>
    {
        policy.RequireRole(namedGroups);
    });
});
```

Tratcher:
Yes this change will probably address your scenario. In the meantime using the sids should be ok.

mrsheepuk:
Thanks @Tratcher - I'll check using names again once the next RC comes out.

Issue #67, opened 2016-01-14
Remove X-IIS-WindowsAuthToken


Tratcher:
This changed between HttpPlatformHandler v1.2 and 1.3 form X-IIS-WindowsAuthToken to MS-PLATFORM-HANDLER-WINAUTHTOKEN. Remove the old one after RC2.
https://github.com/aspnet/IISIntegration/pull/66

muratg:
Why RTW?

Tratcher:
Compat bridge. Otherwise you have to update HttpPlatformHandler and AspNet in lockstep.

muratg:
Moving back to RC2 because of the recent schedule changes. If needed, we'll move it out again.

Tratcher:
Merged with #96

PR #77, opened 2016-01-25
Skip IISPlatformHandlerMiddleware if the id token is not present.

Tratcher:
Requires the next version of HttpPlatformHandler (v1.3?).
@blowdart @muratg 

We won't commit this until after RC2 because it will prevent the use of HttpPlatformHandler v1.2. This allows the full RC2-RTW window for hosters to update HttpPlatformHandler. RTW will have a hard dependency on 1.3.

RE: https://github.com/aspnet/IISIntegration/issues/67


muratg:
Moving back to RC2 due to the schedule changes. I hope we can get HPH 1.3 out before this.

Tratcher:
Merged with #96

PR #78, opened 2016-01-25
README.md: ASP.NET 5 -> ASP.NET Core 1.0

MaximRouiller:
Naming change.


Eilon:
We will be doing these updates in bulk, but thank you.

PR #80, opened 2016-02-03
Adding ASPNET_APPLICATIONBASE environment variable to fix MVC views in IIS

moozzyk:


GuardRex:
Curious ... this is required because the path can't be specified to `WebHostBuilder`? I'm using the environment variable now, but I want to be able to drop it. This PR implies that I'll be stuck with it.

moozzyk:
No. To avoid copying views each time they change you would do this: https://github.com/aspnet/cli-samples/blob/master/HelloMvc/Program.cs#L13. But then if you start the application from a different folder (and http platform handler always starts the app from wwwroot) the published app won't be able to find the views. Resetting application base using an environment variable will fix this because it will overwrite the folder set by `UseApplicationBasePath`

GuardRex:
Why can't `UseApplicationBasePath` set the path [even if you put a string in it (not ideal ... I'm just asking)] that points to the true base path and thus eliminate the need for this environment variable? (Am I confused by your comment "copy the views each time they change?")

This is prob the wrong place to discuss this. Shouldn't I just ask on JabbR?

moozzyk:
What string would you put though? The string needs to be different between the app during development and after publishing. Also being able to overwrite a hardcoded setting with an env variable seems to be useful. 
"copy the views each time they change" relates to a development scenario where your Views are in the project root but the application is compiled to a bin/Debug/dnxcore50/... folder. Without setting the app base to the project root the app will look for views in bin/Debug/dnxcore50/. So, if the views are not copied the app won't find them. Even if the views were copied on build changing them without rebuilding the application would have no effect because the app would look for the views in the bin/Debug/...

(I haven't setup my [JabbR](https://github.com/moozzyk/JabbrClientCpp) client :smile: on this box yet )

GuardRex:
@moozzyk Good points.

Environment variables ... not a big fan. I just want the app to :sparkles: magically :sparkles: know its base path without any sort of special settings (or env vars), and I know what I'll do to take care of this little (personal) problem of mine now.

I'm good now. Thanks for your help.

muratg:
@pakrym Could you do a PR?

src/dotnet-publish-iis/WebConfigTransform.cs
```diff
@@ -64,6 +64,25 @@ private static void TransformHttpPlatform(XElement httpPlatformElement, string a
...
+
+        private static void AddApplicationBase(XElement httpPlatformElement)
+        {
+            const string appBaseKeyName = "ASPNET_APPLICATIONBASE";
+
+            var envVariables = GetOrCreateChild(httpPlatformElement, "environmentVariables");
+            var appBaseElement = envVariables.Elements("environmentVariable").SingleOrDefault(e =>
+                string.Equals((string)e.Attribute("name"), appBaseKeyName, StringComparison.CurrentCultureIgnoreCase));
+
+            if (appBaseElement == null)
```
pakrym: Test for case where `ASPNET_APPLICATIONBASE` already exists?

pakrym:
:ship: :it:

src/dotnet-publish-iis/WebConfigTransform.cs
moozzyk: This code is only modifying the web.config file - so there is a test verifying that the config is modified as [expected](https://github.com/aspnet/IISIntegration/pull/80/files#diff-2d39602621e11916e8439e3e56c8b466R17). ASPNET_APPLICATIONBASE is read somewhere in Hosting so I guess a testcase if any could be there (not sure if we really want to modify env variables during testing). Finally, the real E2E scenario for this involves dotnet publish, IIS, httpplatformhandler and hosting - I don't think we have any test like this but anytime you publish to iis you will invoke this code..

pakrym: I meant test that verifies case where web.config you are modifying already has ASPNET_APPLICATIONBASE and you shouldn't not change it.

moozzyk: Yes https://github.com/aspnet/IISIntegration/pull/80/files#diff-2d39602621e11916e8439e3e56c8b466R180. Kind of... I just noticed it is missing the `[Fact]` attribute. The value of the `value` attribute needs to be set always to match how the application was published.

PR #84, opened 2016-02-09
Strong naming dotnet-publish-iis

moozzyk:


NTaylorMullen:
:shipit: 

I'd add a comment to the commit saying why you couldn't do it before for clarification.

moozzyk:
Not sure if a year from now anyone will care that we could not sign because a dependency was not signed or that I forgot to even try originally (or both)..

PR #90, opened 2016-02-20
Creating webroot if one does not exists

moozzyk:
One scenario it is useful is where the webroot does not contain any files. In this case even if the webroot is included in the "content" section dotnet publish does will not publish it.


davidfowl:
Looks good. Do we have functional  tests for this thing?

GuardRex:
> In this case even if the webroot is included in the "content" section dotnet publish does will not publish it.

Related request: https://github.com/dotnet/cli/issues/1396

moozzyk:
@GuardRex :+1:

moozzyk:
@davidfowl - we have ~90% test coverage but no functional tests. Unfortunately adding functional tests to this repo may be troublesome due to https://github.com/dotnet/cli/issues/1215.  The tool however is run each time you want to publish for IIS which should cover the missing 10%.

PR #94, opened 2016-03-09
Appending .exe to the project name instead of just changing the exten…

### LINKED ISSUE #93: Wrong processPath attribute value in published web.config for project name containing dot
Published web.config processPath attribute has wrong value when project folder contains dot. My project name is "MyProject.Api" which is then transformed by Path.ChangeExtension to "MyProject.exe". It should be "MyProject.Api.exe" because thats the name of the executable.

https://github.com/aspnet/IISIntegration/blob/dev/src/dotnet-publish-iis/PublishIISCommand.cs#L56

Current approach works only for project names that does not contain dot in their name. It should always add ".exe" extension.

moozzyk:
…sion

Fixes #93


Tratcher:
:shipit:

Issue #95, opened 2016-03-09
Make the extension method for WebHostBuilder do everything required to correctly set up default app to work in IIS


DamianEdwards:
Right now, an app needs to configure itself in 3 places for IIS integration:
- Dependency and post-publish script in `project.json`
- Call to add IIS integration during `WebHostBuilder` phase in `Program.cs`
- Call to add HTTP forwarders & IIS integration middleware in `Startup.cs`

This is too many places. I'd like us to remove the need for any code in `Startup.cs` for IIS integration by default. Instead, a single extension-method call in the `WebHostBuilder` phase should configure the IIS URL, and add both the required middleware via an `IStartupFilter`.

- [x] Add the IIS auth middleware via the startup filter
- [x] Add the header overrides middleware via the startup filter

I'd also like the package names to be:
- [x] Microsoft.AspNetCore.Server.IISIntegration
- [x] Microsoft.AspNetCore.Server.IISIntegration.Tools

Example resultant `project.json`:
``` json
{
  "compilationOptions": {
    "emitEntryPoint": true
  },
  "runtimeOptions": {
    "gcServer": true
  },
  "dependencies": {
    "Microsoft.NET.DotNetCore": {
      "version": "1.0.0",
      "type": "platform"
    },
    "Microsoft.AspNetCore.Server.Kestrel": "1.0.0",
    "Microsoft.Extensions.Logging.Console": "1.0.0",
    "Microsoft.AspNetCore.Server.IISIntegration": "1.0.0"
  },
  "tools": {
    "Microsoft.AspNetCore.Server.IISIntegration.Tools": "1.0.0"
  },
  "frameworks": {
    "dotnetcore1.0": {}
  },
  "scripts": {
    "postpublish": "dotnet publish-iis --publish-folder %publish:OutputPath%"
  }
}
```

DamianEdwards:
ping @muratg @moozzyk @Tratcher

moozzyk:
@DamianEdwards  - `dotnet-publish-iis can't be renamed` to `Microsoft.AspNetCore.Server.IISIntegration.Tools`. Dotnet tools (and therefore the packages) have to start with `dotnet-` prefix

DamianEdwards:
I don't believe that is true, or at least will continue to be. The dispatching logic looks for assemblies in referenced packages without regard to the package name itself.

DamianEdwards:
By that I mean, we'd need the package name to be different to the tool assembly name.

[...]

PR #98, opened 2016-03-11
Pawelka/aspnetcore

moozzyk:
Issue #95 (point 2, part 1)

### DIFF
--- a/test/Microsoft.AspNetCore.Server.IISIntegration.Tools.Tests/PublishIISCommandFacts.cs
+++ b/test/Microsoft.AspNetCore.Server.IISIntegration.Tools.Tests/PublishIISCommandFacts.cs
@@ -24,7 +24,7 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.Tools.Tests
             new PublishIISCommand(folders.PublishOutput, folders.ProjectPath, null).Run();
 
             var processPath = (string)GetPublishedWebConfig(folders.PublishOutput, webRoot)
-                .Descendants("httpPlatform").Attributes("processPath").Single();
+                .Descendants("aspNetCore").Attributes("processPath").Single();
 
             Assert.Equal($@"..\projectDir.exe", processPath);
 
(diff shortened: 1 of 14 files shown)


Tratcher:
So this is a straight copy and rename of both the tool and tests? I assume you're planning to merge it at leave the old tool in place for a while?

You missed updating the SLN file and the nugetpackageverifier.json.

moozzyk:
Yes. 1st commit is a copy. 2nd commit is changing to aspnetcoremodule. 3rd commit is reacting to changes in env variables. dotnet-publish-iis stays as is for now. Once aspnetcoremodule ships we will be able to remove it. 
Good point about sln and package verifier. I will update both.

test/Microsoft.AspNetCore.Server.IISIntegration.Tools.Tests/PublishIISCommandFacts.cs
```diff
@@ -0,0 +1,153 @@
...
+        public void PublishIIS_modifies_existing_web_config()
+        {
+            var webRoot = "wwwroot";
+            var folders = CreateTestDir("{}", webRoot);
+
+            File.WriteAllText(Path.Combine(folders.PublishOutput, webRoot, "web.config"),
+@"<configuration>
+  <system.webServer>
+    <handlers>
+      <add name=""aspNetCore"" path=""*"" verb=""*"" modules=""aspNetCoreModule"" resourceType=""Unspecified""/>
```
Tratcher: FYI: the web.config is case sensitive, this needs to say `AspNetCoreModule`.

moozzyk: Good catch. I must have broken it with one of the bulk replaces.. Will fix.

muratg:
You'll also need to update the `packages.csv` file. This is used by the split-packages tool (categorizing packages to shipping/non-shipping/external).

File is here: https://github.com/aspnet/specs/blob/master/packages/packages.csv

moozzyk:
:up::date:

Tratcher:
:shipit:

PR #100, opened 2016-03-18
Reacting to ASPNETCORE_APPLICATIONBASE -> ASPNETCORE_CONTENTROOT rename

moozzyk:


src/dotnet-publish-iis/WebConfigTransform.cs
```diff
@@ -70,7 +70,7 @@ private static void TransformHttpPlatform(XElement httpPlatformElement, string a
 
         private static void AddApplicationBase(XElement httpPlatformElement)
         {
-            const string appBaseKeyName = "ASPNETCORE_APPLICATIONBASE";
+            const string appBaseKeyName = "ASPNETCORE_CONTENTROOT";
```
Tratcher: not changing your variable names?

Tratcher:
:shipit:

src/dotnet-publish-iis/WebConfigTransform.cs
moozzyk: good point

PR #106, opened 2016-03-25
Updating publish-iis to be a portable app

moozzyk:


Tratcher:
Does this unblock it so we can actually use it now?

moozzyk:
It does for the old publish-iis (i.e. the one that updated the config for httpPlatformHandler) if you have the right version of the runtime (as per https://github.com/dotnet/cli/issues/2071). 
It moves the new publish-iis further - I want to make other changes and tests separately.

samples/IISSample/project.json
```diff
@@ -15,7 +14,12 @@
   "frameworks": {
     "net451": { },
     "netstandardapp1.5": {
-      "imports": "portable-net451+win8"
+      "imports": [
+        "dnxcore50", "portable-net451+win8"
+      ],
+      "dependencies": {
+        "NETStandard.Library": "1.5.0-*"
```
weshaggard: This should be `1.5.0-rc2-*`

src/dotnet-publish-iis/project.json
```diff
@@ -14,12 +14,21 @@
     "Microsoft.Extensions.Configuration.Json": "1.0.0-*",
     "Microsoft.DotNet.ProjectModel": "1.0.0-*",
     "Microsoft.DotNet.Cli.Utils": "1.0.0-*",
-    "NETStandard.Library": "1.5.0-*",
     "System.Diagnostics.Process": "4.1.0-*"
   },
 
   "frameworks": {
-    "dnxcore50": {
+    "netstandard1.5": {
```
weshaggard: If this is an app perhaps it should be `netcoreapp1.0`. That should also eliminate the need for the imports.

samples/IISSample/project.json
moozzyk: The feed use (aspnetcidev/aspnetvolatile) only has compatible packages (i.e. we won't take rc3 packages)

src/dotnet-publish-iis/project.json
moozzyk: @weshaggard - this is a tool.

src/Microsoft.AspNetCore.Server.IISIntegration.Tools/project.json
```diff
@@ -14,16 +14,21 @@
...
+    "netstandard1.5": {
+       "imports": [
         "portable-net451+win8",
         "dnxcore50"
-      ]
+      ],
+      "dependencies": {
+        "Microsoft.NETCore.App": {
+          "type": "platform",
+          "version": "1.0.0-rc2-*"
```
moozzyk: I should actually drop rc2 from this version...

src/dotnet-publish-iis/project.json
weshaggard: It is a tool but it is referencing the Microsoft.NETCore.App meta-package trying to be a shared .NET Core app so the correct TFM is netcoreapp1.0 for those.

src/Microsoft.AspNetCore.Server.IISIntegration.Tools/project.json
weshaggard: You guys can decide but I would suggest keeping rc2 in the version.

src/dotnet-publish-iis/project.json
moozzyk: OK. In that case I will try changing it to `netcoreapp1.0`

weshaggard: As an FYI this might not work until @anurse actually disabled RID inference in the cli.

troydai: @brthor does mention in this [issue](https://github.com/dotnet/cli/issues/1970) that tools should target `netstandard1.5`. I guess the suggestion has been changed? changed?

weshaggard: I don't believe there is anything special about tools here. Until recently portable apps needed to use netstandard1.5 but with the latest cli tools netcoreapp1.0 should be the tfm.

moozzyk:
:up::date:
rebased on release and updated to netcoreapp1.0

src/Microsoft.AspNetCore.Server.IISIntegration.Tools/project.json
```diff
@@ -14,16 +14,21 @@
...
     "Microsoft.DotNet.Cli.Utils": "1.0.0-*",
-    "NETStandard.Library": "1.5.0-*",
     "System.Diagnostics.Process": "4.1.0-*"
   },
 
   "frameworks": {
-    "netstandardapp1.5": {
-      "imports": [
+    "netcoreapp1.0": {
+       "imports": [
```
anurse: spacing

src/dotnet-publish-iis/project.json
```diff
@@ -14,12 +14,21 @@
...
     "Microsoft.DotNet.ProjectModel": "1.0.0-*",
     "Microsoft.DotNet.Cli.Utils": "1.0.0-*",
-    "NETStandard.Library": "1.5.0-*",
     "System.Diagnostics.Process": "4.1.0-*"
   },
 
   "frameworks": {
-    "dnxcore50": {
+    "netcoreapp1.0": {
+       "imports": [
```
anurse: spacing

anurse:
:shipit:

src/Microsoft.AspNetCore.Server.IISIntegration.Tools/project.json
```diff
@@ -14,16 +14,21 @@
     "Microsoft.Extensions.Configuration.Json": "1.0.0-*",
```
Tratcher: Still needed?

moozzyk: Looks like not needed anymore - removing.

src/Microsoft.AspNetCore.Server.IISIntegration.Tools/project.json
moozzyk: fixed

src/dotnet-publish-iis/project.json
moozzyk: fixed

Eilon:
cc @NTaylorMullen @davidfowl 

I thought we were going to wait to do portable apps until we had roll-forward support, no? Won't this tool just be broken anyway once we take a new CLI?

davidfowl:
The new CLI has it and since the corefx packages haven't changed in a while this should be fine. We just can't grab a newer Microsoft.NetCore.App than the CLI that we have and @NTaylorMullen is going to gate that.

NTaylorMullen:
@Eilon ya this went in premature. @moozzyk and I talked in person. He wasn't aware (he got it in as I was writing up an email). 

@davidfowl I beleive @Eilon is referring to the ability to have auto-roll forward support. @DamianEdwards @Eilon and I talked and the decision was to wait until we had a CLI that supported the roll forward to get portable app changes in. As it stands currently, we wont be broken unless we pull a newer shared framework; I believe that was @DamianEdwards and @Eilon's concern about getting things like this in too early (since we were blocked by the new CLI).

@ryanbrandenburg can comment on how close we are to having a CLI with the auto-rollforward capability.

moozzyk:
So far it did not break anything. However looks like the test stopped running once test projects were converted to `netcoreapp1.0` (maybe this is why it did not break). This conversion revealed two blocking bugs - one in NuGet and one in CLI (https://github.com/dotnet/cli/issues/2347). There is a workaround for the NuGet issue but not for the CLI. @pakrym is working on both (I think the fix in NuGet has already been merged).

davidfowl:
It won't break anything because the version of the shared framework matches the package.

Eilon:
Yeah I'm not *super* concerned about this going in early, but I do think it'll break very soon. But hopefully we get the new CLI ASAP and all will be well.

moozzyk:
I am ready to revert this change if it breaks. I just did not want to do this proactively because it did not break.

Eilon:
Yeah that's fine.

Issue #109, opened 2016-03-30
publish-iis needs to understand portable vs. self contained apps


moozzyk:
Currently publish-iis always sets `processPath` attribute to the exe - e.g. (`processPath="..\myapp.exe"`). If the application is using the shared runtime there will be no .exe and the config file should look differently - `processPath="dotnet" arguments="..\myapp.dll"`

Note that IIS runs using its own account and dotnet may not be on the path/accessible - we will probably need some guidance about this. 
/cc @danroth27

moozzyk:
Fixed in 1410a5a096efbd1ef2a0c6cd7205a56729900319

Issue #128, opened 2016-04-12
publish-iis should pick up the actual app name


moozzyk:
dotnet now supports specifying outputName in which case the name of the application executable/dll might be different from the package name. In this case using just [`Project.Name`](https://github.com/aspnet/IISIntegration/blob/dev/src/dotnet-publish-iis/PublishIISCommand.cs#L83) is not enough.

moozzyk:
Fixed in f58641b

PR #136, opened 2016-04-15
Enabling publishing portable apps

moozzyk:
Addresses #109

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IISIntegration.Tools/Program.cs
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration.Tools/Program.cs
@@ -20,13 +20,15 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.Tools
             app.HelpOption("-h|--help");
 
             var publishFolderOption = app.Option("--publish-folder|-p", "The path to the publish output folder", CommandOptionType.SingleValue);
+            var frameworkOption = app.Option("-f|--framework <FRAMEWORK>", "Target framework of application being published", CommandOptionType.SingleValue);
             var projectPath = app.Argument("<PROJECT>", "The path to the project (project folder or project.json) being published. If empty the current directory is used.");
 
             app.OnExecute(() =>
             {
                 var publishFolder = publishFolderOption.Value();
+                var framework = frameworkOption.Value();
 
-                if (publishFolder == null)
+                if (publishFolder == null || framework == null)
                 {
                     app.ShowHelp();
                     return 2;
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Server.IISIntegration.Tools/Program.cs
```diff
@@ -20,21 +20,23 @@ public static int Main(string[] args)
             app.HelpOption("-h|--help");
 
             var publishFolderOption = app.Option("--publish-folder|-p", "The path to the publish output folder", CommandOptionType.SingleValue);
+            var frameworkOption = app.Option("-f|--framework <FRAMEWORK>", "Target framework of application being published", CommandOptionType.SingleValue);
```
BrennanConroy: Should we say something about this argument being required?

src/Microsoft.AspNetCore.Server.IISIntegration.Tools/PublishIISCommand.cs
```diff
@@ -67,9 +73,14 @@ private string GetApplicationBasePath()
...
-            return ProjectReader.GetProject(Path.Combine(applicationBasePath, "project.json")).Name;
+            var project = ProjectReader.GetProject(Path.Combine(applicationBasePath, "project.json"));
+
+            return new ProjectContextBuilder()
+                .WithProject(project)
+                .WithTargetFramework(framework)
+                .Build();
         }
 
         private static bool ConfigureForAzure()
```
Tratcher: Is there a reason ConfigureForAzure can't be specified via an optional command line arg? Having the env variable too is OK, but it seems strange that you can't just specify this directly.

Tratcher:
What does the new project.json tool script look like?

src/Microsoft.AspNetCore.Server.IISIntegration.Tools/Program.cs
moozzyk: We could. At the moment we don't say anything for --publish-folder so I did not add it here either. The reality is that typically people will not run it manually. Rather it will be in project.json in the post publish script section (created by our templates)

src/Microsoft.AspNetCore.Server.IISIntegration.Tools/PublishIISCommand.cs
moozzyk: It could be a parameter. In practice it won't be used since in majority of cases this tool is run as postpublishscript so you would have to modify the project.json file each time you want to publish when you know if the app will be hosted on Azure or not. Currently VS just sets the env variable when publishing without having to modify project.json. 
If you feel strongly about this feel free to file a bug - it's a different issue (and most likely not RC2)

moozzyk:
Haven't tried it end to end but it should be something along these lines:

```
  "scripts": {
    "postpublish": "dotnet publish-iis --publish-folder %publish:OutputPath% --framework %publish:publish:FullTargetFramework%"
  }
```

moozzyk:
We would also need to add configuration for this: https://github.com/aspnet/IISIntegration/issues/128. I think it is not for RC2 and it would be optional.

Tratcher:
"%publish:publish:" really?
:shipit:

moozzyk:
Copied incorrectly. It is just `%publish:FullTargetFramework%`

PR #143, opened 2016-04-20
Convert to portable app

JunTaoLuo:
This needs https://github.com/aspnet/Hosting/pull/719

cc @pakrym @Tratcher

### DIFF
--- a/samples/IISSample/project.json
+++ b/samples/IISSample/project.json
@@ -10,10 +10,13 @@
   },
   "frameworks": {
     "net451": { },
-    "netstandardapp1.5": {
+    "netcoreapp1.0": {
       "dependencies": {
-          "NETStandard.Library": "1.5.0-*"
-        },
+        "Microsoft.NETCore.App": {
+          "version": "1.0.0-*",
+          "type": "platform"
+        }
+      },
       "imports": "dnxcore50"
     }
   },
--- a/samples/IISSample/web.config
+++ b/samples/IISSample/web.config
@@ -1,9 +1,23 @@
-﻿<?xml version="1.0"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
+
+  <!-- Configure your application settings in appsettings.json. Learn more at http://go.microsoft.com/fwlink/?LinkId=786380 -->
+  <!-- This repository cannot use the publish-iis tool because there's a bug in the installer if the tool package is a project reference -->
+  <!-- See https://github.com/dotnet/cli/issues/1215 -->
+
   <system.webServer>
     <handlers>
-      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
+      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
     </handlers>
-    <aspNetCore processPath="%DNX_PATH%" arguments="%DNX_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" />
+    <!-- This set of attributes are used for launching the sample using IISExpress via Visual Studio tooling -->
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false"/>
+    
+    <!-- This set of attributes are used for launching the sample for full CLR (net451) without Visual Studio tooling
+    <aspNetCore processPath=".\IISSample.exe" arguments="" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false"/>
+    -->
+    
+    <!-- This set of attributes are used for launching the sample for Core CLR (netcoreapp1.0) without Visual Studio tooling
+    <aspNetCore processPath="dotnet" arguments=".\IISSample.dll" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false"/>
+    -->
   </system.webServer>
 </configuration>
--- a/test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/HelloWorldTest.cs
+++ b/test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/HelloWorldTest.cs
@@ -19,13 +19,14 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
         [ConditionalTheory]
         [OSSkipCondition(OperatingSystems.Linux)]
         [OSSkipCondition(OperatingSystems.MacOSX)]
-        //[InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x86, "http://localhost:5061/", ServerType.Kestrel)]
-        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, "http://localhost:5062/", ServerType.Kestrel)]
-        //[InlineData(RuntimeFlavor.CoreClr, RuntimeArchitecture.x86, "http://localhost:5063/", ServerType.Kestrel)]
-        [InlineData(RuntimeFlavor.CoreClr, RuntimeArchitecture.x64, "http://localhost:5064/", ServerType.Kestrel)]
-        public Task HelloWorld_IISExpress(RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl, ServerType delegateServer)
+        //[InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x86, "http://localhost:5061/", ServerType.Kestrel, ApplicationType.Standalone)]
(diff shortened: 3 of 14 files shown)


samples/IISSample/web.config
```diff
@@ -1,9 +0,0 @@
-﻿<?xml version="1.0"?>
```
Tratcher: don't delete this file, update it to match the templates.

moozzyk: why?

moozzyk: because it has `forwardWindowsAuthToken="false"` I suppose...

test/TestSites/web.config
```diff
@@ -6,6 +6,6 @@
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
     </handlers>
-    <aspNetCore processPath=".\TestSites.exe" arguments="" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="true" />
+    <aspNetCore stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="true" />
```
Tratcher: The template includes processPath and arguments with environment variables, correct?

samples/IISSample/web.config
JunTaoLuo: this sample is not used for testing NTLM though so we don't need to set `forwardWindowsAuthToken` to true

Tratcher: Because we use this as a reference to show people what it should look like, and so it's easier to modify when testing.

moozzyk: Should we remove publish-iis tool then? Because then it will go and modify the web.config so what you show is not what you get...

Tratcher: No, keep the tooling. The sample should resemble the templates, which contain a web.config with environment variables.

test/TestSites/web.config
JunTaoLuo: The problem is that if you have a standalone app, the publish-iis tool only overwrite the process path and leaves the envvar in the arguments attribute. Better to remove both and let the publish-iis tool set them if needed.

samples/IISSample/web.config
moozzyk: Then the arguments needs to be empty and probably processPath as well.

test/TestSites/web.config
JunTaoLuo: Or as @moozzyk has mentioned, we can leave both fields empty.

GuardRex: Probably generate less confusion and fewer support issues to have both with no values and a comment  in the file that the values will be provided by the `publish-iis` tool.
```xml
<!-- The values for 'processPath' and 'arguments' will be injected by the 'publish-iis' tool when the 
application is published. -->
<aspNetCore processPath="" arguments="" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" 
forwardWindowsAuthToken="false" />
```

Tratcher: Ask @BillHiebert what VS will do when running IIS Express locally?

samples/IISSample/project.json
```diff
@@ -10,14 +10,23 @@
...
         },
+      },
       "imports": "dnxcore50"
     }
   },
-  "content": [
-    "web.config"
-  ]
+  "tools": {
+    "Microsoft.AspNetCore.Server.IISIntegration.Tools": {
```
Tratcher: We can't add the tool to samples or tests in this repo, you can't use tools when you have a source reference to it.

moozzyk: :+1: - I missed this is in IIS Integration repo

test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/HelloWorldTest.cs
```diff
@@ -19,42 +19,44 @@ public class HelloWorldTests
         [ConditionalTheory]
         [OSSkipCondition(OperatingSystems.Linux)]
         [OSSkipCondition(OperatingSystems.MacOSX)]
-        //[InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x86, "http://localhost:5061/", ServerType.Kestrel)]
-        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, "http://localhost:5062/", ServerType.Kestrel)]
-        //[InlineData(RuntimeFlavor.CoreClr, RuntimeArchitecture.x86, "http://localhost:5063/", ServerType.Kestrel)]
-        [InlineData(RuntimeFlavor.CoreClr, RuntimeArchitecture.x64, "http://localhost:5064/", ServerType.Kestrel)]
-        public Task HelloWorld_IISExpress(RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl, ServerType delegateServer)
+        //[InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x86, "http://localhost:5061/", ServerType.Kestrel, ApplicationType.Standalone)]
+        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, "http://localhost:5062/", ServerType.Kestrel, ApplicationType.Standalone)]
```
JunTaoLuo: This is a little misleading since Standalone CLR app isn't quite right. The reason for this is that for this particular repo, I'm using `AppliationType` to determine whether to use the `TestApp` or `TestApp.Portable`. The published CLR app uses the same `web.config` as the standalone CoreCLR app, launching via `.\TestApp.exe` whereas the porable CoreCLR app launches via `dotnet .\TestApp.Portable.dll`. For other repos, we can rely on the publish-iis tool to do the correct thing and won't have this confusion.

JunTaoLuo:
@Tratcher I added a portable app for testing this repo since we cannot rely on publish-iis to correctly set the fields inside `web.config`. We'll probably use something similar for testing both application types in other repos if this pattern looks good.

test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/HelloWorldTest.cs
JunTaoLuo: To reduce the confusion here, I can create a separate project for full CLR and have the tests grab either the full CLR , standalone CoreCLR or portable CoreCLR apps using `RuntimeFlavor` and `ApplicationType`. But I'm not sure if that's worth the trouble.

moozzyk:
@JunTaoLuo - _we cannot rely on publish-iis to correctly set the fields inside web.config_ - can you elaborate? Is it a problem with publish-iis? (If this is about arguments attribute I have a PR out - I need to address a comment though)

moozzyk:
@JunTaoLuo - ignore my comment - for whatever reason I thought it was Hosting repo.

JunTaoLuo:
@moozzyk Though it wouldn't work for Hosting either due to dependencies.

samples/IISSample/web.config
```diff
@@ -1,9 +1,23 @@
...
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
+
+  <!--
+    Configure your application settings in appsettings.json. Learn more at http://go.microsoft.com/fwlink/?LinkId=786380
+  -->
+
   <system.webServer>
```
Tratcher: Include this comment from the test web.configs:
`<!-- This repository cannot use the iis publish tool because there's a bug in the installer if the tool package is a project reference -->`

test/TestSites/web.config
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0"?>
+﻿<?xml version="1.0"?>
 <configuration>
   <system.webServer>
     <!-- This repository cannot use the iis publish tool because there's a bug in the installer if the tool package is a package reference -->
```
Tratcher: Hmm, this should say project reference rather than package reference.

test/TestSites.Portable/web.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0"?>
+<configuration>
+  <system.webServer>
+    <!-- This repository cannot use the iis publish tool because there's a bug in the installer if the tool package is a package reference -->
```
Tratcher: project reference.

test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/HelloWorldTest.cs
Tratcher: Would it make more sense to have an additional ApplicationType for CLR?

Tratcher:
:shipit:

test/TestSites.Portable/web.config
moozzyk: nit: "publish-iis tool" instead of "iis publish tool"
nit: the bug is that cli does not support resolving commands in if they are not packages

test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/HelloWorldTest.cs
JunTaoLuo: ApplicationType is only relevant for CoreCLR apps. The self host deployer ignores this field for CLR apps so I don't think that makes too much sense either.

Issue #141, opened 2016-04-20
Add optional feature to reject requests without IIS token


blowdart:
This would prevent requests on the same machine bypassing auth/path authorization provided by IIS.

Should default to false though.

Tratcher:
Why default to false? Is there a mainstream scenario where you wouldn't want it on?

Tratcher:
Offline dicussion: On by default, no option to turn it off. The middleware is only added to the pipeline if the AspNetCoreModule environment variables are detected at startup, so enabling this will not affect self-host scenarios.

PinpointTownes:
Why not providing an option to disable this feature? What happens if I want to host my app on various environments? (e.g behind IIS on a Windows machine, behind nginx on a Linux-based machine).

PR #148, opened 2016-04-21
Require the MSAspNetCoreToken or send a 400.

Tratcher:
#141 @blowdart 
@Eilon @DamianEdwards Please approve for RC2.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
@@ -56,8 +56,8 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
         {
             if (!string.Equals(_pairingToken, httpContext.Request.Headers[MSAspNetCoreToken], StringComparison.Ordinal))
             {
-                _logger.LogTrace($"'{MSAspNetCoreToken}' does not match the expected pairing token '{_pairingToken}', skipping {nameof(IISMiddleware)}.");
-                await _next(httpContext);
+                _logger.LogError($"'{MSAspNetCoreToken}' does not match the expected pairing token '{_pairingToken}', request rejected.");
+                httpContext.Response.StatusCode = 400;
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
```diff
@@ -57,7 +57,7 @@ public IISMiddleware(RequestDelegate next, ILoggerFactory loggerFactory, IOption
             if (!string.Equals(_pairingToken, httpContext.Request.Headers[MSAspNetCoreToken], StringComparison.Ordinal))
             {
                 _logger.LogTrace($"'{MSAspNetCoreToken}' does not match the expected pairing token '{_pairingToken}', skipping {nameof(IISMiddleware)}.");
```
PinpointTownes: IMHO, you should consider updating the log level of this message to error (or critical, since it would indicate that the request wasn't handled by IIS, which may be terribad in production).

Tratcher:
Updated the log level.

muratg:
@Tratcher did you start an internal thread for RC2 approval?

blowdart:
Yes he did

muratg:
:+1: :shipit:

superlogical:
I'm getting an error 

`'MS-ASPNETCORE-TOKEN' does not match the expected pairing token '7f91cf1a-9295-4bcc-9e3d-88ae09909389', request rejected.`

Any ideas as to what I'm doing wrong? I'm deploying onto a windows service with IIS

Eilon:
@superlogical can you log a new bug on this, including any info you have on which versions of ASP.NET Core and the ASP.NET Core Module you're using?

Tratcher:
That's very unexpected. This check is used to block requests that bypass IIS and are sent directly to the back-end process. Which port did you configure IIS on, and which port did you send the request to?

What do you mean you're deploying IIS in a windows service? IIS runs as it's own service, not inside your own service. Or are you trying to use IIS Express?

Issue #142, opened 2016-04-20
Update publish-iis to set the dll location to home\site\wwwroot\app.dll instead of home\site\app.dll when deploying to App Service


muratg:
For Antares (App Service) deployments, approot is wwwroot.

cc @BrennanConroy

moozzyk:
```
<?xml version="1.0" encoding="utf-8"?> 
<configuration> 
<system.webServer> 
<handlers> 
<add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" /> 
</handlers> 
<aspNetCore processPath="%home%\site\wwwroot\HelloMvc.exe" stdoutLogEnabled="false" stdoutLogFile="\\?\%home%\LogFiles\stdout" startupTimeLimit="3600" /> 
</system.webServer> 
</configuration>  
```

muratg:
Looks good!

GuardRex:
This is confusing given the (philosophical) direction things have gone with `wwwroot`. It's come to be truly for static file serving. Is this still where Azure Apps will look for `web.config`? I thought that would be changed (i.e., backed out a level, then searched recursively down folders).

I was looking at what @rmja put up in https://github.com/aspnet/IISIntegration/issues/105#issuecomment-212417725, and I was sort of expecting that you would go in the ...
```
processPath="%home%\application\HelloMvc.exe"
```
... direction with the contents of the `publish` folder moved into `application` and the shared framework put into `dotnet` or something. I'm definitely a little rusty on Azure Apps, but I was basing this on what @rmja posted.

Gut feeling: The closer it is to IIS, the less confusion and the fewer the number of support requests down the road.

I thought you would have something like (standalone) ...
```
--- application
    --- Logs
        --- dummy `.txt` file (open issue on this bug; folder won't pub without it)
    --- refs
    --- Views
        --- _ViewImports.cshtml, view folders (Shared, Home, etc.), views
        --- DLL's
    --- wwwroot
        --- static files
    --- DLL's, myapp.dll, myapp.exe
    --- web.config
```
... or shared ...
```
--- dotnet
    --- shared
        --- assets
    --- dotnet.exe
    --- hostfxr.dll
--- application
    --- Logs
        --- dummy `.txt` file (open issue on this bug; folder won't pub without it)
    --- refs
    --- runtimes
        --- folders for RID's native assets
    --- Views
        --- _ViewImports.cshtml, view folders (Shared, Home, etc.), views
[...]

PR #151, opened 2016-04-22
Making the path to app relative for Azure

moozzyk:
Removing startupLimitType - it is obsolete in AspNetCoreModule

Addresses #142


muratg:
:shipit:

PR #157, opened 2016-04-27
Add more rids so build succeeds on OSX and Linux

pranavkm:


pranavkm:
cc @Tratcher \ @moozzyk

PR #162, opened 2016-05-02
Add transitive reference to PlatformAbstractions

ryanbrandenburg:


moozzyk:
Why is this needed now?

src/Microsoft.AspNetCore.Server.IISIntegration.Tools/project.json
```diff
@@ -11,6 +11,7 @@
 
   "dependencies": {
     "Microsoft.Extensions.CommandLineUtils": "1.0.0-*",
+    "Microsoft.Extensions.PlatformAbstractions": "1.0.0-*","
```
davidfowl: this is messed up formatting

ryanbrandenburg:
This project references Microsoft.DotNet.ProjectModel, which recently lost PlatformAbstractions as a transitive dependency, but still needs it at build time.

@NTaylorMullen is that explanation any good?

davidfowl:
Where is it used?

NTaylorMullen:
Actually, looks like this project doesn't use platform abstractions. Don't have to do this project :+1:

ryanbrandenburg:
Ok then. Good catch @davidfowl.

Issue #187, opened 2016-05-23
HttpsTests consistently failed on Windows 2008


troydai:
Test:
```
[InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, "https://localhost:44396/", ApplicationType.Standalone)]
public Task Https_HelloWorld(RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl, ApplicationType applicationType)
{
    return HttpsHelloWorld(ServerType.IISExpress, runtimeFlavor, architecture, applicationBaseUrl, applicationType);
}
```

Failure:
```
System.Net.Http.HttpRequestException : An error occurred while sending the request.
---- System.Net.WebException : The underlying connection was closed: An unexpected error occurred on a send.
-------- System.IO.IOException : Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.
------------ System.Net.Sockets.SocketException : An existing connection was forcibly closed by the remote host
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Testing.RetryHelper.<RetryRequest>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.Server.Testing.RetryHelper.<RetryRequest>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.HttpsTest.<HttpsHelloWorld>d__1.MoveNext() in D:\b\work\35116a8defcd3a21\Universe\IISIntegration\test\Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests\HttpsTest.cs:line 60
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #188, opened 2016-05-23
disable two failed tests temporarily

troydai:
issue: #187 

buddy build note: these two tests fail on Windows 2008 platform only for the moment. disable during the investigation.


Tratcher:
:-1: This is not a scenario we can ignore. Work with @JunTaoLuo to resolve this today.

troydai:
/cc @Eilon 

Both @Tratcher and @JunTaoLuo are against the idea to disable this test. @JunTaoLuo is actively investigate this one for now. 

I'll hold on till later today to see if these two failures are the last block issues toward a coherence-signed.

Eilon:
Disabling is fine temporarily. Getting builds out is critical. We just need to make fixing the tests/product a super high priority. But we shouldn't block everything due to it.

Eilon:
:shipit:

PR #195, opened 2016-05-27
Update metadata

JunTaoLuo:
#186 

cc @Eilon


src/Microsoft.AspNetCore.Server.IISIntegration.Tools/project.json
```diff
@@ -1,5 +1,16 @@
 {
   "version": "1.0.0-preview2-*",
+  "description": "Command line tool for publishing web applications to be hosted using IIS.",
```
Eilon: BTW it might be nice to put the actual command name in the description too. Something like this: https://github.com/aspnet/RazorTooling/blob/dev/src/Microsoft.AspNetCore.Razor.Tools/project.json#L3

Eilon:
One comment, then :shipit:

Issue #242, opened 2016-07-27
UseIISIntegration should delay setting urls


Tratcher:
Today it's a common mistake for callers to reverse UseIISIntegration and UseUrls, breaking their app in IIS/Express and Azure.

Proposal, rather than having UseIISIntegration set the url [immediately](https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs#L39), set it later when Build is called (e.g. inside the [ConfigureServices ](https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs#L42) callback) or in the [IStartupFilter ](https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/IISSetupFilter.cs#L21) via IServerAddressesFeature.

halter73:
Seems like a good idea. Today, you can check whether UseIISIntegration is activated in Program.Main() by checking whether whether the urls setting was changed after calling it. Technically, changing when we change the urls settings is breaking, but it probably won't affect too many apps.

It would be nice if there was still an easy workaround to figure out if the app is running behind IIS in Program.Main() after this change.

Tratcher:
What would you do in Main if you did detect IIS?

Checking for the environment variables directly may be better. Or we could provide an IsRunningInIIS API.

halter73:
To disable Kestrel connection filters that are replaced by IIS modules perhaps. Admittedly it's probably not common to use a single code base to deploy with and without IIS, so I don't think this would break (m)any people.

PR #250, opened 2016-08-15
Delay registering the url until Build

Tratcher:
#242 @davidfowl @JunTaoLuo @shirhatti @muratg 

Users too often break IIS by doing the following:
```
new WebHostBuilder()
 .UseKestrel()
 .UseIISIntegration()
 .UseUrls("http://mysite.com:80/");
```

This change delays registering IIS's dynamic url until Build is called so it won't be overwritten by UseUrls.


Tratcher:
@halter73

halter73:
There goes the most convenient way of determining whether or not you're hosted by IIS. I doubt anyone other than me really cares about that though :smile: 

I wonder if there is a good way to notify users they are calling `UseUrls` after `UseIISIntegration`. Maybe just throw, since you would never want to call those two methods in that order. Better than what we have today though :+1:

JunTaoLuo:
:shipit:

PR #265, opened 2016-09-19
Do not call WindowsIdentity.AuthenticationType

### LINKED ISSUE #231: Work around UnauthorizedAccessException for WindowsIdentity.get_AuthenticationType on IIS
### Title
A `System.UnauthorizedAccessException` occurs when accessing `System.Security.Principal.WindowsIdentity.WindowsIdentity.AuthenticationType` getter on IIS, breaking the Windows Authentication of IISIntegration.

### Functional impact
When using Windows Authentication, any time the `Microsoft.AspNetCore.Server.IISIntegration.AuthenticationHandler.AuthenticateAsync` method is called, which accesses the `AuthenticationType` getter, the following exception is thrown:
````
System.UnauthorizedAccessException: Attempted to perform an unauthorized operation.
   at System.Security.Principal.WindowsIdentity.get_AuthenticationType()
````
This renders the Windows Authentication unusable on our IIS deployment server.

### Minimal repro steps
* We use a combination of Windows Authentication with `IISOptions` `AutomaticAuthentication` and `ForwardWindowsAuthentication` set to true, and Cookies authentication via `UseCookieAuthentication()`. Having Windows Authentication alone should however be sufficient to reproduce.
* Annotate some controller with an authorization policy that specifies
````
policy.RequireAuthenticatedUser();
policy.AddAuthenticationSchemes("NTLM");
````
* Deploy to an IIS (at least with specs mentioned below) with an application pool running under the default "ApplicationPoolIdentity"
* Access the annotated controller

### Expected result
Browser's Windows Authentication popup due to HTTP 401 response that gets the `WWW-Authenticate` header field with "NTLM" injected by IIS.

### Actual result
We get an unhandled exception as response, with the above mentioned exception details and causes.

### Further technical details
*Deployment server:*
* IIS 7.5
* Windows Server 2008 R2 x64
* .NET Framework 4.5.2
* ASP.NET Core 1.0.0
* HTTP Platform Handler 1.2

*Details*
Accessing the `AuthenticationType` getter works when running on IIS Express on our development machines. It also works when changing the IIS application pool identity to "LocalSystem". It also works when currently not being logged in or logged in via Cookies authentication.
[This stackoverflow report](http://stackoverflow.com/questions/11587305/unauthorizedaccessexception-after-using-logonuser) describes the problem in detail. There it is suggested to use another `WindowsIdentity` [constructor with the AuthenticationType parameter](https://msdn.microsoft.com/en-us/library/w5whws2k(v=vs.110).aspx), which should solve the problem.

*Solution suggestions*
Maybe you could change the `AuthenticationHandler` to catch this exception and if it occurs assume Windows Authentication.
Maybe use the other `WindowsIdentity` constructor in `IISMiddleware`.
(if that does not disregard the `AuthenticationDescriptions` option)

*Notes*
The exception was also mentioned in [issue 75](https://github.com/aspnet/IISIntegration/issues/75#issuecomment-230927600), although in a different context.

### PR DESCRIPTION
**Pull request by ph1ll**
Do not call WindowsIdentity.AuthenticationType as this throws an exception under IIS, as per #231.

@Tratcher made a couple of suggestions on how to fix #231.

Whilst fixing WindowsIdentity to not throw from AuthenticationType would be the best solution, I dont believe that code is on GitHub and it certainly seems faster to fix the problem in this repo.

Tratcher's second option of comparing to `context.AuthenticationScheme` makes sense as this should match `IISOptions.AuthenticationDescriptions` (https://github.com/aspnet/IISIntegration/blob/cad873d9495c93f5fcb19383dad9f43502b5d841/src/Microsoft.AspNetCore.Server.IISIntegration/IISOptions.cs#L33), whereas `User.Identity.AuthenticationType` can contain other values such as "Kerberos".

### CODE REVIEW

File: src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs [hunk 1]
```diff
@@ -38,7 +38,7 @@ public Task AuthenticateAsync(AuthenticateContext context)
                 {
                     context.Authenticated(User, properties: null,
                         description: Options.AuthenticationDescriptions.FirstOrDefault(descrip =>
-                            string.Equals(User.Identity.AuthenticationType, descrip.AuthenticationScheme, StringComparison.Ordinal))?.Items);
+                            string.Equals(context.AuthenticationScheme, descrip.AuthenticationScheme, StringComparison.Ordinal))?.Items);
```
Tratcher: This won't work for "Automatic" auth. e.g. `[Authorize]`

ph1ll: I've just double checked and you are right. When `context.AuthenticationScheme` is "Automatic", the Description will end up being null, however the user is still authenticated.

This behaviour is the same as when `User.Identity.AuthenticationType` returns "Kerberos" under IIS Express (user is authenticated with a null Description), as `IISOptions.AuthenticationDescriptions` only contains "Negotiate" and "NTLM".

To me, an empty description seems preferable to throwing an exception, especially when that exception is only going to be thrown on the server, not dev workstations, making it harder to debug. 

Is there a preferred way of fixing this issue? If so I will be happy to update the MR.

Another solution would be to give the WindowsIdentity a type when it is created in https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs#L112 (e.g. `var winIdentity = new WindowsIdentity(handle, "IntegratedWindows");`), but this would mean the above description comparison, as well as `IISOptions.AuthenticationDescriptions` is redundant. This would have the added benefit of anyone calling `User.Identity.AuthenticationType` in user code would get the given string, rather than an exception.

pakrym: @ph1ll lets just do `null` for description if we cannot reliably detect it.

pakrym:
Closed via https://github.com/aspnet/IISIntegration/pull/282

PR #275, opened 2016-09-30
Only challenge if not already accepted.

Tratcher:
@kichalla @HaoK
Fixes music store conflict between cookie auth and windows


HaoK:
Well, I was pretty sure our change would break someone, didn't think it would be ourselves :)

Tratcher:
This is pretty specific to having multiple automatic handlers, which we already know is a terrible idea, we just can't disable this one without breaking existing apps using windows auth.

Tratcher:
@kichalla please re-enable the associated tests.

Issue #292, opened 2016-10-24
[Test] Modify tests to consume the newest build of ANCM


pranavkm:
Related to https://github.com/aspnet/ServerTests/issues/46

Once aspnet/AspNetCoreModule#20 is done, we'll need to modify IISExpress tests to run using the most recent build of ANCM package from the feed.

muratg:
Let's give it a shot in 1.1.0.  If we can't plan and finish the work, we can move it out to 1.2.0.

Tratcher:
Complication: The ANCM repo / nupkg does not support Win7. The real MSIs are built from a private repo with Win7 support.

Tratcher:
Hosting workaround for the Win7 issue: https://github.com/aspnet/Hosting/commit/4064b42a19d1f19f887a3f9b0688e8616889b93a
Prototype: https://github.com/aspnet/IISIntegration/commit/907301bd4a26c306c1b3ac6b68370e129ce0bb02
Waiting for the new package structure: https://github.com/aspnet/AspNetCoreModule/issues/37

PR #296, opened 2016-11-09
Test with the bin deployed aspnetcore.dll

Tratcher:
#292 
Depends on https://github.com/aspnet/Hosting/pull/879
@JunTaoLuo @pranavkm 
/cc: @shirhatti @pan-wang

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -3,6 +3,7 @@
   <packageSources>
     <clear />
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
+    <add key="ANCM" value="https://dotnet.myget.org/F/aspnetcoremodule/api/v3/index.json" />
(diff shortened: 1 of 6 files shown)


JunTaoLuo (approved):
looks good to me

Tratcher:
@muratg, mind if we make these test changes for 1.1? It gives us real coverage for ancm.

shirhatti:
Did we chose to run against 64-bit IIS Express (and hence test only 64-bit version of ANCM) as a conscious decision or is 32-bit IIS Express just an omission?

Tratcher:
That's an existing gap in our coverage. The current approach would test 32bit on 32bit OSs, but we don't run any of those. VS uses 32bit express, correct? Maybe we need a variant for it.

Tratcher:
https://github.com/aspnet/Hosting/issues/880

muratg:
@Tratcher can you start a thread for approval?

Tratcher:
Moving to dev for now.

PR #322, opened 2017-02-03
Upgrade to VS 2017

natemcmaster:
Remove project.json and move to MSBuild.

Had to disable standalone tests because of dotnet/sdk#696. I will open an issue to re-enable them.

Resolves https://github.com/aspnet/IISIntegration/issues/321


test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/NtlmAuthentationTest.cs
```diff
@@ -11,35 +11,51 @@
 using Microsoft.Extensions.Logging;
 using Xunit;
 using Xunit.Sdk;
+using Xunit.Abstractions;
+using Microsoft.AspNetCore.Testing;
```
BrennanConroy: Sort

test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/NtlmAuthentationTest.cs
```diff
@@ -11,35 +11,51 @@
 using Microsoft.Extensions.Logging;
 using Xunit;
 using Xunit.Sdk;
```
Tratcher: sort

test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/XunitLogger.cs
```diff
@@ -0,0 +1,52 @@
...
+// See License.txt in the project root for license information
+
+using System;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+using Xunit.Abstractions;
+
+namespace Microsoft.AspNetCore.Testing
+{
+    public class XunitLogger : ILogger, IDisposable
```
Tratcher: Should we add this to Testing?

BrennanConroy:
Disabling is a little different from removing the Standalone project entirely. Shouldn't it still be "ported"

test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/XunitLogger.cs
natemcmaster: Yes. https://github.com/aspnet/Testing/issues/243

natemcmaster:
We don't need separate standalone projects anymore. Standalone vs portable is a publish time decision, not a build decision.

.editorconfig
```diff
@@ -0,0 +1,24 @@
+﻿# EditorConfig is awesome:http://EditorConfig.org
```
Tratcher: @muratg is this something we want in all the projects?

@natemcmaster Which editors understand this? Does it require plugins?

natemcmaster: > Which editors understand this?

VS 2017 (no plugins). 
VS Code with plugin.

It's not yet enforced by csc, but IIUC they are planning to add that.

PR #326, opened 2017-02-14
Bump test projects up to .NET 4.5.2

dougbu:
- aspnet/Testing#248
- xUnit no longer supports .NET 4.5.1
- build tests for desktop .NET only on Windows
  - except Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests; will skip that another way (later)

Also fix line endings in project files; unable to proceeed otherwise


dougbu:
Would like a few eyes on this one because the line endings make things difficult to see and we don't have AppVeyor or Travis coverage.

Tratcher:
What messed up the line endings? Blame?

dougbu:
@natemcmaster's .csproj migration script. I blame `node` of course.

dougbu:
But, `git` shows differences as soon as you clone this repo on Windows ☹️

dougbu:
FYI I just checked and `?w=1` gets rid of the line ending fixup in the /files view of this PR.

natemcmaster:
@dougbu these line ending changes shouldn't have been caused by my migration scripts. What is your version of Git.exe and your line-ending settings? 
`git --version`
`git config --global core.autocrlf`

dougbu:
Why not just fix the line endings?

In any case,
```
> git --version
git version 2.9.0.windows.1
> git config core.autocrlf # can be overridden at any config level so --global may hide real value
true
```

natemcmaster:
The reason I asked is that you seem to be the only one hitting issues with line endings. But it seems your settings are fine so go ahead and commit it.

Tratcher:
@dougbu I pulled this repo and the line endings are fine. However, I also pulled your branch and the line endings are fine there too. You just broke Github?

dougbu:
No idea but this has only happened with a couple of repos and a few files in each. In any case, I'll push this and hope the problem doesn't recur anywhere.

dougbu:
8ae481c

PR #328, opened 2017-02-15
Use .NET 4.5.2 target framework when deploying test sites

dougbu:
- follow-up to 8ae481c
- sites are also built for .NET 4.5.2 in desktop runs


dougbu:
Similar to aspnet/ServerTests#56 pull

dougbu:
55f31a8 checked in because this fixed build break

PR #329, opened 2017-02-15
CI doesn't have all of the Integration repos side-by-side

dougbu:
- likely would have been a benign issue until we started building the solution file


dougbu:
Checking in immediately because this fixes another build break. Worse, it doesn't repro locally for developers that have Hosting checked out beside IISIntegration.

dougbu:
c5c134b

PR #331, opened 2017-02-22
React to xUnit change

cesarbs:
I can't test this locally (tests pass but nothing really appears to happen). Can anyone point me to instructions?

### DIFF
--- a/test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/Helpers.cs
+++ b/test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/Helpers.cs
@@ -10,7 +10,7 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
     {
         public static string GetTestSitesPath(ApplicationType applicationType)
         {
-            return Path.GetFullPath(Path.Combine("..", "..", "..", "..", "..", applicationType == ApplicationType.Standalone ? "TestSites" : "TestSites.Portable"));
+            return Path.GetFullPath(Path.Combine(Directory.GetCurrentDirectory(), "..", applicationType == ApplicationType.Standalone ? "TestSites" : "TestSites.Portable"));
         }
     }
 }


test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/Helpers.cs
```diff
@@ -10,7 +10,7 @@ public class Helpers
     {
         public static string GetTestSitesPath(ApplicationType applicationType)
         {
-            return Path.GetFullPath(Path.Combine("..", "..", "..", "..", "..", applicationType == ApplicationType.Standalone ? "TestSites" : "TestSites.Portable"));
+            return Path.GetFullPath(Path.Combine(Directory.GetCurrentDirectory(), "..", applicationType == ApplicationType.Standalone ? "TestSites" : "TestSites.Portable"));
```
ajaybhargavb: Do you even need the `Directory.GetCurrentDirectory()` here? It was not there before

JunTaoLuo:
We are removing IISIntegration and IISIntegration.Tools in dev but why are we removing it in our patch release?

davidfowl:
> We are removing IISIntegration and IISIntegration.Tools in dev but why are we removing it in our patch release?

Where is IISIntegration going? That's the first I've heard of this? (Tools makes sense).

pranavkm:
Sorry, my bad. We're building both the tool and the source in this particular branch. The tool needs to go away in this branch. I had a brain fart and thought this was the tool.

JunTaoLuo:
@davidfowl I'm confusing IISIntegration with something else.

cesarbs:
I was able to test by changing `ConditionalTheory` to `Theory`. This means our patch branches are being affected by the same test discovery issues we've recently experienced in `dev` after xUnit 2.2 changes. I'm porting our fixes in Testing to that repo's `rel/1.1.0` branch.

cc @natemcmaster

JunTaoLuo:
we'll do what is needed to get the build out but I think it's best we start pinning the xunit versions in the patch branches to avoid these kind of churn.

pranavkm:
The alternative is to pin the version of xunit to an earlier build. The 1.1.0 package referenced `xunit (>= 2.2.0-beta4-build3444)`

cesarbs:
Actually nevermind. This must be an issue local to my machine. The tests still run in the CI (otherwise I wouldn't have sent this PR to fix them 😝).

If anything, let's pin the version of xUnit. I think that's a less dramatic change than porting stuff from Testing.

PR #338, opened 2017-03-20
Auth 2.0 + IISListener

HaoK:
@Tratcher 

Given that ForwardWindows is on by default, not sure making auth optional is going to fly.  All of the tests blew up with Auth missing until I added them all...

### DIFF
--- a/samples/IISSample/IISSample.csproj
+++ b/samples/IISSample/IISSample.csproj
@@ -11,6 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Core" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
--- a/samples/IISSample/IISSample.csproj
+++ b/samples/IISSample/IISSample.csproj
@@ -11,7 +11,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Authentication.Core" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
--- a/samples/IISSample/Startup.cs
+++ b/samples/IISSample/Startup.cs
@@ -1,6 +1,5 @@
-﻿using System;
+using System;
 using System.Linq;
-using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
@@ -13,7 +12,6 @@ namespace IISSample
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddAuthenticationCore();
             // These two middleware are registered via an IStartupFilter in UseIISIntegration but you can configure them here.
             services.Configure<IISOptions>(options =>
             {
--- a/samples/IISSample/Startup.cs
+++ b/samples/IISSample/Startup.cs
@@ -13,7 +13,7 @@ namespace IISSample
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddAuthentication();
+            services.AddAuthenticationCore();
             // These two middleware are registered via an IStartupFilter in UseIISIntegration but you can configure them here.
             services.Configure<IISOptions>(options =>
             {
--- a/samples/IISSample/Startup.cs
+++ b/samples/IISSample/Startup.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
@@ -12,6 +13,7 @@ namespace IISSample
     {
         public void ConfigureServices(IServiceCollection services)
         {
+            services.AddAuthentication();
             // These two middleware are registered via an IStartupFilter in UseIISIntegration but you can configure them here.
             services.Configure<IISOptions>(options =>
             {
--- a/src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
@@ -1,138 +1,99 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Linq;
-using System.Security.Claims;
+using System.Globalization;
+using System.Security.Principal;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Authentication;
-using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.IISIntegration
 {
     internal class AuthenticationHandler : IAuthenticationHandler
     {
-        internal AuthenticationHandler(HttpContext httpContext, IISOptions options, ClaimsPrincipal user)
+        private const string MSAspNetCoreWinAuthToken = "MS-ASPNETCORE-WINAUTHTOKEN";
+
+        internal AuthenticationScheme Scheme { get; private set; }
+
+        public Task<AuthenticateResult> AuthenticateAsync(AuthenticateContext context) 
         {
-            HttpContext = httpContext;
-            User = user;
-            Options = options;
+            var user = GetUser(context.HttpContext);
+            if (user != null)
+            {
+                return Task.FromResult(AuthenticateResult.Success(new AuthenticationTicket(user, Scheme.Name)));
+            }
+            else
+            {
+                return Task.FromResult(AuthenticateResult.None());
+            }
         }
 
-        internal HttpContext HttpContext { get; }
-
-        internal IISOptions Options { get; }
-
-        internal ClaimsPrincipal User { get; }
-
-        internal IAuthenticationHandler PriorHandler { get; set; }
-
-        public Task AuthenticateAsync(AuthenticateContext context)
+        private WindowsPrincipal GetUser(HttpContext httpContext)
         {
-            if (ShouldHandleScheme(context.AuthenticationScheme))
+            var tokenHeader = httpContext.Request.Headers[MSAspNetCoreWinAuthToken];
+
+            int hexHandle;
+            WindowsPrincipal winPrincipal = null;
+            if (!StringValues.IsNullOrEmpty(tokenHeader)
+                && int.TryParse(tokenHeader, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out hexHandle))
             {
-                if (User != null)
-                {
-                    context.Authenticated(User, properties: null, description: null);
-                }
-                else
-                {
-                    context.NotAuthenticated();
-                }
+                // Always create the identity if the handle exists, we need to dispose it so it does not leak.
+                var handle = new IntPtr(hexHandle);
+                var winIdentity = new WindowsIdentity(handle);
+
+                // WindowsIdentity just duplicated the handle so we need to close the original.
+                NativeMethods.CloseHandle(handle);
+
+                httpContext.Response.RegisterForDispose(winIdentity);
+                winPrincipal = new WindowsPrincipal(winIdentity);
             }
 
-            if (PriorHandler != null)
-            {
-                return PriorHandler.AuthenticateAsync(context);
-            }
-
-            return TaskCache.CompletedTask;
+            return winPrincipal;
         }
+
 
         public Task ChallengeAsync(ChallengeContext context)
         {
-            // Some other provider may have already accepted this challenge. Having multiple providers with
-            // AutomaticChallenge = true is considered invalid, but changing the default would breaking
-            // normal Windows auth users.
-            if (!context.Accepted && ShouldHandleScheme(context.AuthenticationScheme))
+            switch (context.Behavior)
             {
-                switch (context.Behavior)
-                {
-                    case ChallengeBehavior.Automatic:
-                        // If there is a principal already, invoke the forbidden code path
-                        if (User == null)
-                        {
-                            goto case ChallengeBehavior.Unauthorized;
-                        }
-                        else
-                        {
-                            goto case ChallengeBehavior.Forbidden;
-                        }
-                    case ChallengeBehavior.Unauthorized:
-                        HttpContext.Response.StatusCode = 401;
-                        // We would normally set the www-authenticate header here, but IIS does that for us.
-                        break;
-                    case ChallengeBehavior.Forbidden:
-                        HttpContext.Response.StatusCode = 403;
-                        break;
-                }
-                context.Accept();
+                case ChallengeBehavior.Automatic:
+                    // If there is a principal already, invoke the forbidden code path
+                    if (GetUser(context.HttpContext) == null)
+                    {
+                        goto case ChallengeBehavior.Unauthorized;
+                    }
+                    else
+                    {
+                        goto case ChallengeBehavior.Forbidden;
+                    }
+                case ChallengeBehavior.Unauthorized:
+                    context.HttpContext.Response.StatusCode = 401;
+                    // We would normally set the www-authenticate header here, but IIS does that for us.
+                    break;
+                case ChallengeBehavior.Forbidden:
+                    context.HttpContext.Response.StatusCode = 403;
+                    break;
             }
-
-            if (PriorHandler != null)
-            {
-                return PriorHandler.ChallengeAsync(context);
-            }
-
             return TaskCache.CompletedTask;
         }
 
-        public void GetDescriptions(DescribeSchemesContext context)
+        public Task InitializeAsync(AuthenticationScheme scheme, HttpContext context)
         {
-            foreach (var description in Options.AuthenticationDescriptions)
-            {
-                context.Accept(description.Items);
-            }
-
-            if (PriorHandler != null)
-            {
-                PriorHandler.GetDescriptions(context);
-            }
+            Scheme = scheme;
+            return TaskCache.CompletedTask;
         }
 
         public Task SignInAsync(SignInContext context)
         {
-            // Not supported, fall through
-            if (PriorHandler != null)
-            {
-                return PriorHandler.SignInAsync(context);
-            }
-
-            return TaskCache.CompletedTask;
+            throw new NotSupportedException();
         }
 
         public Task SignOutAsync(SignOutContext context)
         {
-            // Not supported, fall through
-            if (PriorHandler != null)
-            {
-                return PriorHandler.SignOutAsync(context);
-            }
-
             return TaskCache.CompletedTask;
-        }
-
-        private bool ShouldHandleScheme(string authenticationScheme)
-        {
-            if (Options.AutomaticAuthentication && string.Equals(AuthenticationManager.AutomaticScheme, authenticationScheme, StringComparison.Ordinal))
-            {
-                return true;
-            }
-
-            return Options.AuthenticationDescriptions.Any(description => string.Equals(description.AuthenticationScheme, authenticationScheme, StringComparison.Ordinal));
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
@@ -15,6 +15,7 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
     internal class AuthenticationHandler : IAuthenticationHandler
     {
         private const string MSAspNetCoreWinAuthToken = "MS-ASPNETCORE-WINAUTHTOKEN";
+        private WindowsPrincipal _user;
 
         internal AuthenticationScheme Scheme { get; private set; }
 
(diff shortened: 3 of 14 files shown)


Tratcher:
I really wish we had a way of detecting at startup if/which WindowsAuth schemes were enabled in IIS. @shirhatti, @pan-wang?

src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
```diff
@@ -1,138 +1,99 @@
...
             {
-                if (User != null)
-                {
-                    context.Authenticated(User, properties: null, description: null);
-                }
-                else
-                {
-                    context.NotAuthenticated();
-                }
+                return Task.FromResult(AuthenticateResult.Success(new AuthenticationTicket(user, Scheme.Name)));
```
Tratcher: Note we must not call Merge on this principal, we need to keep the original for IsInRole to work. Same for WebListener.

src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
```diff
@@ -1,138 +1,99 @@
...
+        public Task ChallengeAsync(ChallengeContext context)
         {
-            // Not supported, fall through
-            if (PriorHandler != null)
+            switch (context.Behavior)
             {
-                return PriorHandler.SignInAsync(context);
+                case ChallengeBehavior.Automatic:
+                    // If there is a principal already, invoke the forbidden code path
+                    if (GetUser(context.HttpContext) == null)
```
Tratcher: GetUser needs to cache if this is going to work, it disposes the handle so it can't run more than once.

src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
HaoK: Yuck, ok so that basically means you can't really combine the windows auth with the other auth types or the merging will break things

Tratcher: Yes, but those scenarios don't usually overlap. E.g. you don't do NTLM + Cookies at the same time. So long as we can make Merge smart enough not to merge when there's really just WindowsPrincipal and Anonymous it should be fine.

src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
HaoK: Updated with GetUser caching

HaoK: It only caches on success though, not sure if we expect this to be called repeatedly on failures or if we care...

Tratcher: Should be fine, the failure scenarios are primarily for a missing header which will be fast.

HaoK:
Updated with latests auth changes and packages, tests are all passing now

src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
```diff
@@ -49,6 +47,19 @@ public IISMiddleware(RequestDelegate next, ILoggerFactory loggerFactory, IOption
 
             _next = next;
             _options = options.Value;
+
+            if (_options.ForwardWindowsAuthentication)
+            {
+                var auth = authentication.FirstOrDefault();
+                if (auth == null)
+                {
+                    throw new InvalidOperationException("IServiceCollection.AddAuthentication() is required to use Authentication.");
```
Tratcher: Remove this exception for now, ForwardWindowsAuthentication is enabled by default.

src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
```diff
@@ -49,6 +47,19 @@ public IISMiddleware(RequestDelegate next, ILoggerFactory loggerFactory, IOption
...
+                }
+
+                auth.AddScheme(new AuthenticationScheme("NTLM", /*displayName*/ null,typeof(AuthenticationHandler)));
```
Tratcher: named params instead of inline comments?

src/Microsoft.AspNetCore.Server.IISIntegration/Microsoft.AspNetCore.Server.IISIntegration.csproj
```diff
@@ -11,11 +11,13 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Abstractions" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Http" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.HttpOverrides" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(AspNetCoreVersion)" />
```
Tratcher: revert duplicate

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
```diff
@@ -191,7 +201,7 @@ public void PathBaseHiddenFromServer()
                 {
                     app.Run(context =>
                     {
-                        var auth = context.Features.Get<IHttpAuthenticationFeature>();
+                        var auth = context.RequestServices.GetService<IAuthenticationSchemeProvider>();
```
Tratcher: not equivalent. You should still add the auth service in this case, it's testing the ForwardWindowsAuthentication flag. You may clone this test to make sure everything also works without the auth service.

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/Microsoft.AspNetCore.Server.IISIntegration.Tests.csproj
```diff
@@ -11,6 +11,8 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Server.IISIntegration\Microsoft.AspNetCore.Server.IISIntegration.csproj" />
+    <ProjectReference Include="..\..\..\Security\src\Microsoft.AspNetCore.Authentication\Microsoft.AspNetCore.Authentication.csproj" />
```
Tratcher: revert :-)

test/TestSites/StartupHttpsHelloWorld.cs
```diff
@@ -1,14 +1,21 @@
...
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
 namespace TestSites
 {
     public class StartupHttpsHelloWorld
     {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddAuthenticationCore();
```
Tratcher: Auth shouldn't be needed in this scenario

test/TestSites/StartupHelloWorld.cs
```diff
@@ -1,14 +1,20 @@
...
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
 namespace TestSites
 {
     public class StartupHelloWorld
     {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddAuthenticationCore();
```
Tratcher: Auth shouldn't be needed in this scenario

src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
Tratcher: You made sure this identity never gets merged right?

HaoK: No we'd have to add something to the security helper (or the authorize filter) to special case this somehow.

Tratcher: There normally wouldn't be anything to merge with? Just the anonymous identity?

HaoK: We should revisit the whole merging/building the principal as part of the AuthZ review.  We could add metadata on the scheme as well (a DisallowMerge), which we could use to throw if any authorize policies specified more than one scheme and that flag was set.

HaoK: Sure, but there's nothing today that would prevent someone from listing NTLM + Bearer which would merge, but if you don't think we need to guard against that, I'm fine with leaving things as is for now

Tratcher: Let's make the main scenario work. Anything else is config error :-)

HaoK: Okay so by default we don't have to do anything, merge is only called if the effective policy ends up with multiple schemes requested, which requires a "config error" to have been introduced

HaoK: Main scenario = DefaultAuthenticate/ChallengeScheme = "Windows" correct?

Tratcher: You're registering both Negotiate and NTLM separately right now. The experience would be better if you combine them.

Tratcher: Then you wouldn't have to specify a default

HaoK: Are we calling this Windows too like HttpSys?

test/TestSites/StartupHttpsHelloWorld.cs
HaoK: It is because of the functional tests calling Challenge

HaoK: Ah yes only needed in the Ntlm one.

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
```diff
@@ -162,12 +162,10 @@ public void PathBaseHiddenFromServer()
                     app.Run(async context => 
                     {
                         var auth = context.RequestServices.GetRequiredService<IAuthenticationSchemeProvider>();
-                        var ntlm = await auth.GetSchemeAsync("NTLM");
-                        Assert.NotNull(ntlm);
-                        Assert.Equal("Microsoft.AspNetCore.Server.IISIntegration.AuthenticationHandler", ntlm.HandlerType.FullName);
-                        var negotiate = await auth.GetSchemeAsync("Negotiate");
-                        Assert.NotNull(negotiate);
-                        Assert.Equal("Microsoft.AspNetCore.Server.IISIntegration.AuthenticationHandler", negotiate.HandlerType.FullName);
+                        var windows = await auth.GetSchemeAsync("Windows");
```
Tratcher: File a followup issue to figure out how we can optionally handle granular schemes. This will be more relevant for HttpSys, but the solution will likely apply to both.

HaoK: https://github.com/aspnet/Security/issues/1189

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
HaoK: Updated test to verify handler only added when forward = true, also added a variant that ensures things don't blow up without auth services.

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
```diff
@@ -24,6 +25,7 @@ public class IISMiddlewareTests
                 .UseSetting("PORT", "12345")
                 .UseSetting("APPL_PATH", "/")
                 .UseIISIntegration()
+                .ConfigureServices(services => services.AddAuthentication())
```
Tratcher: Only add this to tests that need it?

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
```diff
@@ -150,15 +156,62 @@ public void PathBaseHiddenFromServer()
...
+                })
+                .Configure(app =>
+                {
+                    app.Run(async context => 
+                    {
+                        var auth = context.RequestServices.GetService<IAuthenticationSchemeProvider>();
                         Assert.NotNull(auth);
-                        Assert.Equal("Microsoft.AspNetCore.Server.IISIntegration.AuthenticationHandler", auth.Handler.GetType().FullName);
+                        var windowsAuth = await auth.GetSchemeAsync("Windows");
+                        if (forward)
```
Tratcher: else assert null? or would GetSchemeAsync throw?

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
```diff
@@ -150,15 +156,62 @@ public void PathBaseHiddenFromServer()
                 .UseSetting("PORT", "12345")
                 .UseSetting("APPL_PATH", "/")
                 .UseIISIntegration()
+                .ConfigureServices(services => services.AddAuthentication())
```
Tratcher: make a copy of this test without AddAuthentication and test auth.GetSchemeAsync("Windows"); returns null.

test/TestSites/StartupNtlmAuthentication.cs
```diff
@@ -53,18 +59,18 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
...
+                    return context.ForbidAsync();
                 }
 
                 if (context.Request.Path.Equals("/AutoForbid"))
                 {
-                    return context.Authentication.ChallengeAsync();
+                    return context.ChallengeAsync();
                 }
 
                 if (context.Request.Path.Equals("/RestrictedNegotiate"))
```
Tratcher: You can remove RestrictedNegotiate, it's not used in the tests.
https://github.com/aspnet/IISIntegration/blob/e2537e1d96f4a74c8e16d067b21eb1fc6455643e/test/IISIntegration.FunctionalTests/NtlmAuthentationTest.cs#L97-L102

Tratcher (approved):

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
HaoK: Yup I forgot to add the else check null, thanks :)

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
HaoK: Well we can't even call GetScheme without AddAuthentication as the AuthSchemeProvider, we already have tests that verify that the server can start without AddAuthentication.

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
HaoK: Yup, done

samples/IISSample/Startup.cs
```diff
@@ -12,6 +13,7 @@ public class Startup
     {
         public void ConfigureServices(IServiceCollection services)
         {
+            services.AddAuthentication();
```
Tratcher: You also need to add the auth middleware if you want HttpContext.User to get set without calling Authenticate, right? E.g. this sample doesn't work right now does it? Oh, the auth middelware is in the security repo, you can't reference it here can you?

HttpSysServer doesn't have this problem because it sets User before the app is invoked. What if the IISMiddleware always set the User? That code in GetUser should be run on every request regardless to clean up handles.

Tratcher (changes requested):

samples/IISSample/Startup.cs
HaoK: Didn't try running the samples, perhaps mistakenly I assumed functional tests were a superset, urgh yeah forgot we can't depend on security in this repo :/

Tratcher: Windows auth doesn't really need the AuthN middleware, it could do it itself / resolve and invoke the service directly.

HaoK:
Updated as we discussed, if ForwardWindowsAuth, add Windows scheme, and always set it to httpContext.User if found.

src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
```diff
@@ -80,80 +85,14 @@ public IISMiddleware(RequestDelegate next, ILoggerFactory loggerFactory, IOption
...
             {
-                var winPrincipal = UpdateUser(httpContext);
-                var handler = new AuthenticationHandler(httpContext, _options, winPrincipal);
-                AttachAuthenticationHandler(handler);
-                try
-                {
-                    await _next(httpContext);
-                }
-                finally
+                var result = await httpContext.AuthenticateAsync("Windows");
```
Tratcher: Constant for Windows. Public static readonly?

test/IISIntegration.FunctionalTests/NtlmAuthentationTest.cs
```diff
@@ -76,6 +76,7 @@ public Task NtlmAuthentication(RuntimeArchitecture architecture, ApplicationType
                         Assert.Equal(HttpStatusCode.OK, response.StatusCode);
                         Assert.Equal("Anonymous?True", responseText);
 
+                        /* Disabled for now
```
Tratcher: Move the skip url here

samples/IISSample/Startup.cs
```diff
@@ -12,6 +13,7 @@ public class Startup
     {
         public void ConfigureServices(IServiceCollection services)
         {
+            services.AddAuthenticationCore();
```
Tratcher: This can be removed from samples and tests now that it's added by the main component.

test/TestSites/TestSites.csproj
```diff
@@ -12,6 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.AspNetCoreModule" Version="$(AspNetCoreModuleVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Core" Version="$(AspNetCoreVersion)" />
```
Tratcher: extra?

samples/IISSample/IISSample.csproj
```diff
@@ -11,6 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Core" Version="$(AspNetCoreVersion)" />
```
Tratcher: extra?

HaoK:
Updated

src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
HaoK: IISMiddleware.AuthenticationScheme

src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
```diff
@@ -18,6 +18,8 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
 {
     public class IISMiddleware
     {
+        public static readonly string AuthenticationScheme = "Windows";
```
Tratcher: Do you want to follow the pattern from Security? E.g. IISDefaults.AuthenticationScheme? 
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Google/GoogleDefaults.cs#L11
Note those are public const string so they can be used in attributes.

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
```diff
@@ -152,13 +153,62 @@ public void PathBaseHiddenFromServer()
...
+                    {
+                        var auth = context.RequestServices.GetService<IAuthenticationSchemeProvider>();
                         Assert.NotNull(auth);
-                        Assert.Equal("Microsoft.AspNetCore.Server.IISIntegration.AuthenticationHandler", auth.Handler.GetType().FullName);
+                        var windowsAuth = await auth.GetSchemeAsync(IISMiddleware.AuthenticationScheme);
+                        if (forward)
+                        {
+                            Assert.NotNull(windowsAuth);
+                            Assert.Null(windowsAuth.DisplayName);
+                            Assert.Equal("AuthenticationHandler", windowsAuth.HandlerType.Name);
```
Tratcher: Vague?

test/Microsoft.AspNetCore.Server.IISIntegration.Tests/IISMiddlewareTests.cs
```diff
@@ -170,8 +220,10 @@ public void PathBaseHiddenFromServer()
             Assert.True(assertsExecuted);
         }
 
-        [Fact]
-        public async Task DoesNotAddAuthenticationHandlerIfWindowsAuthDisabled()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task DoesNotBlowUpWithoutAuth(bool forward)
```
Tratcher: Not relevant anymore? Auth is always added.

Tratcher (approved):

src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
HaoK: Sure IISDefaults.AuthenticationScheme, and HttpSysDefaults.AuthenticationScheme for the other repo?

Tratcher: Sure.

Tratcher:
You'll need to update ServerTests too.

HaoK:
8ed21d56c81e2ffea3af6ac88f6bba97c286d83d

PR #342, opened 2017-03-23
Try renaming functional tests as a potential mitigation for long path…

pranavkm:
… issues


pranavkm:
Think we're running in to https://github.com/Microsoft/msbuild/issues/1786#issuecomment-288799074 on the CI. This is a weak workaround for it.

pranavkm:
Yeah. We were seeing it in Diagnostics too. If we continue seeing this, we can try tweaking the build agent working dirs.

PR #345, opened 2017-03-24
Fix build

JunTaoLuo:
Launch tests in CoreCLR or CLR where appropriate

Merging immediately to fix build. These fixes are ugly and I'll update once I figure out why `SkipIfCurrentRuntimeIsCoreClrAttribute` doesn't seem to work as advertised.

FYI @jbagga


test/FunctionalTests/HttpsTest.cs
```diff
@@ -51,8 +63,8 @@ public Task Https_HelloWorld(RuntimeFlavor runtimeFlavor, RuntimeArchitecture ar
                 using (var deployer = ApplicationDeployerFactory.Create(deploymentParameters, logger))
                 {
                     var deploymentResult = deployer.Deploy();
-                    var handler = new WebRequestHandler();
-                    handler.ServerCertificateValidationCallback = (a, b, c, d) => true;
+                    var handler = new HttpClientHandler();
```
JunTaoLuo: We shouldn't need to do this but it seems like the `WebRequestHandler` on net46 has issues. Instead we are launching the affected test using `HttpClientHandler`. However, this requires us to use the latest 4.3.1 package while launching from CoreCLR since older versions does not contain the required callback property.

PR #349, opened 2017-03-30
react to aspnet/Hosting#996

anurse:
* Also, moved FunctionalTests to "Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests". This name was already in use in some places and VS kept adding it back in various places so I decided to get everything consistent.

Follow up from aspnet/Hosting#996


test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/HttpsTest.cs
```diff
@@ -111,7 +119,7 @@ public Task Https_HelloWorld_NoClientCert_CoreCLR(RuntimeFlavor runtimeFlavor, R
         [ConditionalTheory(Skip = "Manual test only, selecting a client cert is non-determanistic on different machines.")]
         [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
         [FrameworkSkipCondition(RuntimeFrameworks.CoreCLR)]
-        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, "https://localhost:44401/", ApplicationType.Portable)]
+        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, "https://localhost:44301/", ApplicationType.Portable)]
```
mikeharder: Should this use a dynamic port?

Tratcher: https can't be completely dynamic, IIS Express is pre-configured for https only on a specific port range.

IISIntegration.sln
```diff
@@ -18,12 +18,20 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IISSample", "samples\IISSam
...
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{7E80C58E-9CC8-450C-8A8D-94FC76428150}"
+	ProjectSection(SolutionItems) = preProject
+		build\common.props = build\common.props
+		build\dependencies.props = build\dependencies.props
+		build\Key.snk = build\Key.snk
+		build\repo.props = build\repo.props
+	EndProjectSection
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests", "test\Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests\Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.csproj", "{0DA78C79-7676-4C2F-B161-0481679F5222}"
```
Tratcher: @pranavkm did this due to long path issues. Was that fixed?

pranavkm: Nope. Please keep shorter names

pranavkm: https://github.com/Microsoft/msbuild/issues/1786

test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/HttpsTest.cs
anurse: What @Tratcher said. There's a comment at the top of the class, which I can clarify. SSL bindings require pre-registering with http.sys, which includes specifying a port. Conveniently, IIS Express does this for us automatically for the `443xx` range of ports. We could use a random port allocation strategy here, or build some custom code to register the SSL binding, but it would be a lot of overhead.

IISIntegration.sln
anurse: If only SLN files could have comments ;P

test/FunctionalTests/HelloWorldTest.cs
```diff
@@ -9,68 +9,74 @@
...
+        }
+
         [ConditionalTheory]
         [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
         [FrameworkSkipCondition(RuntimeFrameworks.CoreCLR)]
-        //[InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x86, "http://localhost:5061/", ServerType.Kestrel, ApplicationType.Portable)]
-        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, "http://localhost:5062/", ServerType.Kestrel, ApplicationType.Portable)]
-        public Task HelloWorld_IISExpress(RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, string applicationBaseUrl, ServerType delegateServer, ApplicationType applicationType)
+        //[InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x86, ServerType.Kestrel, ApplicationType.Portable)]
+        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, ServerType.Kestrel, ApplicationType.Portable)]
```
pranavkm: Do you need the inline data at all? It kinda looks misleading since it says `ServerType.Kestrel` but runs the test for IISEx

anurse: All of these tests have multiple `InlineData` attributes, even though all but one are commented out. I'd rather leave them as-is since there are issues to re-enable some of them.

However, the `ServerType.Kestrel` is totally vestigial, I'll clean that up.

test/Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests/HttpsTest.cs
anurse: Not to mention we'd need self-signed certs for every possible port, or to generate a self-signed cert on the fly.

Issue #352, opened 2017-04-06
UseIISIntegration should set IServerAddresses.PreferHostingUrls


Tratcher:
When running behind IIS ANCM should override any Kestrel endpoint configuration by setting PreferHostingUrls
https://github.com/aspnet/IISIntegration/blob/c420addfca0b3ea421cfb2131cae28c4ac556a80/src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs#L54

This can clean up some ugliness from Program.Main:
https://github.com/aspnet/Security/blob/2d7cd6038fea34b84084685648ac9d0c97be610b/samples/SocialSample/Program.cs#L19

See https://github.com/aspnet/Hosting/pull/992
@JunTaoLuo

shirhatti:
Especially given the fact we are injecting UseIISIntegration, this makes me a little uncomfortable. Anything we can do to improve discoverability around this behavior?

JunTaoLuo:
@shirhatti The servers will log if server addresses are used to override the endpoint configurations. We can have UseIISIntegration log that it's setting the PreferHostingUrls flag. Other than that, I can only think of updating docs to make this behaviour more clear.

PR #355, opened 2017-04-09
Set PreferHostingUrls when IIS adding override address

JunTaoLuo:
#352

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs
```diff
@@ -52,6 +52,7 @@ public static IWebHostBuilder UseIISIntegration(this IWebHostBuilder hostBuilder
                 {
                     // Delay register the url so users don't accidently overwrite it.
                     hostBuilder.UseSetting(WebHostDefaults.ServerUrlsKey, address);
+                    hostBuilder.PreferHostingUrls(true);
```
davidfowl: When did we decide to default this to on?

Tratcher: It's only on by default for ANCM. This allows you to transparently switch from Console to IIS.

Tratcher: It gets rid of this problem:
https://github.com/aspnet/Security/blob/dev/samples/SocialSample/Program.cs#L19

davidfowl: Why not just change that template?

Tratcher: It wouldn't belong in the template. You only want it active when the ANCM environment variables trigger it. Otherwise it no-op.

Also, we're taking UseIISIntegration out of the template.

JunTaoLuo: This is only for when you run behind ANCM and starting the app on localhost using the port specified `ASPNETCORE_PORT`. We don't want the default template to PreferHostingUrls. We need to use this default since UseIISIntegration is expected to be injected via the new hosting assembly startup feature.

davidfowl: I'm not talking about all templates, I'm talking about that specific sample.

PR #356, opened 2017-04-10
Add osx.10.12-x64 as a RID

pranavkm:


BrennanConroy:
:shipit:  there might be a few other repos we'll need to do this

PR #376, opened 2017-06-05
Clear MaxRequestBodySize

Tratcher:
IIS already limits the request body size so this avoids having to configure it both in IIS and Kestrel.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
@@ -76,6 +76,13 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
                 return;
             }
 
+            var bodySizeFeature = httpContext.Features.Get<IHttpMaxRequestBodySizeFeature>();
+            if (bodySizeFeature != null)
+            {
+                // IIS already limits this, no need to do it twice.
+                bodySizeFeature.MaxRequestBodySize = null;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
```diff
@@ -76,6 +76,13 @@ public IISMiddleware(RequestDelegate next, ILoggerFactory loggerFactory, IOption
                 return;
             }
 
+            var bodySizeFeature = httpContext.Features.Get<IHttpMaxRequestBodySizeFeature>();
+            if (bodySizeFeature != null)
+            {
+                // IIS already limits this, no need to do it twice.
+                bodySizeFeature.MaxRequestBodySize = null;
```
halter73: Should we log a warning? It might help people figure out why the limit they configured isn't having any effect.

Tratcher: No, kestrel is going to turn this on by default so you would get a warning on every request. I agree the discoverablity is lacking. What if kestrel detected iis and logged one warning at startup if someone had changed the default?

Tratcher: But I guess at that point we'd have Kestrel just not set MaxRequestBodySize when running under IIS and we wouldn't need this PR.

Issue #371, opened 2017-05-18
Only register the auth handler based on the environment variable


Tratcher:
https://github.com/aspnet/AspNetCoreModule/issues/82
ANCM now has an environment variable (ASPNETCORE_IIS_HTTPAUTH=Anonymous;Basic;Windows or None.) indicating if windows auth is enabled. Only register the auth handler if windows or basic are enabled, or if the variable is not found (back compat).

This will make the default-to-only-handler more useful by not registering an unneeded handler (e.g. default to Bearer). This will also make the behavior between IIS and HttpSys more consistent. Users were often confused when calling Challenge without an auth middleware caused a 401 on IIS, but threw an InvalidOperationException on Kestrel.

Tratcher:
@muratg we should do this for preview3

mikaelm12:
Can this be removed from the Cookies sample now https://github.com/aspnet/Security/blob/dev/samples/CookieSample/Startup.cs#L17-L21 @Tratcher

Tratcher:
True, but the sample is more helpful if we leave it in.

PR #377, opened 2017-06-05
Conditionally register auth scheme base on ANCM variable

Tratcher:
#371


src/Microsoft.AspNetCore.Server.IISIntegration/IISDefaults.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
 {
     public class IISDefaults
     {
+        public static readonly string AuthenticationScheme = "Windows";
```
HaoK: Are we going to change this in the future?  Just asking why this is diff than Negotiate/Ntlm

Tratcher: https://github.com/aspnet/Security/issues/1189

PR #383, opened 2017-06-09
Re-enable desktop tests

pranavkm:
* Remove FrameworkSkipCondition to enable desktop tests.
* Remove ConditionalTheory \ ConditionalFact since it's no longer required.
* Replace InlineData with meaningful test names
* Cleanup commented out lines of code


test/IISIntegration.FunctionalTests/NtlmAuthentationTest.cs
```diff
@@ -22,29 +21,23 @@ public NtlmAuthenticationTests(ITestOutputHelper output) : base(output)
         {
         }
 
-        [ConditionalTheory]
-        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
-        [FrameworkSkipCondition(RuntimeFrameworks.CoreCLR)]
-        [InlineData(RuntimeFlavor.Clr, RuntimeArchitecture.x64, ApplicationType.Portable)]
-        public Task NtlmAuthentication(RuntimeFlavor runtimeFlavor, RuntimeArchitecture architecture, ApplicationType applicationType)
```
pranavkm: None of the NTLM tests are set up to run today. When enabled, they fail on my box:
```
Error Message:
 Assert.Equal() Failure
Expected: Unauthorized
Actual:   OK
Stack Trace:
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.NtlmAuthenticationTests.<NtlmAuthentication>d__3.MoveNext() in D:\work\IISIntegration\test\IISIntegration.FunctionalTests\NtlmAuthentationTest.cs:line 123
```

test/IISIntegration.FunctionalTests/HelloWorldTest.cs
```diff
@@ -20,31 +19,23 @@ public HelloWorldTests(ITestOutputHelper output) : base(output)
         {
         }
 
-        [ConditionalTheory]
-        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
```
Tratcher: Are you going to run these on Linux/Mac now? That won't go well. You could promote this condition to the assembly level like we did for HttpSysServer: https://github.com/aspnet/HttpSysServer/blob/dev/test/Microsoft.AspNetCore.Server.HttpSys.FunctionalTests/Properties/AssemblyInfo.cs

test/IISIntegration.FunctionalTests/HelloWorldTest.cs
```diff
@@ -20,31 +19,23 @@ public HelloWorldTests(ITestOutputHelper output) : base(output)
...
-        [FrameworkSkipCondition(RuntimeFrameworks.CoreCLR)]
```
Tratcher: Because the framework we're actually testing runs out of process, It doesn't matter as much which framework is used to run the test. These skips were added to keep the same test from executing twice, one for each runner framework.

test/IISIntegration.FunctionalTests/NtlmAuthentationTest.cs
Tratcher: https://github.com/aspnet/ServerTests/issues/82. These tests must be run with the CLR runner, not the CoreCLR runner due to issues with HttpClient.

test/IISIntegration.FunctionalTests/HelloWorldTest.cs
pranavkm: These get disabled via KoreBuild: https://github.com/aspnet/IISIntegration/blob/dev/build/repo.props#L4. That said, this will cause the test runs to fail if you explicitly call `dotnet test` on this from a xplat machine, but that seems uncommon.

test/IISIntegration.FunctionalTests/NtlmAuthentationTest.cs
pranavkm: Gotcha. Could I use `#ifdef` instead so we get mutually exclusive sets of tests - things that run using CoreCLR and things that run using CLR?

test/IISIntegration.FunctionalTests/HelloWorldTest.cs
Tratcher: Doesn't that prevent the tests from even compiling cross-plat?

pranavkm: Nope, just running. We build the sln (https://github.com/aspnet/KoreBuild/blob/dev/build/targets/KoreBuild.SolutionItems.targets#L7), but look for tests to run (https://github.com/aspnet/KoreBuild/blob/dev/build/targets/KoreBuild.SolutionItems.targets#L16-L18).

pranavkm:
🆙 📅 

```
NETCOREAPP2.0:
Total tests: 8. Passed: 6. Failed: 0. Skipped: 2.

NET461:
Total tests: 2. Passed: 2. Failed: 0. Skipped: 0.
```

test/IISIntegration.FunctionalTests/NtlmAuthentationTest.cs
```diff
@@ -117,10 +115,6 @@ Task NtlmAuthentication_CoreCLR(RuntimeFlavor runtimeFlavor, RuntimeArchitecture
                         responseText = await response.Content.ReadAsStringAsync();
                         Assert.Equal(HttpStatusCode.OK, response.StatusCode);
                         Assert.NotEmpty(responseText);
-
-                        response = await httpClient.GetAsync("/AutoForbid");
-                        responseText = await response.Content.ReadAsStringAsync();
-                        Assert.Equal(HttpStatusCode.Forbidden, response.StatusCode);
```
Tratcher: Autoforbid can also be removed from the test site: https://github.com/aspnet/IISIntegration/blob/c267d192d682b2c22bd38e0c16df21157b10a579/test/TestSites/StartupNtlmAuthentication.cs#L64

Issue #391, opened 2017-07-03
Add an option allowing to configure the scheme/display name associated with the Windows scheme


PinpointTownes:
`IISMiddleware` registers the `Windows` scheme using the hardcoded `null` display name. Yet, there are cases where you want to specify a display name so the scheme appears in the external schemes list.

Please add an option to make the display name configurable (and maybe the scheme name too).

https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs#L60

**Note: this also applies to WebListener.**

/cc @Tratcher @HaoK

PinpointTownes:
@Tratcher not having this option will be blocking for me for at least 3 projects. Is it too late to send a PR adding a `DisplayName` property to `IISOptions`? (and optionally, an `AuthenticationScheme` property)

Tratcher:
If you're blocked go ahead and send the PR. We'll see if we can get it approved.

Tratcher:
rel/2.0.0 branch

PinpointTownes:
Okay, will do that tonight.

Tratcher:
Tentatively approved, I'll do the PR today.

Tratcher:
Let me know if you have a compelling reason to set the auth scheme, but not for 2.0.

PinpointTownes:
@Tratcher haha no, I can live with the scheme not being configurable :trollface: 

Do you plan to port the fix to WebListener?

Tratcher:
Not sure. It wasn't configurable before, and we still have an open item (that I can't find at the moment) about making WebListener work with individual schemes like it used to. Any plans to use WebListener?

Issue #390, opened 2017-07-03
IISMiddleware shouldn't do automagical authentication


PinpointTownes:
Another oddity I discovered when torturing my ASOS port: `IISMiddleware` automagically populates `HttpContext.User` for every request (and you can't even disable it).

It's extremely weird, 'cause it's something that is already done by `AuthenticationMiddleware`, so there's really no reason to do that again in the IIS integration stuff.

/cc @Tratcher @HaoK

Tratcher:
It has always done this and so has WebListener, disabling it now would cause regressions. The hosts have been treated a bit different in that the identity is available for you from the very beginning of the request. You don't even need the auth middleware for this host auth to work (in most cases).

PinpointTownes:
> It has always done this and so has WebListener

With a **fundamental** difference: you could disable it in 1.x: https://github.com/aspnet/IISIntegration/blob/1.0.0/src/Microsoft.AspNetCore.Server.IISIntegration/IISOptions.cs#L17.

... in 2.0, you can't.

And if it did that, it was because of the inherent design of the authentication middleware, that all used to set `HttpContext.User` when active authentication was enabled.

Now that this concept is gone, there's no reason to be inconsistent here and treat the "host" handlers differently than the other ones. Specially since it leads to the fact that 2 handlers can set `HttpContext.User` for the same request, which was the major reason for changing the authentication design in 2.0.

PinpointTownes:
@Tratcher any chance this could be fixed in 2.0? It would surely be a blocker for me, as I only use WIA with passive authentication (it's way too risky to enable automatic authentication).

Tratcher:
We're almost out of time for 2.0. We looked at this today and decided requiring the auth middleware for windows auth to work wasn't a regression we're willing to make.

At best we could put AutomaticAuthentication back to toggle setting HttpContext.User, but it's worth noting WebListener never had that option.

PinpointTownes:
Then I'll boycott 2.0 for apps using WIA as the non-primary authentication mode (typically, any ASOS-based authorization server allowing users to connect via WIA or another provider).

I'll never take the risk of seeing a XSRF attack made possible in an API controller because of this insane design decision. You realize that it will set `HttpContext.User` even if you configure a totally different scheme in `DefaultAuthenticateScheme`, right?

> At best we could put AutomaticAuthentication back to toggle setting HttpContext.User, but it's worth noting WebListener never had that option.

[...]

PR #396, opened 2017-07-07
Add back AutomaticAuthentication for opt-out.

Tratcher:
#390 @PinpointTownes
Reverts an un-needed breaking change.


PinpointTownes:
@Tratcher did you chat with @davidfowl? We discussed this pain point on Slack and I suggested another approach, that would potentially make things look way better.

PinpointTownes:
While you're fixing that, you may also want to fix https://github.com/aspnet/IISIntegration/issues/391, which is also a regression.

Tratcher:
Offline: This is the lowest risk change we can make at the moment.

PinpointTownes:
Less than ideal but better than nothing, I suppose. What about the other issue?

Tratcher:
Yes I expect we'll fix #391, but I don't know how soon.

PR #397, opened 2017-07-07
Specify scheme for tests

HaoK:
@Tratcher

### DIFF
--- a/test/TestSites/StartupNtlmAuthentication.cs
+++ b/test/TestSites/StartupNtlmAuthentication.cs
@@ -52,13 +52,13 @@ namespace TestSites
                     }
                     else
                     {
-                        return context.ChallengeAsync();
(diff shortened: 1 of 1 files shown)


test/TestSites/StartupNtlmAuthentication.cs
```diff
@@ -52,13 +52,13 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
...
                     else
                     {
-                        return context.ChallengeAsync();
+                        return context.ChallengeAsync(IISDefaults.AuthenticationScheme);
                     }
                 }
 
                 if (context.Request.Path.Equals("/Forbidden"))
                 {
-                    return context.ForbidAsync();
```
Tratcher: Should have set the default.

PR #400, opened 2017-07-13
Add back the DisplayName setting

Tratcher:
#391 @PinpointTownes @DamianEdwards

### DIFF
--- a/samples/IISSample/Startup.cs
+++ b/samples/IISSample/Startup.cs
@@ -1,8 +1,10 @@
 using System;
 using System.Linq;
+using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Server.IISIntegration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
@@ -15,13 +17,14 @@ namespace IISSample
             // These two middleware are registered via an IStartupFilter in UseIISIntegration but you can configure them here.
             services.Configure<IISOptions>(options =>
             {
+                options.AuthenticationDisplayName = "Windows Auth";
             });
             services.Configure<ForwardedHeadersOptions>(options =>
             {
             });
         }
 
-        public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
+        public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory, IAuthenticationSchemeProvider authSchemeProvider)
         {
             var logger = loggerfactory.CreateLogger("Requests");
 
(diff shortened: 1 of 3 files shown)


samples/IISSample/Startup.cs
```diff
@@ -15,13 +17,14 @@ public void ConfigureServices(IServiceCollection services)
             // These two middleware are registered via an IStartupFilter in UseIISIntegration but you can configure them here.
             services.Configure<IISOptions>(options =>
             {
+                options.AuthenticationDisplayName = "Windows Auth";
```
HaoK: Should just default this to the scheme like everything else to be consistent: IISDefaults.AuthenticationScheme.  We don't suffix anything else with Auth today

PinpointTownes: Not sure about that. Do you want the Windows scheme to be included by default in the "providers list"?

Pretty sure this has already been discussed in the past.

Tratcher: Yeah, it's null to avoid showing up by default. This is just the sample.

PinpointTownes:
@Tratcher thanks!

Did you consider directly exposing `AuthenticationSchemeBuilder` so you could also change the scheme name?

Something like:

```csharp
public AuthenticationSchemeBuilder Scheme { get; } = new AuthenticationSchemeBuilder
{
    HandlerType = typeof(AuthenticationHandler),
    Name = IISDefaults.AuthenticationScheme
};
```

```csharp
services.Configure<IISOptions>(options =>
{
    options.Scheme.DisplayName = "Intranet";
});
```

HaoK:
That seems more dangerous, I don't think we want to allow people to change the HandlerType at all

Tratcher:
That's interesting, bug I agree exposing the handler would be problematic.

HaoK:
I don't think making the scheme name configurable is that critical, but if we want to expose that as well, its harmless and matches what Security auth does

PinpointTownes:
Well, if that's your only concern, you still can add a check to ensure the type wasn't changed before calling `Buid()`:

```csharp
if (!typeof(AuthenticationHandler).IsAssignableFrom(_options.Scheme.HandlerType))
{
    throw new InvalidOperationException("You're doing something wrong, bro'.");
}
```

HaoK:
That just seems like a more error prone alternative to exposing only AuthenticationScheme / DisplayName

Tratcher:
The structure would be nice..

PinpointTownes:
Sure. But that was a clever way to save one property while adopting the same pattern as the one you've used for `CookieBuilder`... :trollface:

HaoK:
Yeah but unlike CookieBuilder, the Handler IS the main thing, so throwing when you set that defeats the whole purpose of exposing that.

HaoK:
I would go minimal, we can always add the builder later if things get more complicated.

PR #417, opened 2017-09-19
Clear HttpContext.User when the user is disposed.

Tratcher:
https://github.com/aspnet/Logging/issues/543#issuecomment-321907828

Hosting logs between OnCompleted/RegisterForDispose and DisposeContext. Loggers using IHttpContextAccessor can access User and get ODEs. We can't dispose the user any later, but we can clear it once it's been disposed.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


cesarbs:
Can you add a test?

Tratcher:
Not readily, there are too many layers of the stack involved.

cesarbs:
LGTM, assuming you tested this manually. It would be nice to capture the test scenario somewhere so anyone knows how to verify this.

src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs
```diff
@@ -51,13 +52,26 @@ private WindowsPrincipal GetUser()
...
 
             return _user;
         }
 
+        private static Task ClearUser(object arg)
+        {
+            var context = (HttpContext)arg;
+            if (context.User is WindowsPrincipal)
+            {
+                context.User = null;
```
HaoK: Don't we guarantee HttpContext.User to be non null?  Should we just set it to a new ClaimsPrincipal?  Or will the context do that automatically if someone asks for it?

Tratcher: DefaultHttpContext handles that. https://github.com/aspnet/HttpAbstractions/blob/5128efe0e88eaa176d4231ff62d754de59bb0edd/src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs#L131-L143

PR #449, opened 2017-10-13
Adds app-veyor!

jkotalik:
Uses the default one for now.
@Eilon you mentioned that app veyor should at least try to run on the PR even without the .appveyor file, correct?


Eilon:
Yup. And it looks like AppVeyor doesn't like you 😄

jkotalik:
@Eilon seems it is passing now 😄

Issue #438, opened 2017-10-03
Implement Abort in HttpProtocol


jkotalik:
Currently left as a TODO. See https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Core/Internal/Http/HttpProtocol.cs#L406 for the logic needed.

jkotalik:
#453

shirhatti:
Covered by #461

PR #453, opened 2017-10-18
Adds managed to native abort logic 

jkotalik:
This PR targets three main things:
- When context.Abort() is called, call NativeMethods.http_abort_request
- Return FINISH_REQUEST rather than CONTINUE in error cases
- Implement RequestAborted (from Kestrel)

The test I added is disabled for now. Probably best if we add ANCM to universe first.


jkotalik:
Address #438

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -19,10 +19,10 @@ public IISHttpContextOfT(PipeFactory pipeFactory, IHttpApplication<TContext> app
             _application = application;
         }
 
-        public override async Task ProcessRequestAsync()
```
davidfowl: hmm I don’t think this makes sense. Should we just flow the exception? Task\<T\> is very hard to cache

jkotalik: Do you want me to rethrow the exception if there was one, and then catch in the HandleRequest method callback?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -195,6 +196,56 @@ public string ReasonPhrase
...
+                // Otherwise, get the abort CTS.  If we have one, which would mean that someone previously
+                // asked for the RequestAborted token, simply return its token.  If we don't,
+                // check to see whether we've already aborted, in which case just return an
+                // already canceled token.  Finally, force a source into existence if we still
+                // don't have one, and return its token.
+                var cts = _abortedCts;
+                return
+                    cts != null ? cts.Token :
+                    (_requestAborted == 1) ? new CancellationToken(true) :
+                    RequestAbortedSource.Token;
```
anurse: Ugh. This is copied from Kestrel I know, but once you get to two levels of ternary operation I start to wonder if `if`/`else` would be better :(. Not gonna make you change it though. Rather have the code look the same for now.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -90,6 +92,7 @@ public IISHttpContextOfT(PipeFactory pipeFactory, IHttpApplication<TContext> app
                     await _readingTask;
                 }
             }
+            return failed ? NativeMethods.REQUEST_NOTIFICATION_STATUS.RQ_NOTIFICATION_FINISH_REQUEST : NativeMethods.REQUEST_NOTIFICATION_STATUS.RQ_NOTIFICATION_CONTINUE;
```
davidfowl: Why do you finish the request if this failed??

jkotalik: Do we want other modules to execute if there was an application failure? Seems intuitive to finish the request if there is a failure.

anurse: > Modules that implement request handling should return RQ_NOTIFICATION_FINISH_REQUEST when an error occurs.

https://msdn.microsoft.com/en-us/library/ms690802(v=vs.90).aspx

Seems like we should be using FINISH_REQUEST according to docs. I'm not sure how error handling modules in IIS plug in though. In theory, we still want those to run.

davidfowl: I see. I'd prefer it if we handled this all in a single place. Right now it's split across IISContextOfT and IISServer (to handle the synchronous case).

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -542,9 +591,29 @@ private static void FreePinnedHeaders(List<GCHandle> pinnedHeaders)
...
         public void Abort()
         {
-            // TODO
+            // Use code has 
+            if (Interlocked.Exchange(ref _requestAborted, 1) == 0)
+            {
+                // call native method to abort the request
+                NativeMethods.http_abort_request(_pHttpContext);
+                Output.Abort(null);
+                CancelRequestAbortedToken();
```
Tratcher: Abort doesn't need to do this, it's redundant with the disconnect listener.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -195,6 +196,56 @@ public string ReasonPhrase
             }
         }
 
+        public CancellationToken RequestAborted
```
Tratcher: Copy from HttpSys instead of Kestrel. This is only half of the feature.

jkotalik: I am hesitant to directly copy HttpSys for the disconnect listener as IIS has an OnClientDisconnect event. I'd take it you would like me to have both the abort and client disconnect changes in this PR?

Tratcher: I'm saying they're separate features. You can remove RequestAborted from this PR as it's not directly related to Abort(), it's only triggered indirectly via OnClientDisconnected.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
jkotalik: We can make the Task return a bool for success or failure? Then all of the RequestNotification returns would be in IISServer.

jkotalik: Would it be better to rethrow the exception instead of returning Task<T>

Directory.Build.props
```diff
@@ -1,10 +1,11 @@
-﻿<Project>
+<Project>
   <Import Project="version.xml" />
 
   <PropertyGroup>
     <Product>Microsoft ASP.NET Core</Product>
     <RepositoryUrl>https://github.com/aspnet/IISIntegration</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
+    <RepositoryRoot>$(MSBuildThisFileDirectory)</RepositoryRoot>
```
davidfowl: ?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -379,7 +376,8 @@ public unsafe void SendResponseHeaders(bool appCompleted)
             fixed (byte* pReasonPhrase = reasonPhraseBytes)
             {
                 // This copies data into the underlying buffer
-                NativeMethods.http_set_response_status_code(_pHttpContext, (ushort)StatusCode, pReasonPhrase);
+                var hr = NativeMethods.http_set_response_status_code(_pHttpContext, (ushort)StatusCode, 0, pReasonPhrase);
+                CheckHResult(hr);
```
davidfowl: What happens if this throws?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -544,7 +542,13 @@ private static void FreePinnedHeaders(List<GCHandle> pinnedHeaders)
 
         public void Abort()
         {
-            // TODO
+            // Use code has
```
davidfowl: This comment isn't complete

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -19,10 +19,10 @@ public IISHttpContextOfT(PipeFactory pipeFactory, IHttpApplication<TContext> app
             _application = application;
         }
 
-        public override async Task ProcessRequestAsync()
+        public override async Task<bool> ProcessRequestAsync()
         {
             var context = default(TContext);
-
+            var failed = false;
```
davidfowl: Don't use the negative state for the boolean.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -107,6 +106,12 @@ private static void CompleteRequest(IISHttpContext context)
             context.Dispose();
         }
 
+        private static NativeMethods.REQUEST_NOTIFICATION_STATUS ConvertRequestCompletionResults(bool failed)
```
davidfowl: This is much cleaner but it should be success instead of failed.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -544,7 +542,13 @@ private static void FreePinnedHeaders(List<GCHandle> pinnedHeaders)
 
         public void Abort()
         {
-            // TODO
+            // Use code has 
+            if (Interlocked.Exchange(ref _requestAborted, 1) == 0)
+            {
+                // call native method to abort the request
+                NativeMethods.http_abort_request(_pHttpContext);
+                Output.Abort(null);
```
davidfowl: nit style: `Output.Abort(ex: null)`

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -544,7 +542,13 @@ private static void FreePinnedHeaders(List<GCHandle> pinnedHeaders)
 
         public void Abort()
         {
-            // TODO
+            // Use code has 
+            if (Interlocked.Exchange(ref _requestAborted, 1) == 0)
+            {
+                // call native method to abort the request
+                NativeMethods.http_abort_request(_pHttpContext);
```
davidfowl: I'm assuming this also returns an HResult that needs to be checked. Kestrel does this in a very low level part of the p-invoke layer so that it's central. The check method is used around every single pinvoke https://github.com/aspnet/KestrelHttpServer/blob/204cf7a1dc9d5f9a999df3f9f288b4b535b3132a/src/Kestrel.Transport.Libuv/Internal/Networking/LibuvFunctions.cs so exceptions don't slip through the cracks. We still need to be careful where we throw to make sure we don't corrupt anything

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -55,6 +55,7 @@ internal abstract partial class IISHttpContext : NativeRequestContext, IDisposab
         protected Task _writingTask;
 
         protected int _requestAborted;
+        protected CancellationTokenSource _abortedCts;
```
davidfowl: Where is this implemented?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: Yeah I think we should do this. We shouldn't throw when it is required that we return a REQUEST_NOTIFICATION_STATUS to IIS.

jkotalik: https://github.com/aspnet/IISIntegration/issues/458 let's make it part of this.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: We should throw I think as this is a fundamental failure. Investigating where the control flow will go.

jkotalik: It would propagate up to whoever called read or write (to user code). Is that okay?

Tratcher: You'd want to control/wrap the exception type in that case. Stream should only throw IOException and InvalidOperationExeption for most scenarios.

jkotalik: Ahhh that makes sense. Would that be done in the stream itself? I will need to verify which exceptions are being thrown via bad hr status codes.

Tratcher: Yes, somewhere up the stack between here and your response stream you'll need to catch this exception and wrap it.

On a related note, HttpSys and Kestrel's response streams do not throw exceptions for most failure scenarios. The caller usually doesn't care and it makes a lot of log noise. You may end up catching this exception at the stream level and suppressing it.

jkotalik: So would this be a good place to catch the exceptions:
https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpRequestBody.cs#L40
and 
https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpResponseBody.cs#L57?
Or would it be better to catch it after the managed to native write.

Tratcher: More like here:
https://github.com/aspnet/IISIntegration/blob/94c895a1bd1b9bba48ce496200fd386322fb6a74/src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs#L249
And this only applies to the response body. You don't call this from the request body.

jkotalik:
Closing until DisconnectListener changes are in ANCM.

PR #454, opened 2017-10-19
[WIP] Migrate to new pipe APIs

pakrym:


jkotalik:
Do we still need to target these external dependencies? https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/Microsoft.AspNetCore.Server.IISIntegration.csproj#L13-L18 otherwise LGTM.

pakrym:
@jkotalik yes, in theory we only need pipelines reference, but version of `System.*` pipelines depends on is different from ones we have on feed so we have to pull them

Issue #427, opened 2017-09-29
Distinguish between Path and PathBase


jkotalik:
Right now we always set the PathBase to empty and the Path to be the whole path. This property needs to be read from IISConfig, either through a native pInvoke to ANCM, by https://msdn.microsoft.com/en-us/library/system.web.hosting.iprocesshostsupportfunctions.getapplicationproperties(v=vs.110).aspx, or by some other means.

PR #457, opened 2017-10-23
Parse Path and PathBase from Virtual Directory.

jkotalik:
For: #427


src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISApplicationConfiguration.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.IISIntegration
```
jkotalik: this is super ugly. I wanted to make a config file that is loaded once and available statically. Not sure what the best pattern is here.

src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs
```diff
@@ -45,8 +45,11 @@ public static IWebHostBuilder UseIISIntegration(this IWebHostBuilder hostBuilder
                 hostBuilder.UseSetting(nameof(UseIISIntegration), "true");
                 hostBuilder.CaptureStartupErrors(true);
 
-                var applicationPath = NativeMethods.http_get_application_full_path();
-                hostBuilder.UseContentRoot(applicationPath);
+                // TODO consider adding a configuration load where all variables needed are loaded from ANCM in one call.
+                var hResult = NativeMethods.http_get_application_paths(out var fullPath, out var virtualPath);
+                IISApplicationConfiguration.SetVirtualPath(virtualPath);
```
davidfowl: Yea, nah. This can be an part of the options that flows into the `IISHttpServer` implementation.

jkotalik:
Make this function in a separate class and add unit tests. - Offline feedback

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -988,5 +981,38 @@ private enum CurrentOperationType
             Write,
             Flush
         }
+
+        private void ExtractPathAndPathBase(string originalRequestPath, int numSegmentsInVirtualPath, out string requestPathBase, out string requestPath)
```
Tratcher: https://github.com/aspnet/HttpAbstractions/blob/87cd79d6fc54bb4abf07c1e380cd7a9498a78612/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs#L54-L59
In fact you may not need to do this here at all. You could do it the way IISIntegration out-of-proc does and just call UsePathBase once at startup.

jkotalik: Yeah that may be the way to go. Let me try to get that working.

jkotalik:
@Tratcher super updated to call UsePathBase().

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISServerSetupFilter.cs
```diff
@@ -0,0 +1,28 @@
...
+        public IISServerSetupFilter(string virtualPath)
+        {
+            _virtualPath = virtualPath;
+        }
+
+        public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
+        {
+            return app =>
+            {
+                app.UsePathBase("/hello");
```
Tratcher: /hello?

jkotalik: oops.

src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs
```diff
@@ -45,11 +45,14 @@ public static IWebHostBuilder UseIISIntegration(this IWebHostBuilder hostBuilder
                 hostBuilder.UseSetting(nameof(UseIISIntegration), "true");
                 hostBuilder.CaptureStartupErrors(true);
 
-                var applicationPath = NativeMethods.http_get_application_full_path();
-                hostBuilder.UseContentRoot(applicationPath);
+                // TODO consider adding a configuration load where all variables needed are loaded from ANCM in one call.
+                var hResult = NativeMethods.http_get_application_paths(out var fullPath, out var virtualPath);
```
davidfowl: Do we care if this fails?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -80,27 +80,17 @@ internal unsafe IISHttpContext(PipeFactory pipeFactory, IntPtr pHttpContext)
...
                 else
                 {
-                    // url: /base/path, prefix: /base/, base: /base, path: /path
-                    // url: /, prefix: /, base: , path: /
-                    PathBase = originalPath.Substring(0, prefix.Length - 1);
-                    Path = originalPath.Substring(prefix.Length - 1);
+                    // Path and pathbase are unescaped by RequestUriBuilder
+                    // The UsePathBase middleware will modify the pathbase and path correctly
+                    PathBase = string.Empty;
+                    Path = originalPath;
```
davidfowl: Is this correct? Does the path properly exclude the path base?

jkotalik: We add the PathBase middleware right after the Server, just like IISIntegration did.  https://github.com/aspnet/IISIntegration/blob/dev/src/Microsoft.AspNetCore.Server.IISIntegration/IISSetupFilter.cs

davidfowl: Maybe I misunderstand how it works but doesn’t that just set the path base? The path shouldn’t include the path base. What I mean by that is that if we go that route, we have to make sure that the way we’re getting the url from the underlying API excludes the path base.

If he path base is /foo and an http request comes in to /foo/bar. The path should be /bar and the path base should be /foo.

Is that the case?

jkotalik: Yes that would be the case. The UsePathBase middleware will change the path and pathbase. https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseMiddleware.cs#L58-L59

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
```
Tratcher: used?

Issue #458, opened 2017-10-24
Make sure all HRESULT values are checked in managed 


jkotalik:
Our error handling after calling any NativeMethod is lacking. We need to add checks after all calls to NativeMethods and check Marshal.GetExceptionForHR().

muratg:
cc @shirhatti who will collect this and related bugs in a uber diagnostics-features list.

PR #462, opened 2017-10-25
[WIP] Mock ANCM for testing purposes, clean up pInvokes.

jkotalik:
https://github.com/aspnet/IISIntegration/issues/458


Tratcher:
Do you have a good grasp on what exceptions are expected to be thrown and who's expected to handle them for each call site?

jkotalik:
https://docs.microsoft.com/en-us/dotnet/framework/interop/how-to-map-hresults-and-exceptions
Looking through now.

jkotalik:
It seems that Marshal.ThrowExceptionFromHR will most of the time return COMExceptions, which is definitely not what we want. We should create our own custom exception to wrap the COM exceptions?

jkotalik:
Or don't use Marshal.GetExecptionFromHR at all?

Tratcher:
Does it at least give you relevant error messages?
I expect context appropriate wrapping. E.g. Stream defines what exceptions it will throw.

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.AspNetCore.Server.IISIntegration
 {
-    internal class NativeMethods
+    internal unsafe class NativeMethods
```
Tratcher: what's unsafe now? you only removed code.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISFunctions.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using Microsoft.AspNetCore.HttpSys.Internal;
+using static Microsoft.AspNetCore.Server.IISIntegration.IISDeleagates;
+
+namespace Microsoft.AspNetCore.Server.IISIntegration
+{
+    internal unsafe interface IISFunctions
```
Tratcher: IIISFunctions? :-) Then DefaultIISFunctions can be IISFunctions.

jkotalik: Nooooooooooooooo that sounds so bad.

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
jkotalik: Ooops.

jkotalik:
Working on tests now :). I'll change the title.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/DefaultIISFunctions.cs
```diff
@@ -0,0 +1,201 @@
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Text;
+using Microsoft.AspNetCore.HttpSys.Internal;
+using static Microsoft.AspNetCore.Server.IISIntegration.IISDeleagates;
```
dallancarr: Typo in namespace. ☺

jkotalik:
Note this is still in progress.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISFunctions.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using Microsoft.AspNetCore.HttpSys.Internal;
+using static Microsoft.AspNetCore.Server.IISIntegration.IISDelegates;
+
+namespace Microsoft.AspNetCore.Server.IISIntegration
+{
+    internal unsafe interface IISFunctions
+    {
+        void IISPostCompletion(IntPtr pHttpContext, int cbBytes);
```
davidfowl: Since these are a bunch of IIS functions you can remove IIS from each method name.

jkotalik:
Closing until mock is complete. There is too much churn currently.

Issue #461, opened 2017-10-25
Implement IHttpLifetimeFeature


jkotalik:
Currently the abort logic only works for managed to native. When native sends a signal for ClientDisconnected, we need to stop reading and writing and cleanup the connection once fired.

Eilon:
@muratg / @jkotalik - anything to do here for preview1?

jkotalik:
Push this to preview2. I have a WIP branch to implement this but I need to do a bit of redesign.

Eilon:
I'll let @muratg do the honors.

muratg:
@jkotalik could you file one more bug to fix crashes? We can keep this one for the feature for managed events etc.

pakrym:
https://github.com/aspnet/IISIntegration/issues/1020

PR #464, opened 2017-10-30
Implements abort and client disconnect

jkotalik:
For #461 and #438


src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -96,15 +96,31 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS OnAsyncCompletion(IntPt
             return NativeMethods.REQUEST_NOTIFICATION_STATUS.RQ_NOTIFICATION_PENDING;
         }
 
-        private static void CompleteRequest(IISHttpContext context)
+        private static bool OnClientDisconnect(IntPtr pvManagedHttpContext)
+        {
+            var context = (IISHttpContext)GCHandle.FromIntPtr(pvManagedHttpContext).Target;
+            // There is a race here between the context cleaning up and notifying the cancellation token.
+            //
```
davidfowl: I’d like to understand this comment

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -19,10 +19,10 @@ public IISHttpContextOfT(PipeFactory pipeFactory, IHttpApplication<TContext> app
             _application = application;
         }
 
-        public override async Task ProcessRequestAsync()
+        public override async Task<bool> ProcessRequestAsync()
```
davidfowl: Is this change related? Why isn’t this another PR?

jkotalik: It is related. If the request is aborted, we would like to return FINISH_REQUEST to the IIS pipeline. Unless you want to go with the logic of always returning continue.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
jkotalik: I'll clean that up. My b.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
davidfowl: Is that what’s happening here? I don’t see it. Where does abort cause success to be false?

jkotalik: Calling abort could throw an exception from user code on writing (or other operations on the context).

davidfowl: What does that have to do with ending the request? The user code can catch the exception.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -96,7 +96,15 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS OnAsyncCompletion(IntPt
...
         }
 
-        private static void CompleteRequest(IISHttpContext context)
+        private static bool OnClientDisconnect(IntPtr pvManagedHttpContext)
+        {
+            var context = (IISHttpContext)GCHandle.FromIntPtr(pvManagedHttpContext).Target;
+            context?.OnClientDisconnect();
+            return true;
+        }
+
```
davidfowl: nit: extra space

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
jkotalik: https://github.com/aspnet/IISIntegration/pull/469

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
jkotalik: I removed that, not sure why it is still showing up.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -938,6 +1003,11 @@ internal void OnAsyncCompletion(int hr, int cbBytes)
             }
         }
 
+        internal void OnClientDisconnect()
+        {
+            Task.Run(() => CancelRequestAbortedToken());
```
davidfowl: Ahh closure. Fix that 😄

jkotalik:
Pending https://github.com/aspnet/AspNetCoreModule/pull/227, which has a few issues that need to be addressed.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -159,6 +160,69 @@ internal unsafe IISHttpContext(PipeFactory pipeFactory, IntPtr pHttpContext)
...
+                return cts;
+            }
+        }
+
+        private void CancelRequestAbortedToken()
+        {
+            try
+            {
+                RequestAbortedSource.Cancel();
+                _abortedCts = null;
```
Tratcher: No reason to null this out. It risks race conditions and null refs

jkotalik:
Will reopen once shim is in.

PR #465, opened 2017-10-31
Remove header serialization in favor of setting IIS struct directly. 

jkotalik:
After discussing with @pan-wang how we write response headers, he mentioned how he thought writing directly to the IIS headers would be preform better than header serialization. I went ahead and made the change and checked the initial performance on localhost, and the performance seemed to be ~5% better. Tomorrow I will do performance tests on azure VMs. The sample app I will use will look like:
```csharp
app.Run(async (context) =>
{
      // Set a bunch of response headers
      // ex context.Response.Headers["test"] = "foo";
      await context.Response.WriteAsync("Hello world");
}
```
Also this removes a dependency on HttpSys, which is all the better 😄


davidfowl:
What requests did you test?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -374,159 +370,16 @@ public unsafe void SendResponseHeaders(bool appCompleted)
...
-                        if (unknownHeaders == null)
-                        {
-                            unknownHeaders = new HttpApiTypes.HTTP_UNKNOWN_HEADER[numUnknownHeaders];
-                            gcHandle = GCHandle.Alloc(unknownHeaders, GCHandleType.Pinned);
-                            pinnedHeaders.Add(gcHandle);
-                            pHttpResponse->Response_V1.Headers.pUnknownHeaders = (HttpApiTypes.HTTP_UNKNOWN_HEADER*)gcHandle.AddrOfPinnedObject();
-                            pHttpResponse->Response_V1.Headers.UnknownHeaderCount = 0; // to remove the iis header for server=...
-                        }
+                var headerNameBytes = Encoding.UTF8.GetBytes(headerPair.Key);
+                var headerValueBytes = Encoding.UTF8.GetBytes(headerPair.Value);
```
Tratcher: Any way to make IIS send multi-line headers rather than squashing them together? E.g.
HeaderName1: Value1
HeaderName1: Value2
...
This is legal, and required by some headers like auth and set-cookie.

jkotalik:
I'll try setting the replace flag to false tomorrow.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: I think we will need to check for multilineheaders here. IIS will write them as a comma separated header. Do we want to remove any existing IIS headers? If not, we can always set fReplace to false.

jkotalik:
Results of perf testing:
Setup: Azure Win server 2016 as server, Ubuntu 16.04 client running wrk.
Ran the app:
```csharp
  app.Run(async (context) =>
            {
                context.Response.ContentType = "text/plain";
                for (var i = 0; i < 100; i++)
                {
                    context.Response.Headers[$"{i}"] = $"{i}";
                }
                await context.Response.WriteAsync("Hello World - " + DateTimeOffset.Now + Environment.NewLine);
            }
```

Results:
Serialized Headers:
Run 1: 5987.67 requests/second
Run 2: 3273.55 requests/second
Run 3: 2284.59 requests/second

Settings IIS Header directly:
Run 1: 11614.19 requests/second
Run 2: 11107.50 requests/second
Run 3: 10899.79 requests/second

Tratcher:
Interesting but exaggerated.

What about a slightly more realistic data set?
```
                for (var i = 0; i < 10; i++)
                {
                    context.Response.Headers[$"HeaderName{i}"] = $"HeaderValueABCDEFGHIJKLMNOPQRSTUVWXYZ{i}";
                }
```

jkotalik:
With your test, I got:
17172 requests/second without serialization
14751 requests/second with serialization.

jkotalik:
@Tratcher Seems the perf is definitely better here and is less error prone. Cool if we get this in?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -374,159 +370,21 @@ public unsafe void SendResponseHeaders(bool appCompleted)
 
             HttpApiTypes.HTTP_RESPONSE_V2* pHttpResponse = NativeMethods.http_get_raw_response(_pHttpContext);
```
Tratcher: No longer used.

jkotalik: I will still leave the native method though it isn't used.

Tratcher: Why?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -372,16 +372,30 @@ public unsafe void SendResponseHeaders(bool appCompleted)
             foreach (var headerPair in HttpResponseHeaders)
             {
                 var headerValues = headerPair.Value;
-                var headerNameBytes = Encoding.UTF8.GetBytes(headerPair.Key);
-
-                for (var i = 0; i < headerValues.Count; i++)
+                var knownHeaderIndex = HttpApiTypes.HTTP_RESPONSE_HEADER_ID.IndexOfKnownHeader(headerPair.Key);
+                if (knownHeaderIndex != -1)
```
Tratcher: Isn't this check backwards?

jkotalik:
@Tratcher @davidfowl _ping_

PR #470, opened 2017-11-03
Use ipv4 when binding to ANCM port

davidfowl:
- Today we use localhost which binds to both the ipv4 and ipv6 addresses.
ANCM never communicates over ipv6 so it's pointless. This should also properly crash
the process if the ipv4 address fails to bind (which we've seen reports in real applications for).


Tratcher:
And the new env var for the full url?

davidfowl:
@Tratcher don't want that to be part of this PR. This is the immediate change that solves most of the issues. Everything else is basically speculative at this point. 

I'd imagine it would be ASPNETCORE_URLS but there's likely a precedence issue as the application can override it.

PR #471, opened 2017-11-08
Adds windows Auth support

jkotalik:


Tratcher:
Needs a few more parts:
1) Only register the auth scheme if windows auth is enabled. You'll need to backtrack how ANCM checks this and sets that env var:
https://github.com/aspnet/IISIntegration/blob/9c065bf1075af2092be468d75dcbb06c6f6dd49b/src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs#L68-L92
2) Actually register the auth scheme: https://github.com/aspnet/IISIntegration/blob/9c065bf1075af2092be468d75dcbb06c6f6dd49b/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs#L58-L61
3) An option to not set HttpContext.User, some people hate that. To maximize compat maybe we should keep using IISOptions and honor all of its settings.
https://github.com/aspnet/IISIntegration/blob/9c065bf1075af2092be468d75dcbb06c6f6dd49b/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs#L115
4) Functional tests. You should be able to directly copy these from out-of-proc.

test/IISIntegration.IISServerFunctionalTests/AuthenticationTests.cs
```diff
@@ -0,0 +1,115 @@
...
+                        return deploymentResult.HttpClient.GetAsync(string.Empty);
+                    }, logger, deploymentResult.HostShutdownToken, retryCount: 30);
+
+                    var responseText = await response.Content.ReadAsStringAsync();
+                    try
+                    {
+                        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                        Assert.Equal("Hello World", responseText);
+
+                        response = await httpClient.GetAsync("/Anonymous");
```
jkotalik: I don't understand the check on Anonymous here. @Tratcher why does this pass for the out of proc stuff?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISServerAuthenticationHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                return Task.FromResult(AuthenticateResult.Success(new AuthenticationTicket(user, Scheme.Name)));
+            }
+            else
+            {
+                return Task.FromResult(AuthenticateResult.NoResult());
+            }
+        }
+
+        private static Task ClearUser(object arg)
```
davidfowl: Nobody uses this

src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs
```diff
@@ -46,7 +46,12 @@ public static IWebHostBuilder UseIISIntegration(this IWebHostBuilder hostBuilder
                 hostBuilder.CaptureStartupErrors(true);
 
                 // TODO consider adding a configuration load where all variables needed are loaded from ANCM in one call.
-                var hResult = NativeMethods.http_get_application_paths(out var fullPath, out var virtualPath);
+                var hResult = NativeMethods.http_get_application_properties(out var fullPath,
```
davidfowl: This doesn't look great. Should we consider looking at using a struct?

korebuild-lock.txt
```diff
@@ -1,2 +1,2 @@
-version:2.1.0-preview1-15552
-commithash:526c2d8d521343e5a29c2a1323925528cb94e873
+version:2.1.0-preview1-15559
+commithash:2c59937fbbe6f2ae5ceb6c5ff5a3bd35577d96ca
```
davidfowl: @natemcmaster When does this happen?

davidfowl: Nobody needs this, so it’s probaly just happening

natemcmaster: yeah, you can probably revert. This change will happen if you delete the file and re-run build.cmd, or if you run `build.cmd -Update`

jkotalik: Yeah I ran build.cmd -u and it pulled in this version. Eventually this will need to happen on every PR that has a corresponding change to ANCM.

natemcmaster: build.cmd -u just updates build tools, not ANCM. To update ANCM, change this: https://github.com/aspnet/IISIntegration/blob/9c065bf1075af2092be468d75dcbb06c6f6dd49b/build/dependencies.props#L7

test/IISIntegration.IISServerFunctionalTests/AuthenticationTests.cs
Tratcher: Those tests are set up for windows auth + anonymous. What happens?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISServerAuthenticationHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Security.Principal;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Server.IISIntegration
+{
+    public class IISServerAuthenticationHandler : IAuthenticationHandler
+    {
+        private static readonly Func<object, Task> ClearUserDelegate = ClearUser;
```
Tratcher: Remove

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -120,6 +123,11 @@ internal unsafe IISHttpContext(PipeFactory pipeFactory, IntPtr pHttpContext)
                 HttpResponseHeaders = new HeaderCollection(); // TODO Optimize for known headers
                 ResponseHeaders = HttpResponseHeaders;
 
+                WindowsUser = GetWindowsPrincipal();
```
Tratcher: WindowsUser needs to get disposed at the end of the request. Make sure it's after all possible app code like https://github.com/aspnet/IISIntegration/pull/471/files#diff-92041534861d73bda639e31d3273ca1dR72

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -827,7 +840,10 @@ protected virtual void Dispose(bool disposing)
                     // TODO: dispose managed state (managed objects).
                     _thisHandle.Free();
                 }
-
+                if (WindowsUser != null)
```
davidfowl: Does this work?

```C#
if (WindowsUser?.Identity is WindowsIdentity wi)
{
  wi.Dispose();
}
```

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -851,5 +867,22 @@ private enum CurrentOperationType
...
         }
+
+        private WindowsPrincipal GetWindowsPrincipal()
+        {
+            var hr = NativeMethods.http_get_authentication_primary_token(_pHttpContext, out IntPtr token);
+            if (hr == 0 && token != IntPtr.Zero)
+            {
+                var identity = new WindowsIdentity(token);
+                if (identity != null
+                    && identity.AuthenticationType[0] == 'n'
```
Tratcher: What is the AuthenticationType for the default app pool identity?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -851,5 +867,22 @@ private enum CurrentOperationType
...
+                    && (identity.AuthenticationType.Equals(AuthenticationSchemes.Negotiate.ToString(), StringComparison.OrdinalIgnoreCase)
```
Tratcher: Use constant strings rather than ToString on every request.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -851,5 +867,22 @@ private enum CurrentOperationType
             Write,
             Flush
         }
+
+        private WindowsPrincipal GetWindowsPrincipal()
+        {
+            var hr = NativeMethods.http_get_authentication_primary_token(_pHttpContext, out IntPtr token);
+            if (hr == 0 && token != IntPtr.Zero)
+            {
+                var identity = new WindowsIdentity(token);
```
Tratcher: We may need to set the auth type parameter.
https://github.com/aspnet/IISIntegration/issues/231
https://github.com/aspnet/HttpSysServer/blob/7f96eab775d721a3bf8d05be446086dab4b45aae/shared/Microsoft.AspNetCore.HttpSys.Sources/RequestProcessing/NativeRequestContext.cs#L203-L232

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -6,15 +6,16 @@
 using Microsoft.AspNetCore.Hosting.Server;
 using System.Threading;
 using System.IO.Pipelines;
+using Microsoft.AspNetCore.Builder;
```
Tratcher: Sort

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -851,5 +867,22 @@ private enum CurrentOperationType
...
+                if (identity != null
+                    && identity.AuthenticationType[0] == 'n'
+                    && (identity.AuthenticationType.Equals(AuthenticationSchemes.Negotiate.ToString(), StringComparison.OrdinalIgnoreCase)
+                    || identity.AuthenticationType.Equals(AuthenticationSchemes.Ntlm.ToString(), StringComparison.OrdinalIgnoreCase)))
```
Tratcher: Add Basic. Out of proc supported it and you're checking for it at startup.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -851,5 +871,29 @@ private enum CurrentOperationType
...
+        private WindowsPrincipal GetWindowsPrincipal()
+        {
+            var hr = NativeMethods.http_get_authentication_information(_pHttpContext, out var authenticationType, out var token);
+
+            if (hr == 0 && token != IntPtr.Zero && authenticationType != null)
+            {
+                var identity = new WindowsIdentity(token, authenticationType);
+                if (identity != null)
+                {
+                    if ((authenticationType.Equals(NtlmString, StringComparison.OrdinalIgnoreCase)
```
Tratcher: Move this check before you create the identity so you don't have to dispose it.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -851,5 +871,29 @@ private enum CurrentOperationType
...
         }
+
+        private WindowsPrincipal GetWindowsPrincipal()
+        {
+            var hr = NativeMethods.http_get_authentication_information(_pHttpContext, out var authenticationType, out var token);
+
+            if (hr == 0 && token != IntPtr.Zero && authenticationType != null)
+            {
+                var identity = new WindowsIdentity(token, authenticationType);
+                if (identity != null)
```
Tratcher: new WindowsIdentity never returns null.

jkotalik: 😮

PR #480, opened 2017-11-21
React to ANCM shim/request handler changes.

jkotalik:
Do not merge until ANCM shim refactor is in. I want to remove IHttpContext from all pInvokes and just have the inprocess handler.


davidfowl:
>  I want to remove IHttpContext from all pInvokes and just have the inprocess handler.

Why?

jkotalik:
The InProcess handler already contains the IHttpContext, it is repetitive to be passing two native pointers.

davidfowl:
Seems bizzare to not pass the native IHttpContext where it makes sense. Most of these methods just need the http context. The inprocess handler is an implementation detail that shouldn't need to exist in this layer.

jkotalik:
The inprocesshandler needs to exist at this layer as we need to store data in native on a per request basis. We used to do this through a stored context based on the module id on the IHttpContext, but because we removed the notion of a module from the request handler, we can't do this anymore. From managed's perspective, it is just receiving an IntPtr to a Http Context, which it cannot access directly.

PR #488, opened 2017-12-11
Upgrade deps

BrennanConroy:
@ryanbrandenburg


BrennanConroy:
We need to update the deps early because the tests run Kestrel and will fail when running with an old Kestrel but new runtime due to coreclr API changes

PR #497, opened 2017-12-30
Adds RequestHandler dll and hostfxr changes from ANCM

jkotalik:
Combination of this change https://github.com/aspnet/AspNetCoreModule/commit/9b794cac971250cc433212eceba6d1e130ad0964 and creating the RequestHandler dll package.

@natemcmaster until we figure out what to do with RIDs, I have left the request handler as one nuget package. Ideally, I would like to swap it to two nuget packages if necessary. Also, we will need to figure out how to pass the shim to @jimmyca15


jkotalik:
@natemcmaster updated, however tests are still disabled. Currently trying to fix with #500 .

natemcmaster:
Ok.

nuget/AspNetCoreRequestHandler.nuspec
```diff
@@ -0,0 +1,23 @@
...
+    <projectUrl>http://www.asp.net/</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>ASP.NET Core Module Request Handler</description>
+    <language>en-US</language>
+    <tags>Microsoft.AspNetCore.AspNetCoreModule.RequestHandler</tags>
+  </metadata>
+  <files>
+  <!-- TODO make build output outside of content files. -->
+    <file src="src\RequestHandler\bin\$Configuration$\Win32\aspnetcorerh.dll" target="runtimes\win-x86\native\aspnetcorerh.dll" />
```
natemcmaster: To make this TFM specific, you can use 'nativeassets'. Example from the SQLite package: https://github.com/natemcmaster/libsqlite3-package/blob/master/SQLite.nuspec#L26-L28

```
runtimes/win-x86/nativeassets/netcoreapp2.1/aspnetcorerh.dll
```

Issue #498, opened 2017-12-30
IISIntegration functional tests are currently failing


jkotalik:
We are now using ANCM as an internal dependency in IISIntegration. Tests seem to either be all passing or failing. I will disable for now and investigate issues this weekend/next week.

jkotalik:
We will either need to wait on hostfxr changes to re-enable these or find a way to specify the request handler globally. When loading the aspnetcore.dll (shim), it will look for the corresponding aspnetcorerh.dll(request handler) in the application, and then looking in a global location. The problem is there is no way for our tests to specify the global location of the request handler, and without hostfxr changes, these tests will never work. Tests were passing locally with the updated hostfxr changes, but those changes are still in prototype/ PR.

jkotalik:
Looks like we will go with the assumption that the request handler will be side by side with the shim (in the same folder).

jkotalik:
New issue: E2E tests seem to be hanging in some scenarios, responding with 400 status codes. Causes tests to timeout eventually. This seems like an issue with the OutOfProcess implementation. @pan-wang I'll try to gather more information.

jkotalik:
An update on this, we are only seeing failures for out of process tests. We can push this to preview2, however I would like to try fixing this asap.

jkotalik:
Going to reenable Out of proc tests now.

jkotalik:
Done pending the web.config generation issue.

PR #500, opened 2018-01-02
Use location of aspnetcore.dll rather than the activating process. 

jkotalik:
Part of fixing #498 , instead of using the location of the process starting ANCM, we will use the location of ANCM to find the request handler. This broke tests because we were using an aspnetcore.dll outside of the iisexpress folder, meaning it couldn't find the request handler.


src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -290,13 +290,15 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
...
     DWORD dwPosition = 0;
+    HMODULE hModuleANCM;
 
     // Though we could call LoadLibrary(L"aspnetcorerh.dll") relying the OS to solve
     // the path (the targeted dll is the same folder of w3wp.exe/iisexpress)
     // let's still load with full path to avoid security issue
     while (!fDone)
     {
-        DWORD dwReturnedSize = GetModuleFileName(NULL, struFilename->QueryStr(), dwSize);
+        hModuleANCM = LoadLibrary(L"aspnetcore.dll");
```
pan-wang: this may not work for site extension scenario

test/IISIntegration.FunctionalTests/HttpsTest.cs
```diff
@@ -138,7 +138,13 @@ public Task Https_HelloWorld_ClientCert_CoreCLR_X64_Portable()
                     ServerConfigTemplateContent = (serverType == ServerType.IISExpress) ? File.ReadAllText("Https.config") : null,
                     SiteName = "HttpsTestSite", // This is configured in the Https.config
                     TargetFramework = runtimeFlavor == RuntimeFlavor.Clr ? "net461" : "netcoreapp2.0",
-                    ApplicationType = applicationType
+                    ApplicationType = applicationType,
+                    Configuration =
+#if DEBUG
```
pan-wang: where is DEBUG defined? VS by default only defines _DEBUG

src/AspNetCore/Src/applicationinfo.cpp
jkotalik: What would we do instead for the site extension scenario? The previous solution would never work with the site extension scenario?

test/IISIntegration.FunctionalTests/HttpsTest.cs
jkotalik: I believe it is defined by MSBuild. Other tests of ours use the same check, I just forgot to add it here too.

jkotalik: See: https://github.com/aspnet/IISIntegration/pull/500/files/ba6658cc25d7d55c2feab7bf69005dab9c771bfb#diff-ba13b4c289f087409caa63ff7ae00bd1R60

jkotalik:
This should fix some of the failures being seen in music store and server tests.

PR #502, opened 2018-01-03
Update gitignore to ignore correct msg files

jkotalik:


natemcmaster:
Assuming these are generated, you also need to remove them from src control. `git rm src/RequestHandler/aspnetcore_msg.h src/RequestHandler/aspnetcore_msg.rc`

jkotalik:
Will be handled by #508

PR #511, opened 2018-01-06
Package aspnetcorerh in Microsoft.AspNetCore.Server.IIS

natemcmaster:
Add a package that carries only the request handler and MSBuild properties required to use it.

Usage: 

 - this package can only be installed into "netcoreapp2.1" as InProcess hosting is not supported on .NET Framework
 - this package contains Windows-only native pieces, so it won't be referenced directly by the aspnetcore metapackages.

TODO:
 - [ ] Update websdk to support generating the `hostingModel` attribute in web.config when the MSBuild property `AspNetCoreModuleHostingModel` is set.
 - [ ] Update the shared framework to consume this for windows-only builds of the shared fx
 - [ ] Update test projects within aspnetcore to use this (again) on Windows only.


natemcmaster:
Followup: 
https://github.com/aspnet/websdk/issues/283
https://github.com/aspnet/Universe/issues/762

@jkotalik are you already looking into updating test projects?

jkotalik:
Yeah I'll start doing that now.

jkotalik:
Test projects will not need to be updated until we start adding standalone Request Handler tests. So the 3rd checkpoint is considered done.

PR #512, opened 2018-01-08
Update sample to use aspnetcore.dll and aspnetcorerh.dll

jkotalik:
I am currently hard coding this to x64. I don't believe $Platform would work. @natemcmaster is there a way to determine the bitness of the published application?


natemcmaster:
> is there a way to determine the bitness of the published application?

This sample is a portable .NET Core app, so CPU arch cannot be determined until runtime.

You could restrict the sample to win-x64 by adding `<RuntimeIdentifier>win-x64</RuntimeIdentifier>` to the csproj.

samples/NativeIISSample/NativeIISSample.csproj
```diff
@@ -5,7 +5,18 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Server.IISIntegration\Microsoft.AspNetCore.Server.IISIntegration.csproj" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(OS)' == 'Windows_NT'">
+    <!-- x86 -->
+    <None Include="..\..\src\RequestHandler\bin\$(Configuration)\Win32\aspnetcorerh.dll" CopyToOutputDirectory="PreserveNewest" Visible="false" Link="x86\aspnetcorerh.dll" />
```
natemcmaster: Does the global shim pickup `<app dir>/{x86,x64}/aspnetcorerh.dll`?

jkotalik: It will look in the directory that aspnetcore.dll is loaded from for the requesthandler.

natemcmaster: 👍 cool. thanks for clarifying

PR #513, opened 2018-01-10
Make paths absolute for in-process, fix relative path check

jkotalik:
When running behind the IISExpressDeployer, the Arguments attribute contains a relative path to the dll (for example .\MyApp.dll). This needs to be an absolute path when calling hostfxr_main. We make a best effort to convert the relative path to an absolute path.
@ryanbrandenburg I want to enable one test on CI for in proc to see if things are consistent on CI. If you see this test failing ever, please ping me and I can disable it.


ryanbrandenburg:
Sounds good. Also your tests are failing on AppVeyor.

jkotalik:
Debugging now :D

jkotalik:
Currently can't enable tests due to #515 .

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -25,24 +25,21 @@ HOSTFXR_UTILITY::~HOSTFXR_UTILITY()
 //
 HRESULT
 HOSTFXR_UTILITY::GetStandaloneHostfxrParameters(
+    STRU*              pStruExePath,
```
pan-wang: why not pass in PWCSTR

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -200,21 +201,21 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
     {
         goto Finished;
     }
 
     // As we use the logic from core-setup, we are opting to use std here.
     // TODO remove all uses of std?
-    UTILITY::FindDotNetFolders(strHostFxrSearchExpression.QueryStr(), &vVersionFolders);
+    UTILITY::FindDotNetFolders(struHostFxrSearchExpression.QueryStr(), &vVersionFolders);
 
     if (vVersionFolders.size() == 0)
```
pan-wang: if we did not find the folder, is it an error?

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -276,14 +277,22 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
     PCWSTR*     argv = NULL;
     LPWSTR*     pwzArgs = NULL;
+    STRU        struTempPath;
 
     pwzArgs = CommandLineToArgvW(struArgumentsFromConfig->QueryStr(), &argc);
 
-    if (pwzArgs == NULL)
+    if (pwzArgs == NULL || argc < 1)
     {
         goto Finished;
```
pan-wang: is this an error

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -276,14 +277,22 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
     }
 
+    // Try to convert the application dll from a relative to an absolute path
+    // Don't record this failure as pwzArgs[0] may already be an absolute path to the dll.
+    if (!FAILED(UTILITY::ConvertPathToFullPath(pwzArgs[0], pConfig->QueryApplicationPhysicalPath()->QueryStr(), &struTempPath)))
+    {
+        pwzArgs[0] = struTempPath.QueryStr();
+    }
+
```
pan-wang: argv[0] = SysAllocString, you may want to do NULL check on argv[0] and argv[1] to see whether the SysAllocString success or not

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -308,4 +317,4 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
         DBG_ASSERT(pwzArgs == NULL);
     }
     return hr;
-}
\ No newline at end of file
+}
```
pan-wang: in case of failure, you need to call SysFreeString to free memory used by argv otherwise there will be memory leak

src/CommonLib/hostfxr_utility.cpp
jkotalik: Yes, that means we can't find hostfxr.dll, so we can't run in process. All of these need logs once your logging PR gets in.

src/CommonLib/hostfxr_utility.cpp
jkotalik: pwzArgs[0] will actually be argv[2]. I can add those checks.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -268,8 +268,8 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
 // 
 HRESULT
 HOSTFXR_UTILITY::SetHostFxrArguments(
-    STRU* struArgumentsFromConfig,
-    STRU* pstruExePath,
+    PCWSTR struArgumentsFromConfig,
```
pan-wang: rename the parameter as it is not stru anymore

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -297,18 +306,42 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
     {
         hr = E_OUTOFMEMORY;
-        goto Finished;
+        goto Failure;
+    }
+
+    argv[0] = SysAllocString(pstruExePath);
+    if (argv[0] == NULL)
+    {
+        goto Failure;
```
pan-wang: hr value

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -297,18 +306,42 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
     }
+    dwArgsProcessed++;
 
-    argv[0] = SysAllocString(pstruExePath->QueryStr());
     argv[1] = SysAllocString(L"exec");
+    if (argv[1] == NULL)
+    {
+        goto Failure;
```
pan-wang: hr value

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -297,18 +306,42 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+    }
+    dwArgsProcessed++;
 
     for (INT i = 0; i < argc; i++)
     {
         argv[i + 2] = SysAllocString(pwzArgs[i]);
+        if (argv[1] == NULL)
+        {
+            goto Failure;
```
pan-wang: hr value

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -297,18 +306,42 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+        }
+        dwArgsProcessed++;
     }
 
     pConfig->SetHostFxrArguments(argc + 2, argv);
```
pan-wang: SetHostFxrArguments may fail. Get the hresult value

jkotalik: SetHostFxrArguments doesn't return an HRESULT and there is no way for it to fail right now.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -94,18 +91,31 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
+        pConfig->QueryApplicationPhysicalPath()->QueryStr(),
+        &struExeLocation
+    );
+
+    if (FAILED(hr))
+    {
+        goto Finished;
+    }
+
+    if (UTILITY::CheckIfFileExists(struExeLocation.QueryStr()))
```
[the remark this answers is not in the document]
jkotalik: This was fixed here, do you mind verifying this fixed your issue?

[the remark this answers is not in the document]
jhkimnew: Okay, it works, but I found another usability bug. In 64 bit mode, specifying only dotnet.exe worked for both outofprocess and inprocess mode. However, in 32 bit mode, specifying only dotnet.exe works only for outofprocess mode. Would you make that works in 32 bit mode as it does in 64 bit mode?

src/CommonLib/hostfxr_utility.cpp
pan-wang: this SetHostFxrArguments is from Config class. I mistakenly read it with utility

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -267,39 +268,83 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
+    if (argc < 1)
+    {
+        // Invalid arguments
+        hr = E_INVALIDARG;
+        goto Failure;
+    }
+
+    // Try to convert the application dll from a relative to an absolute path
+    // Don't record this failure as pwzArgs[0] may already be an absolute path to the dll.
+    if (!FAILED(UTILITY::ConvertPathToFullPath(pwzArgs[0], pConfig->QueryApplicationPhysicalPath()->QueryStr(), &struTempPath)))
```
pan-wang: !FAILED() looks wired. please use SUCCESS()

src/CommonLib/hostfxr_utility.cpp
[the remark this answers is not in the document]
jkotalik: I'll address this in another PR.

jkotalik:
@pan-wang I removed the memory leak on pwzArgs too.

PR #517, opened 2018-01-17
[WIP] Adds Client Disconnect to InProc

jkotalik:
ClientDisconnect works, however there is another open issue. Right now, the only time we receive disconnect events is whenever we return RQ_NOTIFICATION_PENDING which only occurs when async operations are occurring. We cannot guarantee any async operations will occur, as it is dependent on if reads or writes complete asynchronously or synchronously. 
@pan-wang for native, @Tratcher for managed, @davidfowl for whatever you feel like roasting me on.


Tratcher:
That catch negates the point of the feature. The read/write operations would fail and inform you of the disconnect directly. This new notification is for interrupting long running tasks when you're not interacting with the client.

Don't merge this until we understand how Asp.Net 4 made this work.

jkotalik:
Punting this to preview-2, risky feature to add.

jkotalik:
After further investigation, it seems to potentially be an issue with disconnect notifications for localhost requests. I need to investigate further, however it does seem the disconnect event doesn't need to return to the IIS request pipeline to receive the notification.

jkotalik:
Closing as there will be a lot of churn in this implementation.

Issue #519, opened 2018-01-17
ANCM: Inprocess: 500 error when 32 bit dotnet installation path is set before 64 bit's installation path for the PATH environment variable


jhkimnew:
1. Configure PATH environment variable to put the path of 32 bit dotnet.exe before 64 bit path.
Ex.) %SystemRoot%\system32;%SystemRoot%;%SystemRoot%\System32\Wbem;%SYSTEMROOT%\System32\WindowsPowerShell\v1.0\;C:\Program Files (x86)\dotnet;C:\Program Files\dotnet;
2. Restart IIS services to apply the change
3. Configure inprocess mode and set relative path for dotnet.exe, not full path
4. Send a request

EXPECTED:

No error

ACTUAL:

500 error

muratg:
@jkotalik close?

PR #524, opened 2018-01-18
Default to checking program files if dotnet on path doesn't match bitness of worker process.

jkotalik:
For #519. We could also check further on the path too, but I believe this is the approach we wanted to take.


Tratcher:
Does this work both ways? E.g. 32->64 and 64->32?

Also, I remembered why we didn't do this to start with, it breaks dev environments where the dotnet we want isn't in program files, it's in the user profile. What was that you said about continuing to search down the user's path? Can SearchPath be called again to keep looking?

jkotalik:
@Tratcher can you clarify the scenario for going both ways? Like switching between 32 and 64 bit worker processes?

For IISExpress, this will work in dev environments because we still look on the system path for dotnet first (where we expect people to add C:\Users\USER_PROFILE\.dotnet\x64 to the path). However it won't work in the IIS scenario as it won't have access to the user profile. This could cause issues running full IIS tests on our CI machines.

Instead of looking in program files, we could instead look further on the path, where other dotnet locations would be specified. For example, my path is C:\Users\USER_PROFILE\.dotnet\x64;C:\Program Files (x86)\dotnet;C:\Program Files\dotnet. SearchPath would return the first C:\Users\USER_PROFILE\.dotnet\x64, however I would like to get the x86 location as it matches the bitness of my process. I'm going to investigate whether we can continue calling SearchPath to look further on the path. My guess is no, but I will confirm.

jkotalik:
Closing in favor of #541.

jhkimnew:
@jkotalik Can you list the required test scenarios after this code update?

jkotalik:
Yeah I'll let you know once #541 is in.

Issue #521, opened 2018-01-17
ANCM: Inprocess: Exception error in 32 bit mode


jhkimnew:
If users switch IIS apppool to 32bit mode, it does not work and the worker process crashes.
I got the stack information when the first chance of exception happen and attached it below. 

FYI, adding one link regarding the error: https://stackoverflow.com/questions/3914750/hows-rtc-checkesp-implemented

0:005:x86> kn
# ChildEBP RetAddr  
00 00ebf6e4 6b49fbb9 aspnetcorerh!failwithmessage+0x1ea
01 00ebf6fc 6b49f02f aspnetcorerh!_RTC_Failure+0x29
02 00ebf720 6b483130 aspnetcorerh!_RTC_CheckEsp+0x1f
03 00ebf74c 6f1d8138 aspnetcorerh!IN_PROCESS_APPLICATION::OnExecuteRequest+0x40 [d:\gitroot\iisintegration\src\requesthandler\inprocess\inprocessapplication.cpp @ 160] 
04 00ebf9e8 6f5f2792 aspnetcore!ASPNET_CORE_PROXY_MODULE::OnExecuteRequestHandler+0x2a8 [d:\gitroot\iisintegration\src\aspnetcore\src\proxymodule.cxx @ 168] 
WARNING: Stack unwind information not available. Following frames may be wrong.
05 00ebfa08 6f5f1c58 iiscore+0x22792
06 00ebfa94 6f5e34ed iiscore+0x21c58
07 00ebfae0 6f5f33cc iiscore+0x134ed
08 00ebfb08 6f5f2e84 iiscore+0x233cc
09 00ebfd7c 6f5e6b00 iiscore+0x22e84
0a 00ebfda4 6f5e6f73 iiscore+0x16b00
0b 00ebfdbc 6f5b84f2 iiscore+0x16f73
0c 00ebfdd4 6f5c35c6 w3dt!HTTP_WRAPPER::QueryState+0x35bd
0d 00ebfdec 6f5b7b02 w3dt!HTTP_WRAPPER::QueryState+0xe691
0e 00ebfe04 6ff03213 w3dt!HTTP_WRAPPER::QueryState+0x2bcd
0f 00ebfe38 6ff03124 W3TP!THREAD_POOL::PostCompletion+0x142
10 00ebfe58 6ff0209f W3TP!THREAD_POOL::PostCompletion+0x53
11 00ebfe74 76237c04 W3TP+0x209f
[...]

PR #527, opened 2018-01-19
Fixes bitness alignment issue for 32 bit builds

jkotalik:
For #521 
I believe this issue only occurred in Debug builds. Adding the WINAPI is the same as adding __stdcall, which I believe is needed as these functions are defined in managed.


jkotalik:
@pan-wang approved this offline. Going to merge in as it is important for preview1

Issue #522, opened 2018-01-18
ANCM: Inprocess: 500 error after switching to inprocess from outofprocess mode


jhkimnew:
REPRO STEPS:
1. Set hostingMode with outofprocess or empty value (default value is outofprocess)
2. Send a request and you can see w3wp.exe is started
3. Set hostingMode with inprocess
4. Send a request

EXPECTED:

No error 

ACTUAL:

500 error

NOTE:

1. In the aspnetcorerh!IN_PROCESS_APPLICATION::ExecuteApplication function, LoadLibraryW(m_pCOnfig->QueryHostFxrFullPath()) returns false and that's why the 500 error happens.

At the moment, the m_struHostFxrLocation was empty value, which means that the value was used here without waiting it is initialized. 

2. Attached the stack information.

0:018> kn
 # Child-SP          RetAddr           Call Site
00 00000096`d0e0f850 00007ffa`21653e7a aspnetcorerh!IN_PROCESS_APPLICATION::ExecuteApplication+0x71 [d:\gitroot\iisintegration\src\requesthandler\inprocess\inprocessapplication.cpp @ 525] 
*** ERROR: Module load completed but symbols could not be loaded for C:\Windows\SYSTEM32\vfbasics.dll
01 00000096`d0e0f900 00007ffa`21731d54 aspnetcorerh!IN_PROCESS_APPLICATION::ExecuteAspNetCoreProcess+0x3a [d:\gitroot\iisintegration\src\requesthandler\inprocess\inprocessapplication.cpp @ 501] 
02 00000096`d0e0f940 00007ffa`3ba413d2 vfbasics+0x11d54
03 00000096`d0e0f980 00007ffa`3de75444 KERNEL32!BaseThreadInitThunk+0x22
04 00000096`d0e0f9b0 00000000`00000000 ntdll!RtlUserThreadStart+0x34

muratg:
Workaround is manually recycling the process after changing the config.

PR #529, opened 2018-01-19
Always load hostfxr parameters if the app isn't loaded

jkotalik:
Load the hostfxr parameters if the application isn't loaded, rather than when the request handler is loaded. This is because we store the hostfxr parameters in the aspnetcore_config, which will be deleted on recycle. 
For #522. I updated dependencies too because out of process had pipeline errors in kestrel.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -3,42 +3,42 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
   <PropertyGroup Label="Package Versions">
-    <InternalAspNetCoreSdkPackageVersion>2.1.0-preview1-15651</InternalAspNetCoreSdkPackageVersion>
-    <MicrosoftAspNetCoreAllPackageVersion>2.1.0-preview1-27965</MicrosoftAspNetCoreAllPackageVersion>
-    <MicrosoftAspNetCoreAuthenticationCorePackageVersion>2.1.0-preview1-27965</MicrosoftAspNetCoreAuthenticationCorePackageVersion>
-    <MicrosoftAspNetCoreHostingAbstractionsPackageVersion>2.1.0-preview1-27965</MicrosoftAspNetCoreHostingAbstractionsPackageVersion>
-    <MicrosoftAspNetCoreHostingPackageVersion>2.1.0-preview1-27965</MicrosoftAspNetCoreHostingPackageVersion>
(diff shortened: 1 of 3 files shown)


build/dependencies.props
```diff
@@ -3,42 +3,42 @@
...
-    <MicrosoftExtensionsConfigurationEnvironmentVariablesPackageVersion>2.1.0-preview1-27965</MicrosoftExtensionsConfigurationEnvironmentVariablesPackageVersion>
-    <MicrosoftExtensionsConfigurationJsonPackageVersion>2.1.0-preview1-27965</MicrosoftExtensionsConfigurationJsonPackageVersion>
-    <MicrosoftExtensionsLoggingAbstractionsPackageVersion>2.1.0-preview1-27965</MicrosoftExtensionsLoggingAbstractionsPackageVersion>
-    <MicrosoftExtensionsLoggingConsolePackageVersion>2.1.0-preview1-27965</MicrosoftExtensionsLoggingConsolePackageVersion>
-    <MicrosoftExtensionsLoggingDebugPackageVersion>2.1.0-preview1-27965</MicrosoftExtensionsLoggingDebugPackageVersion>
-    <MicrosoftExtensionsLoggingPackageVersion>2.1.0-preview1-27965</MicrosoftExtensionsLoggingPackageVersion>
-    <MicrosoftExtensionsLoggingTestingPackageVersion>2.1.0-preview1-27965</MicrosoftExtensionsLoggingTestingPackageVersion>
-    <MicrosoftExtensionsOptionsPackageVersion>2.1.0-preview1-27965</MicrosoftExtensionsOptionsPackageVersion>
+    <InternalAspNetCoreSdkPackageVersion>2.1.0-preview1-15677</InternalAspNetCoreSdkPackageVersion>
```
pan-wang: is this version change required? if we have change it every time, it seems very annoying.

korebuild-lock.txt
```diff
@@ -1,2 +1,2 @@
-version:2.1.0-preview1-15661
-commithash:c9349d4c8a495d3085d9b879214d80f2f45e2193
+version:2.1.0-preview1-15677
```
pan-wang: not sure about this change

jkotalik: Same as above, except this is to pull the latest version of our build tools.

build/dependencies.props
jkotalik: Yes. These are changes to pick up new dependencies from other repos. This process is automatic, however if you would like to update the managed dependencies, you can run run.ps1 upgrade deps

PR #530, opened 2018-01-19
Panwang/filewatcher

pan-wang:
update thread terminate logic as the file watcher thread may get reuse by OS

### DIFF
--- a/src/AspNetCore/Src/filewatcher.cxx
+++ b/src/AspNetCore/Src/filewatcher.cxx
@@ -14,14 +14,34 @@ FILE_WATCHER::~FILE_WATCHER()
 {
     if (m_hChangeNotificationThread != NULL)
     {
-        DWORD dwRetryCounter = 10;
+        DWORD dwRetryCounter = 20;      // totally wait for 1s
+        DWORD dwExitCode = STILL_ACTIVE;
 
         // signal the file watch thread to exit
         PostQueuedCompletionStatus(m_hCompletionPort, 0, FILE_WATCHER_SHUTDOWN_KEY, NULL);
         while (!m_fThreadExit && dwRetryCounter > 0)
         {
-            Sleep(10); // totally wait for 100 ms
+            if (GetExitCodeThread(m_hChangeNotificationThread, &dwExitCode))
+            {
+                if (dwExitCode == STILL_ACTIVE)
+                {
+                    // the file watcher thread will set m_fThreadExit before exit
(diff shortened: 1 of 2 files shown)


src/AspNetCore/Src/filewatcher.cxx
```diff
@@ -5,16 +5,45 @@
...
-        WaitForSingleObject(m_hChangeNotificationThread, INFINITE);
+        while (!m_fThreadExit && dwRetryCounter > 0)
+        {
+            if (GetExitCodeThread(m_hChangeNotificationThread, &dwExitCode))
+            {
+                if (dwExitCode == STILL_ACTIVE)
+                {
+                    // the file watcher thread will set m_fThreadExit before exit
+                    WaitForSingleObject(m_hChangeNotificationThread, 50);
+                }
```
jkotalik: I am pretty sure you need an else here to set m_fThreadExit to true. If the exit code is not active, the thread has already terminated.
The could would be:
```c++
if (dwExitCode == STILL_ACTIVE)
{
    // the file watcher thread will set m_fThreadExit before exit
    WaitForSingleObject(m_hChangeNotificationThread, 50);
}
else
{
    m_fThreadExit = TRUE;
}
```

src/AspNetCore/Src/filewatcher.cxx
```diff
@@ -126,7 +155,8 @@ Win32 error
         cbCompletion = 0;
     }
 
-    return 0;
+    pFileMonitor->m_fThreadExit = TRUE;
+    ExitThread(0);
```
jkotalik: Why are we calling ExitThread instead of returning here?

src/AspNetCore/Src/filewatcher.cxx
pan-wang: another thread should already set m_fThreadExit to true before its exit

src/AspNetCore/Src/filewatcher.cxx
pan-wang: just want to make sure the thread signal/status does change.  https://msdn.microsoft.com/en-us/library/windows/desktop/ms682659(v=vs.85).aspx, as we call waitforsingleobject.

src/AspNetCore/Src/filewatcher.cxx
jkotalik: Oh I see. That makes sense.

jkotalik:
Rename this pull request before you merge too to something more informative.

PR #532, opened 2018-01-19
Add project capability for inproc hosting

shirhatti:


src/Microsoft.AspNetCore.Server.IIS/Microsoft.AspNetCore.Server.IIS.targets
```diff
@@ -11,4 +11,12 @@
     -->
     <AspNetCoreModuleHostingModel Condition=" '$(AspNetCoreModuleHostingModel)' == '' AND '$(DisableInProcessHosting)' != 'true' ">InProcess</AspNetCoreModuleHostingModel>
   </PropertyGroup>
+  
+  <!--
+    Capability that enables Visual Studio supoort for hosting Asp.Net Core applications in the IIS process
```
natemcmaster: typo: support

jkotalik:
Do we still need both of these properties? Otherwise LGTM

shirhatti:
Capability is to indicate whether in-process is supported or not, i.e., should tooling even give you the option to switch.
Property is to indicate what mode we're currently operating in.

jkotalik:
For #532

shirhatti:
@jkotalik Can you approve?

shirhatti:
> Property is to indicate what mode we're currently operating in.

Based on this and our offline conversation we should no longer set the `AspNetCoreModuleHostingModel` property

PR #533, opened 2018-01-19
Fail fast on .NET Full Framework and log exceptions in hostfxr utility

jkotalik:
To check if the application is full framework, I check for the existence of APPNAME.runtimeconfig.json. 
@pan-wang I passed in the EventLog handle to each of these functions as it is part of CommonLib. We want to keep hostfxr_utility in CommonLib in the case where we need to parse configuration in the request handler.
For https://github.com/aspnet/AspNetCoreModule/issues/255


shirhatti:
Can you share an actual entry from eventlog.xml?

src/CommonLib/resources.h
```diff
@@ -27,3 +27,4 @@
 #define ASPNETCORE_EVENT_INPROCESS_THREAD_EXIT_MSG           L"Application '%s' with physical root '%s' hit unexpected managed background thread exit, ErrorCode = '0x%x."
 #define ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_MSG              L"Application '%s' is recycled due to app_offline file was detected."
 #define ASPNETCORE_EVENT_MODULE_DISABLED_MSG                 L"AspNetCore Module is disabled"
+#define ASPNETCORE_EVENT_INPROCESS_FULL_FRAMEWORK_APP_MSG    L"Application '%s' was compiled for .NET Framework. Please compile for .NET core to run the inprocess application."
```
Tratcher: or change the process mode to out of process.

jkotalik:
```xml
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
 <System>
  <Provider Name="IIS AspNetCore Module" /> 
  <EventID Qualifiers="0">1014</EventID> 
  <Level>2</Level> 
  <Task>0</Task> 
  <Keywords>0x80000000000000</Keywords> 
  <TimeCreated SystemTime="2018-01-19T22:22:40.418569800Z" /> 
  <EventRecordID>212475</EventRecordID> 
  <Channel>Application</Channel> 
  <Computer>MININT-8BN252A.redmond.corp.microsoft.com</Computer> 
  <Security /> 
  </System>
 <EventData>
  <Data>Application '/LM/W3SVC/2/ROOT' was compiled for .NET Framework. Please compile for .NET core to run the inprocess application or change the process mode to out of process.</Data> 
  </EventData>
  </Event>
```

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -26,35 +26,74 @@ HOSTFXR_UTILITY::~HOSTFXR_UTILITY()
...
     HRESULT             hr = S_OK;
     STRU                struDllPath;
     STRU                struArguments;
+    STRU                struHostFxrPath;
+    STRU                struRuntimeConfigLocation;
+    STRU                strEventMsg;
     DWORD               dwPosition;
 
-    if (FAILED(hr))
+    // Obtain the app name from the processPath section.
```
pan-wang: is this pwzExePath full path or relative path? does it must have file extension? you can rename it or add comment.

jkotalik:
I added a few other log events to the hostfxr code path.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -148,6 +178,16 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
             {
                 hr = GetLastError();
```
pan-wang: GetLastError() does not match to Hresult. You need to convert to with HRESULT_FROM_WIN32(GetLastError())

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -219,7 +268,16 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
+            ASPNETCORE_EVENT_HOSTFXR_DIRECTORY_NOT_FOUND_MSG,
+            struHostFxrPath.QueryStr(),
+            hr)))
+        {
+            UTILITY::LogEvent(hEventLog,
+                EVENTLOG_ERROR_TYPE,
+                ASPNETCORE_EVENT_HOSTFXR_DIRECTORY_NOT_FOUND,
+                struEventMsg.QueryStr());
+        }
         hr = ERROR_BAD_ENVIRONMENT;
```
pan-wang: you need move this line up  before logging

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -591,7 +591,6 @@ IN_PROCESS_APPLICATION::RunDotnetApplication(DWORD argc, CONST PCWSTR* argv, hos
     }
     __except (FilterException(GetExceptionCode(), GetExceptionInformation()))
     {
```
pan-wang: you may want to add log here. As this failure is critical

muratg: +1

jkotalik: We already log this failure outside of this function https://github.com/aspnet/IISIntegration/blob/dev/src/RequestHandler/inprocess/inprocessapplication.cpp#L556 and when the thread returns https://github.com/aspnet/IISIntegration/blob/dev/src/RequestHandler/inprocess/inprocessapplication.cpp#L472

jkotalik: It also needs to be outside of that function because __try, __except can't have any objects that require object unwinding, like STRU.

muratg: OK cool 👍

src/CommonLib/hostfxr_utility.cpp
jkotalik: Changed the name to AbsolutePath

jkotalik:
@pan-wang I am now checking and logging the exception code when the coreclr has an unhandled exception.

pan-wang:
If hostfxrutility us only used by shim,  you don't need to pass the eventlog handle these is a global handle.

jkotalik:
Though today hostfxr_utility is only used in the shim, we wanted to give the option for the request handler to reparse aspnetcore_config, which would require use to find hostfxr again. Hence I left it as a utility rather than a first class member of the AspNetCore project.

jkotalik:
@pan-wang updated.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -25,58 +25,123 @@ HOSTFXR_UTILITY::~HOSTFXR_UTILITY()
 //
 HRESULT
 HOSTFXR_UTILITY::GetStandaloneHostfxrParameters(
-    PCWSTR              pwzExePath,
-    ASPNETCORE_CONFIG *pConfig
+    PCWSTR              pwzExeAbsolutePath, // includes .exe file extension.
+	PCWSTR				pcwzApplicationPhysicalPath,
```
pan-wang: format

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -25,58 +25,123 @@ HOSTFXR_UTILITY::~HOSTFXR_UTILITY()
...
+                                   EVENTLOG_ERROR_TYPE,
+                                   ASPNETCORE_EVENT_INPROCESS_FULL_FRAMEWORK_APP,
+                                   strEventMsg.QueryStr() );
+            }
+        }
+        else
+        {
+            // If a runtime config file does exist, report a file not found on the app.exe
+            hr = HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND);
+			if (SUCCEEDED(strEventMsg.SafeSnwprintf(
```
pan-wang: format

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -85,7 +150,13 @@ HOSTFXR_UTILITY::GetStandaloneHostfxrParameters(
 
 HRESULT
 HOSTFXR_UTILITY::GetHostFxrParameters(
-    ASPNETCORE_CONFIG *pConfig
+    HANDLE              hEventLog,
+	PCWSTR				pcwzProcessPath,
```
pan-wang: format

src/CommonLib/hostfxr_utility.h
```diff
@@ -14,26 +14,40 @@ class HOSTFXR_UTILITY
...
+		_Out_ PWSTR**		ppwzArgv
+	);
 
 private:
     static
     HRESULT
     GetStandaloneHostfxrParameters(
-        PCWSTR             pStruExePath,
-        ASPNETCORE_CONFIG *pConfig
+		PCWSTR              pwzExeAbsolutePath, // includes .exe file extension.
```
pan-wang: format

Issue #537, opened 2018-01-22
ANCM: Inprocess: Appverifier error "Waiting on a thread handle in DllMain"


jhkimnew:
When recycling an IIS worker process where inprocess aspnetcore application is running, the below appverifer error happens.

```
=======================================
VERIFIER STOP 0000000000000304: pid 0x1130: Waiting on a thread handle in DllMain. 

	0000000000000628 : Thread handle.
	0000000000000000 : Not used.
	0000000000000000 : Not used.
	0000000000000000 : Not used.


=======================================
This verifier stop is continuable.
After debugging it use `go' to continue.

=======================================

(1130.91c): Break instruction exception - code 80000003 (first chance)
*** ERROR: Symbol file could not be found.  Defaulted to export symbols for C:\Windows\SYSTEM32\vrfcore.dll - 
*** ERROR: Module load completed but symbols could not be loaded for C:\Windows\SYSTEM32\vfbasics.dll
vrfcore!VerifierStopMessageEx+0x836:
00007ffa`35562626 cc              int     3
0:017> kn
 # Child-SP          RetAddr           Call Site
00 0000001a`8b5ee8b0 00007ffa`308d3018 vrfcore!VerifierStopMessageEx+0x836
01 0000001a`8b5eec00 00007ffa`308e4f07 vfbasics+0x3018
*** ERROR: Symbol file could not be found.  Defaulted to export symbols for C:\Windows\system32\KERNELBASE.dll - 
02 0000001a`8b5eed70 00007ffa`3b371118 vfbasics+0x14f07
03 0000001a`8b5eeda0 00007ffa`308e4a4a KERNELBASE!WaitForSingleObjectEx+0x98
04 0000001a`8b5eee40 00007ffa`308e4ac8 vfbasics+0x14a4a
05 0000001a`8b5eee70 00007ffa`308e4a4a vfbasics+0x14ac8
06 0000001a`8b5eeeb0 00007ffa`308e4aa5 vfbasics+0x14a4a
*** WARNING: Unable to verify checksum for C:\Windows\system32\inetsrv\aspnetcore.dll
07 0000001a`8b5eeee0 00007ffa`232665fb vfbasics+0x14aa5
08 0000001a`8b5eef20 00007ffa`23264bbc aspnetcore!FILE_WATCHER::~FILE_WATCHER+0xbb [d:\gitroot\iisintegration\src\aspnetcore\src\filewatcher.cxx @ 29] 
[...]

Issue #525, opened 2018-01-19
ANCM: Parent level configuration change should recycle applications in child level


jhkimnew:
This is a regression issue. 

STEPS TO REPRO:

1. Create an IIS web application with an aspnetcore app
2. Send a request to the app
3. Change any configuration from the site level

EXPECTED:

The application should be recycled after the configuration change

ACTUAL:

The application is not recycled.

PR #539, opened 2018-01-23
some fix for handling configuration change and recycling

pan-wang:
fixing issue #525 and #537

### DIFF
--- a/src/AspNetCore/Inc/globalmodule.h
+++ b/src/AspNetCore/Inc/globalmodule.h
@@ -28,8 +28,8 @@ public:
     );
 
     GLOBAL_NOTIFICATION_STATUS
-    OnGlobalConfigurationChange(
-        _In_ IGlobalConfigurationChangeProvider * pProvider
+    OnGlobalApplicationStop(
+        _In_ IHttpApplicationStopProvider * pProvider
     );
 
 private:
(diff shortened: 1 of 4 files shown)


jkotalik:
How much did you verify this @pan-wang . We should confirm that application stop does trigger on application stop and configuration change.

shirhatti:
Does this address some prior issues around this space?
- https://github.com/aspnet/AspNetCoreModule/issues/105
- https://github.com/aspnet/AspNetCoreModule/issues/247

pan-wang:
this pr address https://github.com/aspnet/IISIntegration/issues/525 and https://github.com/aspnet/IISIntegration/issues/537. We do have tests covering those scenarios.  @jhkimnew can confirm.

@shirhatti I will try repro that two issues. The code got changed significantly. they may not reproduce anymore.

jkotalik:
@pan-wang are you asking @jhkimnew to verify these changes?

shirhatti:
@pan-wang Given that my question went unanswered, I'm assuming we never tried to repro these issues 😄 
- https://github.com/aspnet/AspNetCoreModule/issues/105
- https://github.com/aspnet/AspNetCoreModule/issues/247

@jhkimnew Can you please give it a try and close the respective issue if they no longer repro?

pan-wang:
@shirhatti, I did try repro it but could not. The app-offline and handling web.config change were changed in the new code. This issue should be not reproducible.

PR #541, opened 2018-01-23
Invoke where.exe to find dotnet.exe

jkotalik:
Another implementation of #519 which calls where.exe instead.


src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,166 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+
+    // Set stdout and error to redirect to the temp file.
+    startupInfo.cb = sizeof(startupInfo);
+    startupInfo.dwFlags |= STARTF_USESTDHANDLES;
+    startupInfo.hStdOutput = hStdOutWritePipe;
+    startupInfo.hStdError = hStdOutWritePipe;
+
+    // CreateProcess requires a mutable string to be passed to commandline
+    // See https://blogs.msdn.microsoft.com/oldnewthing/20090601-00/?p=18083/
+    pwzDotnetName = SysAllocString(L"\"C:\\Windows\\System32\\where.exe\" dotnet.exe");
```
Tratcher: Can you do this in azure web apps?

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,166 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+        if (fIsCurrentProcess64Bit == (dwBinaryType == SCS_64BIT_BINARY)) {
+            // Found a valid dotnet.
+            struDotnetPath->Copy(struDotnetSubstring);
+            fFound = TRUE;
+        }
+    }
+
+    if (!fFound)
+    {
+        hr = ERROR_FILE_NOT_FOUND;
```
Tratcher: How can the user get visibility into this failure? E.g. how can they get a verbose trace of what paths were searched and why they failed?

jkotalik: I was just about the comment on that. I have another PR: https://github.com/aspnet/IISIntegration/pull/533 which logs failures in hostfxr_utility. I was going to add logs throughout this call. I can rebase that branch onto this branch until the PR is in. 

To know which paths were searched, in the windows event log I will output the contents as part of the error message.

src/CommonLib/hostfxr_utility.cpp
jkotalik: @pan-wang how were you verifying the logging changes before?

pan-wang: you cannot use hardcoded absolute path. you can just use "where.exe dotnet". OS will resolve the path of where.exe is.
I opened eventviewer to check the log and the stderr log file.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,166 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+        CREATE_NO_WINDOW,
+        NULL,
+        NULL,
+        &startupInfo,
+        &processInformation
+    );
+
+    if (!fResult)
+    {
+        hr = ERROR_PROCESS_ABORTED;
```
pan-wang: please use HRESULT_FROM_WIN32(GetLastError())

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,166 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+        goto Finished;
+    }
+
+    WaitForSingleObject(processInformation.hProcess, pConfig->QueryRequestTimeoutInMS());
+
+    fResult = GetExitCodeProcess(processInformation.hProcess, &dwExitCode);
```
pan-wang: in case of child process hang or long running. closehandle will not kill the child process.

pan-wang: what's the motivation for calling GetExitCodeProcess here if you don't want to check if the process is still active?

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,166 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+        hr = ERROR_FILE_INVALID;
+        goto Finished;
+    }
+
+    if (!ReadFile(hStdOutReadPipe, pzFileContents, READ_BUFFER_SIZE, &dwNumBytesRead, NULL))
+    {
+        hr = ERROR_FILE_INVALID;
+        goto Finished;
+    }
+    if (dwNumBytesRead == READ_BUFFER_SIZE)
```
pan-wang: it is  safer to use >= instead of ==

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,166 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+    {
+        // This shouldn't ever be this large. We could continue to call ReadFile in a loop,
+        // however I'd rather error out here and report an issue.
+        hr = ERROR_FILE_TOO_LARGE;
+        goto Finished;
+    }
+    struDotnetLocationsString.CopyA(pzFileContents, dwNumBytesRead);
```
pan-wang: hr check for CopyA()

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,166 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+    }
+
+    while (!fFound)
+    {
+        index = struDotnetLocationsString.IndexOf(L"\r\n", prevIndex);
+        if (index == -1)
+        {
+            break;
+        }
+        struDotnetSubstring.Copy(struDotnetLocationsString.QueryStr(), index - prevIndex);
```
pan-wang: hr check

davidfowl:
Are we tracking the fact that we’ll still need to do search paths differently with respect to the process bitness?

src/CommonLib/hostfxr_utility.cpp
pan-wang: null check after SysAllocString

jkotalik:
> Are we tracking the fact that we’ll still need to do search paths differently with respect to the process bitness?

For each dotnet.exe that is obtained from where, I check that dotnet.exe matches the bitness of the currently running process.

src/CommonLib/hostfxr_utility.cpp
jkotalik: Cause I forgot a check XD. Also I should be calling TerminateProcess too.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,189 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+        &processInformation
+    );
+
+    if (!fResult)
+    {
+        hr = HRESULT_FROM_WIN32(GetLastError());
+        goto Finished;
+    }
+
+    if (WaitForSingleObject(processInformation.hProcess, pConfig->QueryRequestTimeoutInMS()) != WAIT_OBJECT_0)
```
jkotalik: Offline feedback: Make this timeout fixed (like 2 seconds).

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,189 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+    {
+        TerminateProcess(processInformation.hProcess, 1);
```
jkotalik: Set hr = E_TIMEOUT and goto Finished;

pan-wang:
You may want to add some logs for the failure cases to help diagnoses.

jkotalik:
I wanted to get https://github.com/aspnet/IISIntegration/pull/533 in first as it has some of the logging changes necessary (passing the logEvent around). Do you have any feedback on that one?

jkotalik:
Verified this works on antares.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -369,3 +342,207 @@ HOSTFXR_UTILITY::SetHostFxrArguments(
...
+    }
+    if (processInformation.hThread != NULL)
+    {
+        CloseHandle(processInformation.hThread);
+    }
+    if (pwzDotnetName != NULL)
+    {
+        SysFreeString(pwzDotnetName);
+    }
+
```
jkotalik: Close the pipe handles

jkotalik:
@pan-wang updated with offline feedback.

jkotalik:
@muratg This will help with the azure website scenario for check that dotnet matches the bitness of your app. Add to preview1?

jkotalik:
Closing for #560 (from preview 1 branch)

PR #543, opened 2018-01-23
Fix app_offline case sensitivity and error conditions

jkotalik:
Continuation of https://github.com/aspnet/AspNetCoreModule/issues/50 as customers still reported issues on case sensitivity.


jkotalik:
Doing a bit more verification.

PR #544, opened 2018-01-23
Reorder looking for app_offline and release the lock

jkotalik:
It seems that there can be a race between creating an app_offline.htm file and starting to monitor for it through the file watcher. After releasing the lock, I added a breakpoint and created an app_offline.htm file, then debug continued. The app was not taken offline, and the only way to trigger offline is to touch the app_offline file. This is especially painful on Azure websites. I switched the ordering of this to remove the issue. @pan-wang was there a good reason to have the call to app_offline outside of the lock?

### DIFF
--- a/src/AspNetCore/Src/applicationmanager.cxx
+++ b/src/AspNetCore/Src/applicationmanager.cxx
@@ -124,11 +124,13 @@ APPLICATION_MANAGER::GetApplicationInfo(
         }
 
         *ppApplicationInfo = pApplicationInfo;
-        ReleaseSRWLockExclusive(&m_srwLock);
-        fExclusiveLock = FALSE;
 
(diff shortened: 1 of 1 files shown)


pan-wang:
For App Service, the deployment issue is a different one. That is the application is running while customer put app_offline. Somehow the file change notification was missing even after re-deploy the app_offline file. For deployment, the app_offline is created by IIS WMSVC service. 
The race condition you described can always exists, i.e., before monitoring and file creation even after reordering.

jkotalik:
Ack, I realized now that file events occur completely separate from the request. Moving this may help a bit, but we should consider creating app_offline and adding it to the queue before we check if it is present.

jkotalik:
Closing as this doesn't help with the app_offline scenario.

PR #545, opened 2018-01-23
Run http request on separate thread.

jkotalik:
Also some changes to remove auth issues in the sample. 
Apparently OnAsyncCompletion will only receive an async notification status by calling PostCompletion, and not when OnExecuteRequestHandler returns. Therefore, we can simply Task.Run the request and everything will work.


src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -80,7 +80,7 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleRequest(IntPtr pI
 
             var context = server._iisContextFactory.CreateHttpContext(pInProcessHandler);
 
-            var task = context.ProcessRequestAsync();
+            var task = Task.Run(() => context.ProcessRequestAsync());
 
             // This should never fail
             if (task.IsCompleted)
```
Tratcher: You can remove this, it should never complete sync

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -80,14 +80,7 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleRequest(IntPtr pI
...
 
-            var task = context.ProcessRequestAsync();
-
-            // This should never fail
-            if (task.IsCompleted)
-            {
-                context.Dispose();
-                return ConvertRequestCompletionResults(task.Result);
-            }
+            var task = Task.Run(() => context.ProcessRequestAsync());
```
davidfowl: Get rid of the closure

PR #549, opened 2018-01-24
Send stderr logs to anonymous pipe instead of file when logging is disabled.

jkotalik:
Should be good to go here


src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -217,68 +225,86 @@ IN_PROCESS_APPLICATION::SetStdOut(
...
-            if (m_hLogFileHandle == INVALID_HANDLE_VALUE)
+            //
+            // CreatePipe for outputting stderr
+            if (!CreatePipe(&hStdErrReadPipe, &hStdErrWritePipe, &saAttr, 0 /*nSize*/))
             {
                 hr = HRESULT_FROM_WIN32(GetLastError());
                 goto Finished;
             }
+            // Verify it was inherited.
+            if (!SetHandleInformation(hStdErrReadPipe, HANDLE_FLAG_INHERIT, 0))
```
pan-wang: what's the purpose the this sethandleinformation?

jkotalik: This example: https://msdn.microsoft.com/en-us/library/windows/desktop/ms682499(v=vs.85).aspx used it for ensuring the read handle was not inherited.

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -544,27 +562,82 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
+        // Try to read std err logs for application exception.
+        // If anything fails here, ignore it.
+
+        if (m_pConfig->QueryStdoutLogEnabled())
+        {
+            if (m_hLogFileHandle != INVALID_HANDLE_VALUE)
+            {
+                if (!GetFileSizeEx(m_hLogFileHandle, &li))
+                {
+                    goto Finished;
```
pan-wang: already in Finished. why do you want to go back

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -544,27 +562,82 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
+        LARGE_INTEGER   li = { 0 };
+
+        // Try to read std err logs for application exception.
+        // If anything fails here, ignore it.
+
+        if (m_pConfig->QueryStdoutLogEnabled())
+        {
+            if (m_hLogFileHandle != INVALID_HANDLE_VALUE)
+            {
+                if (!GetFileSizeEx(m_hLogFileHandle, &li))
```
pan-wang: do we care about the failure here?

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -544,27 +562,82 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
         if (SUCCEEDED(strEventMsg.SafeSnwprintf(
-                ASPNETCORE_EVENT_INPROCESS_THREAD_EXIT_MSG,
-                m_pConfig->QueryApplicationPath()->QueryStr(),
-                m_pConfig->QueryApplicationPhysicalPath()->QueryStr(),
-                m_ProcessExitCode)))
+            ASPNETCORE_EVENT_INPROCESS_THREAD_EXIT_MSG,
+            m_pConfig->QueryApplicationPath()->QueryStr(),
+            m_pConfig->QueryApplicationPhysicalPath()->QueryStr(),
+            hr,
+            struStdErrLog.QueryStr())))
```
pan-wang: if struStdErrLog is empty, do we still want to log empty or hardcoded some info?

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -292,46 +311,43 @@ IN_PROCESS_APPLICATION::SetStdOut(
...
                 }
+                else
+                {
+                    //
+                    // CreatePipe for outputting stderr to the windows event log.
+
+                    // Read pipe
+                    if (!CreatePipe(&hStdErrReadPipe, &hStdErrWritePipe, &saAttr, 0 /*nSize*/))
+                    {
+                        hr = HRESULT_FROM_WIN32(GetLastError());
```
pan-wang: user did not ask for enabling log. this hr result is not consumed by other code path . You can remove this line

src/RequestHandler/inprocess/inprocessapplication.cpp
jkotalik: We don't (can't do anything about it). I'll just check that both calls succeeded before trying to read.

jkotalik:
@pan-wang 🆙 📅

jkotalik:
@pan-wang I modified this PR to read the pipe from a separate thread. The internal buffer size of the pipe can be as small as 4096 bytes. If the user writes over that number of bytes, the app will just hang. To fix this, we will have a thread that constantly reads from the pipe upto 4096 bytes and then closes the handle.

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -544,35 +656,108 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
+                    }
+                    else
+                    {
+                        dwFilePointer = SetFilePointer(m_hLogFileHandle, 0, NULL, FILE_BEGIN);
+                    }
+                    if (dwFilePointer != INVALID_SET_FILE_POINTER)
+                    {
+                        if (ReadFile(m_hLogFileHandle, pzFileContents, 4096, &dwNumBytesRead, NULL))
+                        {
+                            struStdErrLog.CopyA(pzFileContents, dwNumBytesRead);
```
jkotalik: hr.

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -544,35 +656,108 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
-        // error. the thread exits after application started
-        // Question: should we shutdown current worker process or keep the application in failure state?
-        // for now, we reccylce to keep the same behavior as that of out-of-process
-        if (m_fManagedAppLoaded)
+        else
         {
-            Recycle();
+            if (m_dwStdErrReadTotal > 0)
+            {
+                struStdErrLog.CopyA(m_pzFileContents, m_dwStdErrReadTotal);
```
jkotalik: hr.

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -544,35 +656,108 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
+            }
         }
-       
-        // error. the thread exits after application started
-        // Question: should we shutdown current worker process or keep the application in failure state?
-        // for now, we reccylce to keep the same behavior as that of out-of-process
-        if (m_fManagedAppLoaded)
+        else
         {
-            Recycle();
```
jkotalik: Explore leaving Recycle here if the managed loaded?

jkotalik:
@pan-wang updated

PR #558, opened 2018-01-25
Remove AspNetCoreHostingModel Property 

jkotalik:
We only need the the project capability; the decision to run in process/ out of process will be done either by the web.sdk (or something else to be determined).
@muratg we want to get this in preview1 as tooling needs it.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IIS/Microsoft.AspNetCore.Server.IIS.targets
+++ b/src/Microsoft.AspNetCore.Server.IIS/Microsoft.AspNetCore.Server.IIS.targets
@@ -1,17 +1,4 @@
 <Project>
-  <PropertyGroup>
-    <!--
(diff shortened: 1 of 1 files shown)


natemcmaster:
I don't think you should remove this. ProjectCapability is meant to define what a project _could_ support, not what it _does_ support. For example, most csproj files have the "FileGlobs" capability, but that doesn't mean a project _has_ to use globs.

natemcmaster:
Talked in person with @shirhatti. Clarified that the intention here is not to use ProjectCapability to set hosting model. The intention in this PR is to make the default hosting model OutOfProcess, even when M.A.S.IIS is a a package reference. In that case, this change is fine, although I have slight preference for just changing the default value of AspNetCoreModuleHostingModel to OutOfProcess to make this default explicit.

shirhatti:
:shipit:

jkotalik:
Closing as we are doing a merge with preview1

PR #559, opened 2018-01-26
Adding "Invoke where.exe to find dotnet.exe" to preview 1

jkotalik:
For #519


jkotalik:
This PR is exactly the same as #541 and has been approved for preview 1. Merging.

PR #562, opened 2018-01-28
Add IServerVariablesFeature and enable fetching variables directly

### LINKED ISSUE #515: Enable In Process tests once latest ANCM schema is added to Visual Studio
ANCM in process tests have to be disabled until we add the updated ANCM schema to Visual studio. The 1985 build of ANCM was supposed to have the schema change (current build on VS), however it was actually in the 1986 build of ANCM. We need this schema change to run in process tests because adding the new hostingModel="inprocess" to your web.config will fail at schema validation. As we rely on IISExpress to run tests, we can't run any in proc tests until this change is in. We could write a script to manually add the schema to all known locations, but we wouldn't be able to update AppVeyor and it could be overwritten every time we want to update Visual Studio. 
cc/ @shirhatti

natemcmaster:
Provides access to IIS server variables when hosting in-process. Provides a replacement for System.Web.HttpRequest.ServerVariables. This is useful for users who still have old IIS modules that set server variables such as AUTH_USER.

Note: this feature is only available during in-process hosting.

Side notes: also fixes #515


samples/NativeIISSample/Startup.cs
```diff
@@ -60,9 +61,35 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, IAuthent
...
+                var serverVariables = context.Features.Get<IServerVariablesFeature>();
+
+                if (serverVariables != null)
+                {
+                    // accessing IIS server variables
+                    await context.Response.WriteAsync("Server Variables:" + Environment.NewLine);
+
+                    for (var i = 0; i < IISServerVarNames.Length; i++)
+                    {
+                        var varName = IISServerVarNames[i];
```
Tratcher: Consider providing a `GetIISServerVariable(string)` extension method.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Features.cs
```diff
@@ -236,6 +241,10 @@ internal void FastFeatureSet(Type key, object feature)
             {
                 throw new InvalidOperationException("Cannot set IISHttpContext in feature collection");
             };
+            if (key == IServerVariablesFeature)
+            {
+                throw new InvalidOperationException("Cannot set IServerVariablesFeature in feature collection");
```
Tratcher: Most feature should be settable so they can be wrapped/shimmed. IISHttpContextType is the exception here as it's internal. Use a member field and initialize it to `this`.

src/RequestHandler/managedexports.cxx
```diff
@@ -43,6 +43,32 @@ http_get_raw_response(
...
+{
+    PCWSTR pszVariableValue;
+    DWORD cbLength;
+    HRESULT hResult = pInProcessHandler
+        ->QueryHttpContext()
+        ->GetServerVariable(pszVariableName, &pszVariableValue, &cbLength);
+
+    if (hResult == S_OK && cbLength > 0 && pszVariableValue[0] != '\0')
+    {
+        WCHAR* pwszReturn = NULL;
```
jkotalik: I personally prefer to return BSTR from these managed interop functions: https://msdn.microsoft.com/en-us/library/windows/desktop/ms221069(v=vs.85).aspx. They are type def'd to be WCHAR*. 

This function can also be simplified by calling SysAllocString().

```c++
EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
HRESULT
http_get_server_variable
(
    _In_ IN_PROCESS_HANDLER* pInProcessHandler,
    _In_ PCSTR pszVariableName,
    _Out_ BSTR* pwszRetVal
)
{
    PCWSTR  pszVariableValue;
    DWORD   cbLength;
    BSTR    pwzRetVal = NULL;
    HRESULT hr = pInProcessHandler
            ->QueryHttpContext()
            ->GetServerVariable(pszVariableName, &pszVariableValue, &cbLength);

    if (FAILED(hr) ||
        cbLength == 0 ||
        pszVariableValue[0] != L'\0') 
    {
        goto Finished;
    }

    *pwszRetVal = SysAllocString(pszVariableValue);

    if (*pwszRetVal == NULL)
    {
        hr = E_OUTOFMEMORY;
        goto Finished;
    }

Finished:

    return hr;
}
```

samples/NativeIISSample/Startup.cs
```diff
@@ -60,9 +61,35 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, IAuthent
...
+                        await context.Response.WriteAsync(varName + ": " + serverVariables[varName] + Environment.NewLine);
+                    }
+                }
             });
         }
 
+        private static readonly string[] IISServerVarNames =
+        {
+            "AUTH_TYPE",
```
jkotalik: We should consider putting IIS Server Variable Names in HttpAbstractions. UrlRewrite uses them too: https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs

davidfowl: Uhhh nah. These are IIS specific.  If you want to share code, crate a sources package but I don’t think it’s worth it

src/RequestHandler/managedexports.cxx
jkotalik: Can you point me to the CLR code that does that? I wonder if SysAllocString uses CoTaskMemAlloc under the hood or if is compatible.

src/RequestHandler/managedexports.cxx
```diff
@@ -43,6 +43,32 @@ http_get_raw_response(
     return pInProcessHandler->QueryHttpContext()->GetResponse()->GetRawHttpResponse();
 }
 
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+WCHAR*
+http_get_server_variable(
```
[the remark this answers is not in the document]
natemcmaster: Had trouble getting this to play nice. Kept getting AV exceptions trying to implement it this way...the implementation you stubbed out in the other comment didn't work.

samples/NativeIISSample/Startup.cs
natemcmaster: Added

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Features.cs
natemcmaster: Updated this too

src/RequestHandler/managedexports.cxx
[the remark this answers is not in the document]
jkotalik: I'll checkout your changes and see if I can get it working. I remember I was running into issue with that too.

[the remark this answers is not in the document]
jkotalik: So I was incorrect in saying BSTR == WCHAR*. BSTRs also include a length prefix to it. So SysAllocString can only work with BSTRs and LPWSTRs with the way you are allocating it. Also from what I can tell, the CLR will call SysFreeString on BSTRs.  To fix this, you can either use SysAllocString or make the return parameter LPWSTR in managed and native.

natemcmaster:
@jkotalik I think this incidentally resolves #515 by detecting if the ancm schema supports the 'hostingModel' attribute.

src/RequestHandler/managedexports.cxx
natemcmaster: I was reading this doc: https://docs.microsoft.com/en-us/dotnet/framework/interop/default-marshaling-behavior. I updated to use SysAllocString instead and it appears to be OK.

src/RequestHandler/managedexports.cxx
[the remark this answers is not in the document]
natemcmaster: Got the HRESULT piece working with BSTR. My c++ is weak. Had to remember how pointers work.

jkotalik:
I'd assume these tests will never run on AppVeyor until it has the lastest version of the schema?

natemcmaster:
> I'd assume these tests will never run on AppVeyor until it has the lastest version of the schema?

That's basically the idea. This detection should allow the tests to run once ANCM is updated. i.e. our CI agents should be running these tests.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.FeatureCollection.cs
```diff
@@ -234,6 +235,20 @@ ClaimsPrincipal IHttpAuthenticationFeature.User
...
+        string IServerVariablesFeature.this[string variableName]
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(variableName))
+                {
+                    return null;
+                }
+
+                int hr = NativeMethods.http_get_server_variable(_pInProcessHandler, variableName, out var value);
```
jkotalik: I would do
```
return hr == 0 ? value : null;
```

test/IISIntegration.IISServerFunctionalTests/Utilities/IISExpressSupportsInProcessHostingAttribute.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            static AncmSchema()
+            {
+                if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    SkipReason = "IIS Express tests can only be run on Windows";
+                    return;
+                }
+
+                var ancmConfigPath = Path.Combine(
```
jkotalik: Nit: we may want to check x64/x86 based on bitness of the process. Even though the files are the same, the bitness will correspond to the config that is loaded, ex x64 will use ProgramFiles/... and x86 will use ProgramFiles/x86....

src/Microsoft.AspNetCore.Server.IISIntegration/IServerVariablesFeature.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// <summary>
+    /// This feature provides access to request server variables set.
+    /// <para>
+    /// This feature is only available when hosting ASP.NET Core in-process with IIS or IIS Express.
+    /// </para>
+    /// </summary>
+    /// <remarks>
+    /// For a list of common server variables available in IIS, see http://go.microsoft.com/fwlink/?LinkId=52471.
+    /// </remarks>
+    public interface IServerVariablesFeature
```
davidfowl: One down sides to supporting this very minimal interface, you can't enumerate the server variables.

src/Microsoft.AspNetCore.Server.IISIntegration/IISHttpContextExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <param name="context">The http context for the request.</param>
+        /// <param name="variableName">The name of the variable.</param>
+        /// <returns>
+        /// <c>null</c> if the feature does not support the <see cref="IServerVariablesFeature"/> feature.
+        /// May return null or empty if the variable does not exist or is not set.
+        /// </returns>
+        /// <remarks>
+        /// For a list of common server variables available in IIS, see http://go.microsoft.com/fwlink/?LinkId=52471.
+        /// </remarks>
+        public static string GetIISServerVariable(this HttpContext context, string variableName)
```
davidfowl: I don't feel like this should show up on the HttpContext for every single ASP.NET Core 2.1 application. If we're going to keep it, lets change the namespace.

davidfowl: > Microsoft.AspNetCore.Server.IIS

Yes, that can be the namespace.

test/IISIntegration.IISServerFunctionalTests/Utilities/IISExpressSupportsInProcessHostingAttribute.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Xml.Linq;
+using Microsoft.AspNetCore.Testing.xunit;
+
+namespace Microsoft.AspNetCore.Server.IIS.FunctionalTests
+{
+    [AttributeUsage(AttributeTargets.Method)]
+    public sealed class IISExpressSupportsInProcessHostingAttribute : Attribute, ITestCondition
```
davidfowl: Very nice!

samples/NativeIISSample/Startup.cs
Tratcher: And this is only a sample.

src/Microsoft.AspNetCore.Server.IISIntegration/IServerVariablesFeature.cs
Tratcher: @pan-wang IIS doesn't have an API for enumerating server variables, correct? I was told System.Web makes it work by keeping a static list of variables to try.

test/IISIntegration.IISServerFunctionalTests/FeatureCollectionTests.cs
```diff
@@ -22,7 +23,8 @@ public FeatureCollectionTest(ITestOutputHelper output) : base(output)
         {
         }
 
-        [Theory(Skip = "See https://github.com/aspnet/IISIntegration/issues/424")]
+        [ConditionalTheory]
+        [IISExpressSupportsInProcessHosting]
```
Tratcher: You can put this attribute at the assembly level.

test/IISTestSite/Startup.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+
+namespace IISTestSite
```
Tratcher: ServerVariableTestSite? Or do you plan on expanding this?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.FeatureCollection.cs
natemcmaster: Null

test/IISTestSite/Startup.cs
natemcmaster: From what I can tell, most of the tests in this project don't need to deploy a new instance of the IISTestSite. I added it this way and opened #565 to simplify the functional testing to share a common instance of IISTestSite.

src/Microsoft.AspNetCore.Server.IISIntegration/IServerVariablesFeature.cs
davidfowl: Yea you can get the list from IIS. Helios does the same thing. It’s not a deal breaker and can be added in the future if required

test/IISIntegration.IISServerFunctionalTests/Utilities/IISExpressSupportsInProcessHostingAttribute.cs
natemcmaster: I'll just use Environment.SpecialFolder.ProgramFiles. This defaults to the programfiles that matches the bitness of the PC.

test/IISTestSite/Startup.cs
jkotalik: We can overload the single published site with all of our tests to reduce unnecessary deployments. We can also modify the deployer to use different configurations with an already published application.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.FeatureCollection.cs
natemcmaster: Updated the return to check hr.

natemcmaster:
:up: :date: I think I've addressed all feedback now. I don't think it's too hard to add enumeration of all server vars, so I can take a crack at that too if we think there is marginal value in it.

jkotalik:
I don't think there is an incredible amount of value in it currently. We can make a follow up issue if we want this functionality.

natemcmaster:
Added https://github.com/aspnet/IISIntegration/pull/562/commits/6461c2eb0227e55e8d798ec27723a5dd967d8b2c so this AppVeyor should actually start running InProcess tests on PRs

natemcmaster:
@jkotalik opened https://github.com/aspnet/IISIntegration/issues/569 to follow up on enumerating all server variables, if anyone cares enough

PR #564, opened 2018-01-29
 Default to using program files if where.exe doesn't find dotnet.exe

jkotalik:
Per discussion with @shirhatti and @glennc, we will always fallback to ProgramFiles if where.exe doesn't find dotnet.exe. This is for the azure websites scenario where we are considering not putting the x64 version of dotnet on the path.
@muratg we want this for preview1 in case we don't get the chance to update Azure websites scenarios before preview2.

### DIFF
--- a/src/CommonLib/hostfxr_utility.cpp
+++ b/src/CommonLib/hostfxr_utility.cpp
@@ -136,7 +136,6 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
         if (FAILED(hr = HOSTFXR_UTILITY::FindDotnetExePath(&struExeLocation))) {
             goto Finished;
         }
-
     }
 
     if (FAILED(hr = struExeLocation.SyncWithBuffer()) ||
@@ -363,6 +362,7 @@ HOSTFXR_UTILITY::FindDotnetExePath(
     DWORD               dwNumBytesRead;
     DWORD               dwFilePointer;
     DWORD               dwBinaryType;
+    DWORD               dwPathSize = MAX_PATH;
     INT                 index = 0;
     INT                 prevIndex = 0;
     BOOL                fResult = FALSE;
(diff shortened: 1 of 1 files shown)


src/CommonLib/hostfxr_utility.cpp
```diff
@@ -429,105 +429,130 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
-        if (FAILED(hr = struDotnetSubstring.Copy(struDotnetLocationsString.QueryStr(), index - prevIndex)))
+        if (FAILED(struDotnetSubstring.Resize(dwPathSize)))
         {
             goto Finished;
         }
-        prevIndex = index;
 
-        if (!GetBinaryTypeW(struDotnetSubstring.QueryStr(), &dwBinaryType))
+        // Program files will changes based on the 
+        dwNumBytesRead = GetEnvironmentVariable(L"ProgramFiles", struDotnetSubstring.QueryStr(), dwPathSize);
```
shirhatti: You're only trying 64-bit fallback?

davidfowl: Shouldn't it look in the appropriate place based on the bitness?

Tratcher: %ProgramFiles% is bitness aware

davidfowl: Ah I thought this was the path, I didn't see the GetEnvironmentVariable. Hopefully we have tests for this 😄

jkotalik: Yeah as @Tratcher said, PF is bitness aware. 

@davidfowl native tests are coming soon ™️

jkotalik: image
versus
image

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -429,105 +429,130 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
-    // where.exe will return 1 if the file is not found
+    // where.exe will return 0 on success, 1 if the file is not found
     // and 2 if there was an error. Check if the exit code is 1 and set
     // a new hr result saying it couldn't find dotnet.exe
     // 
-    if (!GetExitCodeProcess(processInformation.hProcess, &dwExitCode) ||
-        dwExitCode != 0)
+    if (!GetExitCodeProcess(processInformation.hProcess, &dwExitCode))
+    {
+        hr = HRESULT_FROM_WIN32(GetLastError());
```
pan-wang: fallback instead of finish?

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -429,105 +429,130 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
-    }
-    else
-    {
-        SYSTEM_INFO systemInfo;
-        GetNativeSystemInfo(&systemInfo);
-        fIsCurrentProcess64Bit = systemInfo.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_AMD64;
+            if (GetBinaryTypeW(struDotnetSubstring.QueryStr(), &dwBinaryType) &&
+                fIsCurrentProcess64Bit == (dwBinaryType == SCS_64BIT_BINARY)) {
+                // Found a valid dotnet.
+                struDotnetPath->Copy(struDotnetSubstring);
```
pan-wang: copy may fail. please check hr

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -429,105 +429,130 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
+    // Look in ProgramFiles
     while (!fFound)
     {
-        index = struDotnetLocationsString.IndexOf(L"\r\n", prevIndex);
-        if (index == -1)
-        {
-            break;
-        }
-        if (FAILED(hr = struDotnetSubstring.Copy(struDotnetLocationsString.QueryStr(), index - prevIndex)))
+        if (FAILED(struDotnetSubstring.Resize(dwPathSize)))
```
pan-wang: missing hresult

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -429,105 +429,130 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
+                FAILED(hr = struDotnetSubstring.Append(L"\\dotnet\\dotnet.exe")))
+            {
+                goto Finished;
+            }
+            if (!UTILITY::CheckIfFileExists(struDotnetSubstring.QueryStr()))
+            {
+                hr = HRESULT_FROM_WIN32( GetLastError() );
+                goto Finished;
+            }
             struDotnetPath->Copy(struDotnetSubstring);
```
pan-wang: Copy may fail here

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -429,105 +429,130 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
             fFound = TRUE;
         }
     }
 
-    if (!fFound)
```
pan-wang: Do we have window event log for not being able to find dotnet.exe

jkotalik: I believe it is part of this PR pending a bit of refactoring: https://github.com/aspnet/IISIntegration/pull/533/files#diff-093466e281c521516f40989e2d5cb1caR302

jkotalik:
I think we will wait on this until preview 2, correct @shirhatti ?

Issue #550, opened 2018-01-24
ANCM: Inprocess: If App fails becasue hostfxr.dll does not exist, we should make error event that the hostfx file not found


jhkimnew:
1. Rename hostfxr.dll with a different file name such as C:\Program Files\dotnet\host\fxr\2.1.0-preview1-26110-02\_hostfxr.dll

2. Send a request

EXPECTED:

There is an windows event error which explains the hostfx.dll file not found.

ACTUAL:

There is an error event but that does not explain the root cause.
For example, I got this error but this does not give any hint on the root cause.

Application '/LM/W3SVC/1/ROOT/Test123' with physical root 'C:\inetpub\wwwroot\Test123\' hit unexpected managed background thread exit, ErrorCode = '0x0.

jkotalik:
Will be fixed by #533.

muratg:
@jkotalik #533 is merged. Please verify if this can be closed.

Issue #548, opened 2018-01-24
ANCM: Inprocess: Appverifier error "A heap allocation was leaked. " when recycling IIS apppool while Websocket connection is opened


jhkimnew:
1. Open websocket connection to an aspnetcore app which is running in inprocess mode
2. Recycle IIS worker process where the application was running

EXPECTED:

No error

ACTUAL:

Following AppVerifier error is shown.

```
=======================================
VERIFIER STOP 0000000000000900: pid 0x19F0: A heap allocation was leaked. 

                0000007211AD2FE0 : Address of the leaked allocation. Run !heap -p -a <address> to get additional information about the allocation.
                0000007271ED3370 : Address to the allocation stack trace. Run dps <address> to view the allocation stack.
                00000072130ACFD8 : Address of the owner dll name. Run du <address> to read the dll name.
                00007FFA21EF0000 : Base of the owner dll. Run .reload <dll_name> = <address> to reload the owner dll. Use 'lm' to get more information about the loaded and unloaded modules.


=======================================
This verifier stop is continuable.
After debugging it use `go' to continue.

=======================================

(19f0.1bc8): Break instruction exception - code 80000003 (first chance)
*** ERROR: Symbol file could not be found.  Defaulted to export symbols for C:\Windows\SYSTEM32\vrfcore.dll - 
*** ERROR: Module load completed but symbols could not be loaded for C:\Windows\SYSTEM32\vfbasics.dll
vrfcore!VerifierStopMessageEx+0x836:
00007ffa`35562626 cc              int     3

0:034> dps 0000007271ED3370
00000072`71ed3370  00000000`00000000
00000072`71ed3378  00190000`0000c801
00000072`71ed3380  00007ffa`3df51132 ntdll!RtlpNtMakeTemporaryKey+0x3d2
00000072`71ed3388  00007ffa`3df0a56a ntdll!memset+0x1542a
00000072`71ed3390  00007ffa`3de9086d ntdll!RtlAllocateHeap+0x17d
00000072`71ed3398  00007ffa`3556bc22 vrfcore!VerifierSetAPIClassName+0xc2
00000072`71ed33a0  00007ffa`308e7d3d vfbasics+0x17d3d
00000072`71ed33a8  00007ffa`21efea80 aspnetcore!BUFFER_T<wchar_t,1>::Resize+0xd0 [d:\gitroot\iisintegration\src\iislib\buffer.h @ 152]
[...]

PR #567, opened 2018-01-29
Reference and Dereference application in inprocess application.

jkotalik:
Take a reference to the application in each of the handlers such that when a request is using the application, the application won't be destroyed. 
Part of fixing #550 and #548. We also need to implement shutdown with pending on each request.

### DIFF
--- a/src/RequestHandler/inprocess/inprocessapplication.cpp
+++ b/src/RequestHandler/inprocess/inprocessapplication.cpp
@@ -43,86 +43,89 @@ IN_PROCESS_APPLICATION::Recycle(
     VOID
 )
 {
-    if (m_fInitialized)
+
+    DWORD    dwThreadStatus = 0;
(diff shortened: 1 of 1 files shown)


src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -43,86 +43,89 @@ IN_PROCESS_APPLICATION::Recycle(
...
-            (m_pHttpServer->GetAdminManager() != NULL))
-        {
-            // IIS scenario.
-            // notify IIS first so that new request will be routed to new worker process
-            m_pHttpServer->RecycleProcess(L"AspNetCore Recycle Process on Demand");
-        }
+    AcquireSRWLockExclusive(&m_srwLock);
 
-        m_fRecycleProcessCalled = TRUE;
+    if (!m_fInitialized)
```
jkotalik: Removing a bit of nesting here.

jkotalik:
Closing as it is part of #576

Issue #565, opened 2018-01-29
Reduce duplicate code in ANCM in-process tests


natemcmaster:
Follow up to #562 

 - Update tests to use the same collection fixture to reduce duplication
 - Figure out how to forward logging from a test fixture using ApplicationDeployer to ITestOutputHelper. Fixture currently uses `NullLoggerFactory`. This probably requires capturing the logs during fixture setup and duplicating it to the ITestOutputHelper when tests actually run.

cc @jkotalik

PR #571, opened 2018-01-29
Update tests to use test fixture

jkotalik:
For #565 and the current failures on CI


test/IISIntegration.IISServerFunctionalTests/Utilities/IISTestSiteFixture.cs
```diff
@@ -34,15 +34,16 @@ public IISTestSiteFixture()
             };
 
             _deployer = ApplicationDeployerFactory.Create(deploymentParameters, NullLoggerFactory.Instance);
```
natemcmaster: Any thoughts on this? I didn't see a clean way to forward to ITestOutputHelper. ITestOutputHelper is specific to each test-case.

jkotalik: I'll take a look into it, but from my initial glance I didn't see a great way either.

natemcmaster: It's really only needed when trying to debug why test failed. But when tests fail, this output is really useful.

The only solution I could think of was to have the deployer attach a logger that captures all output during the DeployAsync() and saves it in memory. Add an API to the fixture to copy this output to ITestOutputHelper for each test run

jkotalik: I think we can expose the Logger through the Deployer https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs#L39. For logging the HttpClient, we can maybe create a new factory/ existing logger and modify these https://github.com/aspnet/Hosting/blob/bc4a514d9e4da1c01e369f59306ec3da0d7d0b41/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/LoggingHandler.cs, https://github.com/aspnet/Hosting/blob/1ae1ff01a3573832d111c83ac2cc185d03087c1b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentResult.cs#L66

Tratcher:
Weren't there custom .config files being used for some of the apps? E.g. Auth?

jkotalik:
I believe the .config file already has Auth stuff inside of it, so inproc doesn't need custom config yet.

test/IISIntegration.IISServerFunctionalTests/Utilities/IISTestSiteFixture.cs
jkotalik: No matter what it will require changes to the deployers; I'd prefer to make a separate PR to add logging back in.

natemcmaster: Ok with me. So long as it gets added back.

PR #574, opened 2018-01-31
Update dependencies

ryanbrandenburg:
To react to the new versions of things like `System.Memory`.


pakrym:
Let's kick off a new Update repos build instead.

ryanbrandenburg:
Update repos is a couple hours out from being able to run, so @pakrym and I agreed to merge this.

PR #576, opened 2018-02-02
Implements Server Shutdown

jkotalik:
Native initiated shutdown:
- Calls managed shutdown handler and waits on the main thread to exit with a timeout.
- If the native shutdown is never triggered, we call threadexit on main with the CTRL_C signal. We cannot call back into managed at this point as the app is in an unrecoverable state.
- After shutdown is triggered, native will destruct the Application Info and Application. 
- Once the destructor is called for the application, the worker process will be recycled.

Managed initiated shutdown:
- StopAsync will first call into native to set m_fShutdown, which will reject any incoming requests. The AsyncCompletionHandler can still be used for finalizing requests.
- After all requests are drained, we call back into native again to set another bool to notify managed is shutting down. At this point, native cannot callback into managed.
- By setting m_fShutdown, if native receives a shutdown notification, it will simply ignore it as managed has already commit to shutting down. 
- Once main returns, we need to recycle the app pool, but we cannot directly call Recycle here, as there are issues with double locking and calling recycle multiple times.. In order to remove the application, we need to go to the application manager, which lives in the shim (can't reference from the requesthandler.dll). We do a clever trick to remove the application from the application_manager, by notifying the IHttpServer that there has been a configuration change. This will cause the global notification OnGlobalApplicationStop, which will allow us to remove the application from the manager safely and then recycle the worker process.

Making this iteration style as it is a fairly complicated change.

For #580


Tratcher:
> If the native shutdown is never triggered, we call threadexit on main with the CTRL_C signal.

Why is this step necessary? CTRL_C triggers the same app shutdown code path as your first action.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -56,15 +62,51 @@ public Task StartAsync<TContext>(IHttpApplication<TContext> application, Cancell
 
         public Task StopAsync(CancellationToken cancellationToken)
         {
-            // TODO: Drain pending requests
+            // First call back into native saying "DON'T SEND ME ANY MORE REQUESTS"
+            NativeMethods.http_stop_incoming_requests();
```
Tratcher: Move this down to line 83. The RegisterCancellation code below is designed to prevent StopAsync from executing twice.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -101,11 +148,18 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS OnAsyncCompletion(IntPt
...
-        private static void CompleteRequest(IISHttpContext context, Task<bool> completedTask)
+        private static void CompleteRequest(IISHttpContext context, IISHttpServer server, Task<bool> completedTask)
         {
             // Post completion after completing the request to resume the state machine
             context.PostCompletion(ConvertRequestCompletionResults(completedTask.Result));
 
+            if (Interlocked.Decrement(ref server._outstandingRequests) == 0 && server.Stopping)
+            {
+                // All requests have been drained.
+                NativeMethods.http_stop_calls_into_mananged();
```
Tratcher: You also need to call this on line 92 if there are no active requests when you initiate shutdown.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -77,12 +119,17 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleRequest(IntPtr pI
         {
             // Unwrap the server so we can create an http context and process the request
             var server = (IISHttpServer)GCHandle.FromIntPtr(pvRequestContext).Target;
+            if (server != null)
```
davidfowl: Why bother checking for null here? The next line immediately dereferences server

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -101,11 +148,18 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS OnAsyncCompletion(IntPt
...
+        private static void CompleteRequest(IISHttpContext context, IISHttpServer server, Task<bool> completedTask)
         {
             // Post completion after completing the request to resume the state machine
             context.PostCompletion(ConvertRequestCompletionResults(completedTask.Result));
 
+            if (Interlocked.Decrement(ref server._outstandingRequests) == 0 && server.Stopping)
+            {
+                // All requests have been drained.
+                NativeMethods.http_stop_calls_into_mananged();
+                server._shutdownSignal.TrySetResult(0);
```
davidfowl: nit: Why do you pass 0 instead of null?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -77,12 +119,17 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleRequest(IntPtr pI
...
+                Interlocked.Increment(ref server._outstandingRequests);
+            }
 
             var context = server._iisContextFactory.CreateHttpContext(pInProcessHandler);
 
             var task = Task.Run(() => context.ProcessRequestAsync());
 
-            task.ContinueWith((t, state) => CompleteRequest((IISHttpContext)state, t), context);
+            // TODO make this use continuations correctly instead of closures.
+            task.ContinueWith((t, state) => CompleteRequest(context, (IISHttpServer)state, t), server);
```
davidfowl: You just introduced a closure. I'd rather you store a reference to the IISHttpServer on the IISHttpContext

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -56,15 +62,51 @@ public Task StartAsync<TContext>(IHttpApplication<TContext> application, Cancell
...
         public Task StopAsync(CancellationToken cancellationToken)
         {
-            // TODO: Drain pending requests
+            // First call back into native saying "DON'T SEND ME ANY MORE REQUESTS"
+            NativeMethods.http_stop_incoming_requests();
+            void RegisterCancelation()
+            {
+                cancellationToken.Register(() =>
+                {
+                    if (Interlocked.Exchange(ref _shutdownSignalCompleted, 1) == 0)
```
davidfowl: What is this flag for? Why isn't TrySetResult enough?

samples/NativeIISSample/Startup.cs
```diff
@@ -15,7 +15,7 @@ namespace NativeIISSample
     public class Startup
     {
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IHostingEnvironment env, IAuthenticationSchemeProvider authSchemeProvider)
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, IAuthenticationSchemeProvider authSchemeProvider, IApplicationLifetime appLifetime)
```
davidfowl: What's this for?

tools/move_ancm_to_inetsrv.ps1
```diff
@@ -0,0 +1,8 @@
+$script_path = $myinvocation.mycommand.path
+$script_folder = Split-Path $script_path -Parent
+$aspnetcoredll = Join-Path $script_folder '..\src\AspNetCore\bin\Debug\x64\aspnetcore.dll'
+$aspnetcorerhdll = Join-Path $script_folder '..\src\RequestHandler\bin\Debug\x64\aspnetcorerh.dll'
+echo $aspnetcoredll
+echo $aspnetcorerhdll
+Copy-Item -Force -Path  $aspnetcoredll -Destination 'C:\Windows\System32\inetsrv\aspnetcore.dll'
```
davidfowl: uhhh

jkotalik: Why not?

src/AspNetCore/Inc/applicationinfo.h
```diff
@@ -143,6 +143,19 @@ class APPLICATION_INFO
...
     }
 
+    VOID
+    ShutDown(
+        VOID
+    );
+
+    static
+    VOID
+    ShutDownWrapper(APPLICATION_INFO* pApplicationInfo, PVOID context)
```
pan-wang: what's the context reserved for?

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -23,8 +23,7 @@ APPLICATION_INFO::~APPLICATION_INFO()
 
     if (m_pApplication != NULL)
     {
-        // shutdown the application
-        m_pApplication->ShutDown();
+        // Shutdown should have already been called if gracefully shutting down.
```
pan-wang: if you believe shutdown has already been done, maybe you need a flag to address it. Otherwise, if the shutdown is in the middle with another thread and this thread deference it and delete it, some AV may happen.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -78,6 +77,21 @@ APPLICATION_INFO::Initialize(
...
 
+VOID
+APPLICATION_INFO::ShutDown()
+{
+    if (m_pApplication != NULL)
+    {
+        AcquireSRWLockExclusive(&m_srwLock);
+        if (m_pApplication != NULL)
+        {
+            m_pApplication->ShutDown();
```
pan-wang: I think we need deference the application here.

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -171,6 +173,19 @@ APPLICATION_MANAGER::GetApplicationInfo(
...
+        else if (m_fInShutdown)
+        {
+            if (SUCCEEDED(strEventMsg.SafeSnwprintf(
+                ASPNETCORE_EVENT_APP_IN_SHUTDOWN_MSG,
+                pszApplicationId,
+                pConfig->QueryHostingModel())))
+            {
+                UTILITY::LogEvent(g_hEventLog,
+                    EVENTLOG_ERROR_TYPE,
+                    ASPNETCORE_EVENT_MIXED_HOSTING_MODEL_ERROR,
```
pan-wang: the error message id is wrong here

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -124,10 +124,12 @@ APPLICATION_MANAGER::GetApplicationInfo(
         }
 
         *ppApplicationInfo = pApplicationInfo;
-        ReleaseSRWLockExclusive(&m_srwLock);
-        fExclusiveLock = FALSE;
 
+        // Listen to app offline before releasing the lock, which will avoid races here.
         pApplicationInfo->StartMonitoringAppOffline();
+        ReleaseSRWLockExclusive(&m_srwLock);
```
pan-wang: how can we make sure after releasing the lock, an appoffline happens immediately and the applicationinfo is not release prematurely

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -171,6 +173,19 @@ APPLICATION_MANAGER::GetApplicationInfo(
                     strEventMsg.QueryStr());
             }
         }
+        else if (m_fInShutdown)
+        {
+            if (SUCCEEDED(strEventMsg.SafeSnwprintf(
```
pan-wang: how could this in shutdown happen?
you may want to create a different error message for such situation

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -225,24 +248,21 @@ APPLICATION_MANAGER::RecycleApplication(
...
         AcquireSRWLockExclusive(&m_srwLock);
 
-        // clean up the hash table so that the application will be informed on shutdown
-        m_pApplicationInfoHash->Clear();
+        m_pApplicationInfoHash->Apply(APPLICATION_INFO::ShutDownWrapper, NULL);
 
         ReleaseSRWLockExclusive(&m_srwLock);
     }
-
-    // stop filewatcher monitoring thread
```
pan-wang: where do you clean up mpFileWatcher?

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -30,99 +30,121 @@ IN_PROCESS_APPLICATION::~IN_PROCESS_APPLICATION()
...
 
-        if (!m_pHttpServer->IsCommandLineLaunch() &&
-            !m_fRecycleProcessCalled &&
-            (m_pHttpServer->GetAdminManager() != NULL))
-        {
-            // IIS scenario.
-            // notify IIS first so that new request will be routed to new worker process
-            m_pHttpServer->RecycleProcess(L"AspNetCore Recycle Process on Demand");
-        }
+    m_fShutdownCalledFromNative = TRUE;
```
pan-wang: why you have m_fShutdownCalledFromNative = TRUE while still check !m_fshutdwonCalledFromManaged?
hot could a shutdown is initiated from managed?

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -30,99 +30,121 @@ IN_PROCESS_APPLICATION::~IN_PROCESS_APPLICATION()
...
 
-        m_fRecycleProcessCalled = TRUE;
+    if (!m_fShutdownCalledFromManaged)
+    {
```
pan-wang: Could you please add more comments about the logic of handling shutdown here?

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -564,15 +623,16 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
-            Recycle();
-        }
+
+        //
+        // No matter what we need to cause recycle to be called on the worker process. We also want to remove the application from the application
+        // manager as by dereferencing it in the mananger will trigger recycle. We will post a notify configuration change here, allowing us to receive
+        // a global notification event, which would eventually cause the application to be dereferenced. This is only
+        // done if shutdown has not been called from native.
+        //
+        g_pHttpServer->NotifyConfigurationChange(m_pConfig->QueryConfigPath()->QueryStr());
```
pan-wang: why you want to treat this error as configuration change? nothing changed with configuration.

src/RequestHandler/inprocess/inprocessapplication.h
```diff
@@ -57,6 +52,28 @@ class IN_PROCESS_APPLICATION : public APPLICATION
         IN_PROCESS_HANDLER* pInProcessHandler
     );
 
+    VOID
+    StopCallsIntoManaged(
+        VOID
+    )
+    {
+        AcquireSRWLockExclusive(&m_srwLock);
```
pan-wang: what's the purpose of lock here? is volatile or interlockcomparedexchange not sufficient?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -77,11 +117,13 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleRequest(IntPtr pI
         {
             // Unwrap the server so we can create an http context and process the request
             var server = (IISHttpServer)GCHandle.FromIntPtr(pvRequestContext).Target;
+            Interlocked.Increment(ref server._outstandingRequests);
 
             var context = server._iisContextFactory.CreateHttpContext(pInProcessHandler);
 
             var task = Task.Run(() => context.ProcessRequestAsync());
 
+            // TODO make this use continuations correctly instead of closures.
```
davidfowl: You can get rid of this comment

src/AspNetCore/Inc/applicationinfo.h
jkotalik: It isn't used at all, it is called from the HASH_TABLE.

src/AspNetCore/Src/applicationinfo.cpp
jkotalik: I like adding a flag for knowing if we already called shutdown in the application info. The idea behind this PR is I wanted to separate Shutdown from the destructor entirely, because shutdown can be triggered by either native or managed. I can discuss with you tomorrow about this.

src/AspNetCore/Src/applicationinfo.cpp
jkotalik: Actually, we don't want to set the application to null here.

src/AspNetCore/Src/applicationmanager.cxx
jkotalik: In the destructor, but I think you are right.

src/AspNetCore/Src/applicationmanager.cxx
jkotalik: Shutdown + dereferencing should avoid this issue. Again I can talk to you tomorrow about this.

src/AspNetCore/Src/applicationmanager.cxx
jkotalik: This shutdown can happen through a global event. We don't want to call shutdown multiple times if this happens.

src/RequestHandler/inprocess/inprocessapplication.cpp
jkotalik: We need to remove the application from the application manager somehow, such that we can recycle the worker process eventually.

src/RequestHandler/inprocess/inprocessapplication.cpp
jkotalik: Shutdown from managed can happened either when Main returns or when the server calls IApplicationLifetime.Shutdown(), which will trigger the managed server shutdown.

src/RequestHandler/inprocess/inprocessapplication.cpp
jkotalik: I may be able to change the logic to call Recycle here instead, however I am not sure how to propagate that the server has shutdown to the ApplicationManager.

src/RequestHandler/inprocess/inprocessapplication.cpp
jkotalik: done

jkotalik:
Functionally, I am seeing all of the shutdown scenarios work. I did some stress testing by running wrk against an IIS site, then calling recycle on the app pool constantly. Testing will be difficult, but I may give it a try later.

tools/move_ancm_to_inetsrv.ps1
davidfowl: Because it overwrites your global ANCM installation and uninstall will probably not work if you do this.

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -196,14 +211,26 @@ APPLICATION_MANAGER::RecycleApplication(
...
     if (FAILED(hr))
     {
         goto Finished;
     }
-    AcquireSRWLockExclusive(&m_srwLock);
-    m_pApplicationInfoHash->DeleteKey(&key);
+
+    AcquireSRWLockExclusive(&m_srwLock); 
+    m_pApplicationInfoHash->FindKey(&key, &pApplicationInfo); // FindKey will reference here.
+    ReleaseSRWLockExclusive(&m_srwLock);
```
pan-wang: there could be a race condition that shutdown is called more than once. You can volatile flag to guard it

src/AspNetCore/Src/applicationinfo.cpp
pan-wang: the shutdown of n application may take long time while these could still be incoming requests. I did not see any code to handle such scenario.

src/AspNetCore/Src/applicationmanager.cxx
jkotalik: Reference a local copy of the application info here to avoid this race.

src/AspNetCore/Src/applicationinfo.cpp
jkotalik: For out of process, a new application info will be created, so it is fine if shutdown takes a while. For inproc, there is nothing we can do as we need to recycle the worker process, so we will have to reject requests until a new worker process is created.

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -194,23 +209,55 @@ APPLICATION_MANAGER::RecycleApplication(
...
+    DWORD                   dwPreviousCounter = 0;
+    APPLICATION_INFO*       pApplicationInfo;
+    BOOL                    fLocked = FALSE;
     hr = key.Initialize(pszApplicationId);
     if (FAILED(hr))
     {
         goto Finished;
     }
+
     AcquireSRWLockExclusive(&m_srwLock);
```
jkotalik: @pan-wang please review this section based on our offline discussion.

jkotalik:
Closing as I am merging with @pan-wang PR #610

PR #577, opened 2018-02-02
React to hostfxr api changes

jkotalik:
To allow for the new aspnetcore request handler to be obtained from the dotnet core application, I asked core-setup to add a new api to obtain the native search directories: https://github.com/dotnet/core-setup/blob/73e7fa93e5679eb2639eedb49761b3acf58b4843/src/corehost/cli/fxr/hostfxr.cpp#L298. This change has now has been propagated upstream. However, I realized the API changed for the method slightly, which causes the call to hostfxr_get_native_search_directories to AV. As we always attempt to call hostfxr_get_native_search_directories, the app will always AV on startup.

Must have for 2.1-preview1.

cc/ @Eilon @DamianEdwards


jkotalik:
@DamianEdwards approved via email.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -404,16 +405,17 @@ APPLICATION_INFO::FindNativeAssemblyFromHostfxr(
...
+            dwBufferSize,
+            &dwRequiredBufferSize
         );
 
         if (intHostFxrExitCode == 0)
         {
             break;
         }
-        else if (intHostFxrExitCode == API_BUFFER_TOO_SMALL)
+        else if (dwRequiredBufferSize >= dwBufferSize)
```
pan-wang: why >=? is the null terminator is not included?

jkotalik: fixed.

pan-wang: if the buffer copy is just char copy not a string, you need to add the null terminator otherwise you may run into buffer overrunning

jkotalik: It is null terminated.

PR #582, opened 2018-02-13
Move reading and writing of Pipelines to a single loop; Cancel Read every time we write.

jkotalik:
This will fix a potential issue of deadlocking on reads and writes if someone is reading and writing concurrently. I moved all websocket code to its own file too. 
Issue to add tests to websockets: https://github.com/aspnet/Home/issues/2865.
cc/ @halter73


src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -260,6 +237,14 @@ private IISAwaitable DoFlushAsync()
 
         public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationToken = default(CancellationToken))
         {
+            // To avoid concurrent reading and writing, if we have a pending read,
+            // we must cancel it. TODO check if we need to lock here. I don't think we do because cancel_io will
+            // noop if there is no data to write, and OnAsyncCompletion should only be triggered once per read/write.
+            if (!_wasUpgraded && !_operation.IsCompleted)
+            {
+                var hr = NativeMethods.http_cancel_io(_pInProcessHandler);
```
Tratcher: This seems very unstable. How do your read operations distinguish between a cancellation and real IO exception? This would need a lot of stress testing.

While we understand the benefits of pipelines for writes, are we clear on what value they provide for reads? I'm not aware of any.

halter73: Setting aside the potential issue of distinguishing canceled reads and IO errors, I think there might be a race condition here. What happens if http_cancel_io is called right before the call to http_read_request_bytes? Does the http_read_request_bytes call still get canceled?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -423,145 +408,154 @@ public void Abort()
...
+                        break;
                     }
 
-                    var result = await Input.Writer.FlushAsync();
-
-                    if (result.IsCompleted || result.IsCancelled)
+                    if (_upgradeTcs != null)
                     {
-                        break;
+                        upgrade = true;
```
halter73: Out of curiosity, what causes this loop to complete after an upgrade so that it can get to the Read/WriteWebSockets calls?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -423,145 +408,154 @@ public void Abort()
...
                     }
                 }
 
-                Input.Writer.Complete();
+                // If we upgraded the stream, we can switch to simultaneously reading and writing
+                // create two new tasks and await them finishing.
+                // We don't want to Complete the Reader and Writer as they will be reused in
+                // the other tasks.
+                if (upgrade && _upgradeTcs.TrySetResult(null))
```
halter73: Nit: I don't see why you need the upgrade bool. Couldn't you just null check _upgradeTcs here like you do above?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -423,145 +408,154 @@ public void Abort()
...
+                {
+                    _readWebsocketTask = ReadWebSockets();
+                    _writeWebsocketTask = WriteWebSockets();
```
halter73: Nit: you can get rid of the _read/writeWebsocketTask fields. Just make these local variable then do `await Task.WhenAll(readWebsocketTask, writeWebsocketTask)`.

This way you can always complete the reader and writer in this method and not in the Read/WriteWebSockets methods.

This should also simplify the finally in IISHttpContextOfT.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -423,145 +408,154 @@ public void Abort()
...
+
+                Input.Writer.Advance(read);
+            }
+            catch (COMException comException)
+            {
+                // If the exception was an aborted read operation, ignore it
+                // as we were the one to invoke it.
+                // E_OPERATIONABORTED == 0x800703e3 == -2147023901
+                // TODO check if this error code can be reported through OnAsyncCompletion other than by calling CancelIo.
+                if (comException.ErrorCode != -2147023901)
```
halter73: Could we check the error code before calling `_operation.Complete(0x800703e3, 0)`? I don't like potentially throwing an exception for every write.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: Me forgetting a break. 😨

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: Yeah I was thinking about that race last night and how to fix it. Http_read_request_bytes would not be canceled in this case. We can try adding a timeout on calling ReadToInputPipe(). Other than that, not sure what else to do besides synchronize.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: We can check inside of Complete(). Currently GetResult is the one to throw the exception.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -423,274 +432,241 @@ public void Abort()
...
             try
             {
                 while (true)
                 {
-                    // These buffers are pinned
-                    var wb = Input.Writer.GetMemory(MinAllocBufferSize);
-                    _inputHandle = wb.Retain(true);
-
-                    try
+                    if (!doneReading && _allowedToRead)
```
jkotalik: @halter73 I think having a single loop is fundamentally flawed. If I want to read the entire request body before writing, how can the reads unblock the writes with a single loop? Calling await WriteFromOutputPipe() will block until data arrives, which may not happen until the response body is read.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -29,12 +29,16 @@ internal abstract partial class IISHttpContext : NativeRequestContext, IDisposab
...
 
         protected readonly IntPtr _pInProcessHandler;
-
-        private bool _wasUpgraded;
+        protected bool _wasUpgraded;
         private int _statusCode;
         private string _reasonPhrase;
         private readonly object _onStartingSync = new object();
         private readonly object _onCompletedSync = new object();
+        protected readonly object _readCancelOperation = new object();
```
halter73: nit: _readCancelSync

jkotalik:
@halter73 I went ahead and added you as a reviewer.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -271,11 +281,12 @@ public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationTo
 
         public async Task WriteAsyncAwaited(ArraySegment<byte> data, CancellationToken cancellationToken)
         {
-            await InitializeResponseAwaited(data.Count);
-
             // WriteAsyncAwaited is only called for the first write to the body.
             // Ensure headers are flushed if Write(Chunked)Async isn't called.
             await Output.WriteAsync(data, cancellationToken: cancellationToken);
+
+            // We need to write to the output first before initializing the response.
```
davidfowl: Why is this?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISAwaitable.cs
```diff
@@ -86,7 +86,15 @@ public void UnsafeOnCompleted(Action continuation)
 
         public void Complete(int hr, int cbBytes)
         {
-            _exception = Marshal.GetExceptionForHR(hr);
+            // If the exception was an aborted read operation, ignore it
+            // as we were the one to invoke it.
+            // E_OPERATIONABORTED == 0x800703e3 == -2147023901
+            // TODO check if this error code can be reported through OnAsyncCompletion other than by calling CancelIo.
+            if (hr != -2147023901)
```
halter73: Should `_exception` still be set here if it's a write operation that was aborted?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISAwaitable.cs
```diff
@@ -86,7 +86,15 @@ public void UnsafeOnCompleted(Action continuation)
 
         public void Complete(int hr, int cbBytes)
         {
-            _exception = Marshal.GetExceptionForHR(hr);
+            // If the exception was an aborted read operation, ignore it
+            // as we were the one to invoke it.
+            // E_OPERATIONABORTED == 0x800703e3 == -2147023901
+            // TODO check if this error code can be reported through OnAsyncCompletion other than by calling CancelIo.
```
halter73: Can it?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -260,6 +248,28 @@ private IISAwaitable DoFlushAsync()
...
+                    switch (_currentOperationType)
+                    {
+                        case CurrentOperationType.Read:
+                            NativeMethods.http_cancel_io(_pInProcessHandler);
+                            break;
+                        case CurrentOperationType.Write:
+                        case CurrentOperationType.Flush:
+                        case CurrentOperationType.None:
+                        default:
+                            _allowedToRead = false;
```
halter73: Isn't it important that the data gets written to the PipeWriter before releasing the lock?

Because I think this is important, I also think the lock shouldn't be acquired until after InitializeResponseAwaited is called if it is called. If you do this, you probably won't need to reorder InitializeResponseAwaited and Output.WriteAsync below.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -260,6 +248,28 @@ private IISAwaitable DoFlushAsync()
...
+                            break;
+                    }
+                }
+            }
+
             if (!HasResponseStarted)
             {
                 return WriteAsyncAwaited(data, cancellationToken);
```
halter73: Off topic: Why call an "Awaited" method here already if you don't know yet whether there are any true-async onstarting callbacks registered? I'm guessing there usually aren't such callbacks making the extra async state machine an unnecessary cost. You can look at Kestrel's FireOnStartingMayAwait for how to do this more efficiently.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -423,274 +434,225 @@ public void Abort()
...
             try
             {
                 while (true)
                 {
-                    // These buffers are pinned
-                    var wb = Input.Writer.GetMemory(MinAllocBufferSize);
-                    _inputHandle = wb.Retain(true);
-
-                    try
+                    if (!_doneReading && _allowedToRead)
```
halter73: Checking this without a lock feels like a race. What if _allowedToRead is set to false between this line and the call to NativeReadAsync?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -423,274 +434,225 @@ public void Abort()
...
+                        _inputHandle = wb.Retain(true);
+
+                        try
                         {
-                            read = await ReadWebSocketsAsync(wb.Length);
+                            var read = await NativeReadAsync(wb.Length);
+
+                            if (read == 0)
+                            {
+                                _doneReading = true;
```
halter73: Couldn't `_doneReading` be made a local variable?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: yeah I think we need to lock here.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISAwaitable.cs
jkotalik: Gotta take a deep dive into the source code. See you in a few weeks.

jkotalik: From what Pan I can tell, this error will only be set when CancelIo occurs. So this should be fine.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: A bit confused by this comment.

> Isn't it important that the data gets written to the PipeWriter before releasing the lock?

Are you suggesting I do something like:
```c#
public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationToken = default(CancellationToken))
{
    lock (_readCancelSync)
    {
        // do cancellations
        ....
        return Operation.WriteAsync();
    }
}

```

For the WriteAsyncAwaited case, do we do something like:

```c#
public async Task WriteAsyncAwaited(ArraySegment<byte> data, CancellationToken cancellationToken)
{
    // WriteAsyncAwaited is only called for the first write to the body.
    // Ensure headers are flushed if Write(Chunked)Async isn't called.
    await InitializeResponseAwaited(data.Count);
    if (!_wasUpgraded)
    {
        lock (_readCancelSync)
        {
            // Do cancellations
        }
    }
    await Output.WriteAsync(data, cancellationToken: cancellationToken);
}

```

halter73: Close. In the second one I would still call WriteAsync inside the lock, but store the task and await it outside the lock.

jkotalik:
>We should run a stress scenario to help verify this doesn't race.

>It could be as simple as middleware reads the entire request body and then writes the contents of the request body to the response body twice. It should flush before writing the request body to the response body the second time. This way, if the race occurs, we should be able to notice when it doesn't receive the twice-duplicated response.

Wouldn't it need to be in two separate threads?

halter73:
> Wouldn't it need to be in two separate threads?

I guess it depends on the pipe scheduler. Does the loop reading the output pipe run on the same thread as the caller to Response.Body.Write()? If the consumer is on another thread, I think calling Response.Body.Write() twice on the same thread should still potentially race if a race condition exists.

A way to verify the stress test works is to make verify the old version of the PR without proper synchronization fails the stress test.

jkotalik:
🆙  📅

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Websockets.cs
```diff
@@ -0,0 +1,215 @@
...
+            _readWebSocketsOperation.Complete(hr, cbBytes);
+        }
+
+        private async Task ReadWebSockets()
+        {
+            try
+            {
+                while (true)
+                {
+                    var wb = Input.Writer.GetMemory(MinAllocBufferSize);
```
pakrym: `MinAllocBufferSize` can be set in pipe options.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -142,7 +133,7 @@ internal unsafe IISHttpContext(MemoryPool memoryPool, IntPtr pInProcessHandler,
             ResponseBody = new IISHttpResponseBody(this);
 
             Input = new Pipe(new PipeOptions(_memoryPool, readerScheduler: PipeScheduler.ThreadPool));
-            var pipe = new Pipe(new PipeOptions(_memoryPool,  readerScheduler: PipeScheduler.ThreadPool));
+            var pipe = new Pipe(new PipeOptions(_memoryPool, readerScheduler: PipeScheduler.ThreadPool, writerScheduler: PipeScheduler.ThreadPool, pauseWriterThreshold: 65536, resumeWriterThreshold: 65536));
```
pakrym: const

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -199,18 +190,25 @@ public string ReasonPhrase
             }
         }
 
-        private IISAwaitable DoFlushAsync()
+        private unsafe IISAwaitable NativeFlushAsync()
         {
-            unsafe
+            
+            lock(_readCancelSync)
```
pakrym: nit space

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -260,22 +248,45 @@ private IISAwaitable DoFlushAsync()
 
         public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationToken = default(CancellationToken))
         {
+
```
pakrym: nit: remove

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -82,17 +81,10 @@ public IISHttpContextOfT(MemoryPool memoryPool, IHttpApplication<TContext> appli
...
-                if (_writingTask != null)
-                {
-                    await _writingTask;
-                }
-
-                // The app is finished and there should be nobody reading from the request pipe
                 Input.Reader.Complete();
-
-                if (_readingTask != null)
+                if (_readWriteTask != null)
```
pakrym: Is this racing with the lock?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -260,22 +248,45 @@ private IISAwaitable DoFlushAsync()
...
+                    return Output.WriteAsync(data, cancellationToken: cancellationToken);
+                }
+            }
+            else
+            {
+                return Output.WriteAsync(data, cancellationToken: cancellationToken);
+            }
         }
 
+
```
pakrym: nit: remove

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Websockets.cs
```diff
@@ -0,0 +1,215 @@
...
+            {
+                ReadResult result;
+
+                try
+                {
+                    result = await Output.Reader.ReadAsync();
+                }
+                catch
+                {
+                    Output.Reader.Complete();
```
pakrym: break;

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -418,279 +429,264 @@ public unsafe void SendResponseHeaders(bool appCompleted)
             }
         }
 
+        // Needs to be locked when accessing this method.
+        private void ModifyReadCancellationState()
```
pakrym: Does it need better name? `DisableReads`?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -418,279 +429,264 @@ public unsafe void SendResponseHeaders(bool appCompleted)
...
+        {
+            switch (_currentOperationType)
+            {
+                case CurrentOperationType.Read:
```
pakrym: if/else should be enough here.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -29,12 +29,16 @@ internal abstract partial class IISHttpContext : NativeRequestContext, IDisposab
...
-        private bool _wasUpgraded;
+        protected bool _wasUpgraded;
         private int _statusCode;
         private string _reasonPhrase;
         private readonly object _onStartingSync = new object();
         private readonly object _onCompletedSync = new object();
+        protected readonly object _readCancelSync = new object();
+        private readonly object _createReadWriteBodySync = new object();
+
+        private bool _allowedToRead = true;
```
pakrym: Can this logic be handled with three state? Writing/Reading/None?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -418,279 +429,264 @@ public unsafe void SendResponseHeaders(bool appCompleted)
...
+                        }
                     }
-                    finally
+
+                    ReadResult readResult;
+                    if (!doneReading)
                     {
-                        Input.Writer.Commit();
-                        _inputHandle.Dispose();
+                        if (!Output.Reader.TryRead(out readResult))
```
pakrym: Why?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -418,279 +429,264 @@ public unsafe void SendResponseHeaders(bool appCompleted)
...
+                        _readWriteTask = ProcessRequestAndResponseBody();
+                    }
+                }
             }
         }
 
-        private async Task ProcessRequestBody()
+        // IIS can only handle on async operation per request at a time.
+        // To guarantee this, we have a single loop to read and write the request/response
+        private async Task ProcessRequestAndResponseBody()
```
pakrym: This method is very long and pretty hard to follow (100 lines with 13 nested ifs). Worth refactoring/splitting into smaller methods? Or at least commenting a bit more.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Websockets.cs
jkotalik: done.

jkotalik:
@davidfowl based on offline feedback I rewrote this. Please review when you get the chance.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Websockets.cs
jkotalik: done.

src/Microsoft.AspNetCore.Server.IIS/Microsoft.AspNetCore.Server.IIS.csproj
```diff
@@ -25,6 +25,10 @@
     <ProjectReference Include="..\Microsoft.AspNetCore.Server.IISIntegration\Microsoft.AspNetCore.Server.IISIntegration.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
davidfowl: Why is this here?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Websockets.cs
```diff
@@ -0,0 +1,235 @@
...
+    /// </summary>
+    internal partial class IISHttpContext
+    {
+        private bool _wasUpgraded; // Used for detecting repeated upgrades in IISHttpContext
+
+        private IISAwaitable _readWebSocketsOperation;
+        private IISAwaitable _writeWebSocketsOperation;
+        private TaskCompletionSource<object> _upgradeTcs;
+
+        private async Task CheckForUpgrade()
```
davidfowl: CheckForUpgrade is a strange name since it's really just starting the bidirectional stream.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Websockets.cs
```diff
@@ -0,0 +1,235 @@
...
+
+                    handle = b.Retain(true);
+
+                    chunk.DataChunkType = HttpApiTypes.HTTP_DATA_CHUNK_TYPE.HttpDataChunkFromMemory;
+                    chunk.fromMemory.BufferLength = (uint)b.Length;
+                    chunk.fromMemory.pBuffer = (IntPtr)handle.Pointer;
+
+                    currentChunk++;
+                }
+                hr = NativeMethods.http_websockets_write_bytes(_pInProcessHandler, pDataChunks, nChunks, IISAwaitable.WriteCallback, (IntPtr)_thisHandle, out fCompletionExpected);
```
davidfowl: Where is this hr checked?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Websockets.cs
```diff
@@ -0,0 +1,235 @@
...
+            _readWebSocketsOperation.Complete(hr, cbBytes);
+        }
+
+        private async Task ReadWebSockets()
+        {
+            try
+            {
+                while (true)
+                {
+                    var wb = Input.Writer.GetMemory();
```
davidfowl: Rename wb to memory

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -82,18 +82,15 @@ public IISHttpContextOfT(MemoryPool memoryPool, IHttpApplication<TContext> appli
...
-                {
-                    await _writingTask;
-                }
-
                 // The app is finished and there should be nobody reading from the request pipe
                 Input.Reader.Complete();
 
-                if (_readingTask != null)
+                Task readWriteTask;
+                lock(_createReadWriteBodySync)
```
davidfowl: nit: formatting. Also lets call it processBodiesTask

src/RequestHandler/inprocess/inprocesshandler.cpp
```diff
@@ -59,25 +59,16 @@ IN_PROCESS_HANDLER::OnAsyncCompletion(
     HRESULT     hrCompletionStatus
 )
 {
-    HRESULT hr;
-    if (FAILED(hrCompletionStatus))
+    // For now we are assuming we are in our own self contained box. 
+    // TODO refactor Finished and Failure sections to handle in process and out of process failure.
```
davidfowl: File issues for these TODO? How important are they for the code to function?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -25,16 +25,23 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
     internal abstract partial class IISHttpContext : NativeRequestContext, IDisposable
     {
         private const int MinAllocBufferSize = 2048;
+        private const int PauseWriterThreshold = 65536;
+        private const int ResumeWriterTheshold = 32768;
```
davidfowl: Divide PauseWriterThreshold by 2?

src/RequestHandler/inprocess/inprocesshandler.cpp
jkotalik: I should both of these TODOs.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -423,231 +646,72 @@ public void Abort()
             // TODO
         }
 
-        public void StartReadingRequestBody()
+        private unsafe IISAwaitable WriteToIISAsync(ReadOnlyBuffer<byte> buffer)
```
davidfowl: It's unfortunate this is duplicated between the websocket logic and here.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+
+                // If the pipe was closed, we are done reading, 
+                if (flushResult.IsCompleted || flushResult.IsCancelled)
+                {
+                    _doneReading = true;
+                }
+
+                // We are done reading now, set _reading to false.
+                // _reading may already be set to false through cancellation, but it is okay to be redundant here.
+                lock(_stateSync)
```
davidfowl: nit: formatting

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+                Output.Reader.Complete();
+                return;
+            }
+
+            await WriteLoopAsync();
+        }
+
+        public async Task ReadAndWriteLoopAsync()
+        {
+            // First we check if there is anything to write from the Output pipe
```
davidfowl: I'd move this comment inside of the while loop

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+                        }
+                        else
+                        {
+                            // Flush of zero bytes
+                            await FlushToIISAsync();
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        throw ex;
```
davidfowl: Who observes these exceptions?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+                // Check if Output pipe has anything to write to IIS.
+                if (Output.Reader.TryRead(out var readResult))
+                {
+                    var buffer = readResult.Buffer;
+                    var consumed = buffer.End;
+
+                    try
+                    {
+                        // If the output pipe is termniated, cancel the request.
+                        if (readResult.IsCancelled)
```
davidfowl: Do you ever call CancelPendingRead() on the output pipe?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+
+                        if (!buffer.IsEmpty)
+                        {
+                            // Write to IIS buffers
+                            // Guaranteed to write the entire buffer to IIS
+                            await WriteToIISAsync(buffer);
+                        }
+                        else if (readResult.IsCompleted)
+                        {
+                            break;
```
davidfowl: Isn't _doneWriting true here?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+                    {
+                        throw ex;
+                    }
+                    finally
+                    {
+                        // Always Advance the data pointer to the end of the buffer.
+                        Output.Reader.AdvanceTo(buffer.End);
+                    }
+                }
+
```
davidfowl: nit: extra space

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+               
+                // Check if there was an upgrade. If there is, we will replace the request and response bodies with
+                // two seperate loops
+                await CheckForUpgrade();
```
davidfowl: Won't this loop Complete the input and output writer and reader? Won't that make the code below fail as it tries to access memory after completing?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+
+                // Now we handle the read. 
+                var wb = Input.Writer.GetMemory();
+                _inputHandle = wb.Retain(true);
+
+                try
+                {
+                    // Lock around invoking ReadFromIISAsync as we don't want to call CancelIo
+                    // when calling read
+                    var read = await ReadFromIISAsync(wb.Length);
```
davidfowl: Don't errors thrown from here leak? Does it ever make it to the caller of ReadAsync?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+                if (flushResult.IsCompleted || flushResult.IsCancelled)
+                {
+                    _doneReading = true;
+                }
+
+                // We are done reading now, set _reading to false.
+                // _reading may already be set to false through cancellation, but it is okay to be redundant here.
+                lock(_stateSync)
+                {
+                    _reading = false;
```
davidfowl: Why don't you set this from inside of ReadFromIISAsync

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+                        // Flush of zero bytes will 
+                        await FlushToIISAsync();
+                    }
+                }
+                finally
+                {
+                    // Always Advance the data pointer to the end of the buffer.
+                    Output.Reader.AdvanceTo(buffer.End);
+                }
+                await CheckForUpgrade();
```
davidfowl: Why do we need this?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+                        // Guaranteed to write the entire buffer to IIS
+                        await WriteToIISAsync(buffer);
+                    }
+                    else if (readResult.IsCompleted)
+                    {
+                        break;
+                    }
+                    else
+                    {
+                        // Flush of zero bytes will
```
davidfowl: Incomplete comments?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+            }
+
+            // Close the output pipe as we are done reading from it.
+            Output.Reader.Complete();
+        }
+
+        private unsafe IISAwaitable FlushToIISAsync()
+        {
+            // Calls flush 
+            lock(_stateSync)
```
davidfowl: What state are you protecting here?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+            {
+                var hr = 0;
+                hr = NativeMethods.http_flush_response_bytes(_pInProcessHandler, out var fCompletionExpected);
+                if (!fCompletionExpected)
+                {
+                    _operation.Complete(hr, 0);
```
davidfowl: Calling Complete within the lock might be dangerous.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -258,24 +239,266 @@ private IISAwaitable DoFlushAsync()
...
+                }
+            }
+              
+            return _operation;
+        }
+
+        private void DisableReads()
```
davidfowl: Put a comment that this is always called from within a lock.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Websockets.cs
jkotalik: Forgot the synchronous completion check.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: No. Should I?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: No one does. Should I be observing them in ConsumeAsync?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: You are correct. There is also the inverse issue in the WriteLoop where the Input will already be completed.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: Yeah they would leak. How should I propagate them to the user? The failure here will not sync up with user code.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: Oh wait, I should be calling complete with the exception.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: Correct me if I'm wrong, but we can finish reading the request body before we know whether we are going to upgrade or not. Therefore, when we are writing, we need to check.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: I would need to set it in OnAsyncCompletion instead (which is fine).

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: image

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: You are correct. I think I only need to do it on ReadFromIISAsync

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: I could try to abstract out the calls to native, however it may be over engineering it.

jkotalik:
@davidfowl 🆙 📅

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,431 @@
...
+        }
+
+        private unsafe IISAwaitable ReadFromIISAsync(int length)
+        {
+            lock (_stateSync)
+            {
+                // We don't want to read if there is data available in the output pipe
+                if (Output.Reader.TryRead(out var result) && !result.Buffer.IsEmpty)
+                {
+                    _operation.Complete(hr: IISServerConstants.HResultCancelIO, cbBytes: 0);
```
davidfowl: I would add a comment here describing that we're marking the current  operation as cancelled as a result.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,431 @@
...
+                           length,
+                           out var dwReceivedBytes,
+                           out bool fCompletionExpected);
+                    if (!fCompletionExpected)
+                    {
+                        _operation.Complete(hr, dwReceivedBytes);
+                    }
+                    else
+                    {
+                        _reading = true;
```
davidfowl: Make a comment saying if we complete synchronously then there's no need to set the reading flag to true since there's nothing to cancel.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,431 @@
...
+                    }
+                }
+
+                return _operation;
+            }
+        }
+
+        private unsafe IISAwaitable WriteToIISAsync(ReadOnlyBuffer<byte> buffer)
+        {
+            lock (_stateSync)
```
davidfowl: Why is this locked?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,431 @@
...
+                    // Check if Output pipe has anything to write to IIS.
+                    if (Output.Reader.TryRead(out var readResult))
+                    {
+                        var buffer = readResult.Buffer;
+                        var consumed = buffer.End;
+
+                        try
+                        {
+                            // If the output pipe is terminated, cancel the request.
+                            if (readResult.IsCancelled)
```
davidfowl: Who calls CancelPendingRead on the output pipe? That's the only way this will ever happen.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,431 @@
...
+                        }
+                    }
+
+                    // Check if there was an upgrade. If there is, we will replace the request and response bodies with
+                    // two seperate loops. These will still be using the same Input and Output pipes here.
+                    if (_upgradeTcs?.TrySetResult(null) == true)
+                    {
+                        await StartBidirectionalStream();
+
+                        // Input and Output will be closed in StartBidirectionalStream.
```
davidfowl: 👍

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,431 @@
...
+                    }
+                    finally
+                    {
+                        // Always commit any changes to the Input pipe
+                        Input.Writer.Commit();
+                        _inputHandle.Dispose();
+                    }
+
+                    // Flush the read data for the Input Pipe writer
+                    var flushResult = await Input.Writer.FlushAsync();
```
davidfowl: You're going to wake the reader up even if you read nothing or you cancelled. Is that intentional?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,431 @@
...
+            hr = NativeMethods.http_flush_response_bytes(_pInProcessHandler, out var fCompletionExpected);
+            if (!fCompletionExpected)
+            {
+                _operation.Complete(hr, 0);
+            }
+
+            return _operation;
+        }
+
+        public async Task ReadAndWriteLoopAsync()
```
davidfowl: One thing that would help grok this would be a summary of the functionality in this method and the various ways each of the state (_doneReading, and _doneWriting) ends up being set.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,431 @@
...
+                        _doneReading = true;
+                    }
+                }
+
+                // Complete the input writer as we are done reading the request body.
+                Input.Writer.Complete();
+            }
+            catch (Exception ex)
+            {
+                Output.Reader.Complete(ex);
```
davidfowl: The reader exception doesn't matter as much. The writer exception is about triggering errors in ReadAsync. This will flow to user code so we should make sure that we understand what those can look like.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,431 @@
...
+                        Output.Reader.AdvanceTo(buffer.End);
+                    }
+
+                    // Check if there was an upgrade. If there is, we will replace the request and response bodies with
+                    // two seperate loops
+                    if (_upgradeTcs?.TrySetResult(null) == true)
+                    {
+                        // The Input pipe has already been completed at this point
+                        // Create a new Input pipe for websockets
+                        Input = new Pipe(new PipeOptions(_memoryPool, readerScheduler: PipeScheduler.ThreadPool, minimumSegmentSize: MinAllocBufferSize));
```
davidfowl: Ugh, this is pretty gross. When would this ever happen? How does the input pipe close for an upgraded request? Is this an artifact of how the code is written or can this actually happen?

test/IISTestSite/Startup.cs
```diff
@@ -325,5 +331,81 @@ private void CheckRemoveAuthEnvironmentVariable(IApplicationBuilder app)
...
+        {
+            app.Run(async context =>
+            {
+                var t2 = Task.Run(() => WriteManyTimesToResponseBody(context));
+                var t1 = Task.Run(() => ReadRequestBody(context));
+                await Task.WhenAll(t1, t2);
+            });
+        }
+
+        private async Task ReadRequestBody(HttpContext context)
```
davidfowl: Add a test that does CopyToAsync from the request to the response body.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -82,18 +82,15 @@ public IISHttpContextOfT(MemoryPool memoryPool, IHttpApplication<TContext> appli
...
-                {
-                    await _writingTask;
-                }
-
                 // The app is finished and there should be nobody reading from the request pipe
                 Input.Reader.Complete();
 
-                if (_readingTask != null)
+                Task processBodiesTask;
+                lock(_createReadWriteBodySync)
```
davidfowl: nit: formatting

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -199,83 +201,18 @@ public string ReasonPhrase
...
-        public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationToken = default(CancellationToken))
+        // Always called from witin a lock
+        private void DisableReads()
         {
-            if (!HasResponseStarted)
+            // To avoid concurrent reading and writing, if we have a pending read,
+            // we must cancel it.
+            // _reading will always be false if we upgrade to websockets, so we don't need to check wasUpgrade
+            // Also, we set _reading to false after cancelling to detect redundant calls
+            if (!_doneReading && _reading)
```
davidfowl: Do we ever need to the _doneReading check?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.Websockets.cs
```diff
@@ -0,0 +1,236 @@
...
+        private IISAwaitable _writeWebSocketsOperation;
+        private TaskCompletionSource<object> _upgradeTcs;
+
+        private async Task StartBidirectionalStream()
+        {
+            // IIS allows for websocket support and duplex channels only on Win8 and above
+            // This allows us to have two tasks for reading the request and writing the response
+            var readWebsocketTask = ReadWebSockets();
+            var writeWebsocketTask = WriteWebSockets();
+            await Task.WhenAll(readWebsocketTask, writeWebsocketTask);
```
davidfowl: You don't need to await here, you can just return the task directly and remove the async from the method definition.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
jkotalik: It wasn't. I added a continue before.

test/IISTestSite/Startup.cs
jkotalik: Done.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.FeatureCollection.cs
```diff
@@ -291,15 +291,18 @@ void IHttpResponseFeature.OnCompleted(Func<object, Task> callback, object state)
...
             ReasonPhrase = ReasonPhrases.GetReasonPhrase(StatusCodes.Status101SwitchingProtocols);
-            await UpgradeAsync();
-            NativeMethods.http_enable_websockets(_pInProcessHandler);
-
-            _wasUpgraded = true;
             _readWebSocketsOperation = new IISAwaitable();
             _writeWebSocketsOperation = new IISAwaitable();
+            NativeMethods.http_enable_websockets(_pInProcessHandler);
+
+            //
```
davidfowl: forgot to comment?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,476 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.HttpSys.Internal;
+using System;
```
davidfowl: fix your VS to sort System usings first

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,476 @@
...
+        /// Writes data to the output pipe.
+        /// </summary>
+        /// <param name="data"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (!_hasResponseStarted)
+            {
+                return WriteAsyncAwaited(data, cancellationToken);
```
davidfowl: Was it always this way? we take the slow path if the response hasn't started?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,469 @@
...
+
+                    // If the pipe was closed, we are done reading, 
+                    if (flushResult.IsCompleted || flushResult.IsCancelled)
+                    {
+                        _doneReading = true;
+                    }
+                }
+
+                // Complete the input writer as we are done reading the request body.
+                Input.Writer.Complete();
```
davidfowl: Instead can we put this in a finally?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,469 @@
...
+                        _inputHandle.Dispose();
+                    }
+
+                    // Flush the read data for the Input Pipe writer
+                    var flushResult = await Input.Writer.FlushAsync();
+
+                    // If the pipe was closed, we are done reading, 
+                    if (flushResult.IsCompleted || flushResult.IsCancelled)
+                    {
+                        _doneReading = true;
```
davidfowl: Do we need _doneReading at all or can we just break here?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,469 @@
...
+                        // Lock around invoking ReadFromIISAsync as we don't want to call CancelIo
+                        // when calling read
+                        var read = await ReadFromIISAsync(memory.Length);
+
+                        // read value of 0 == done reading
+                        // read value of -1 == read cancelled, still allowed to read but we
+                        // need a write to occur first. 
+                        if (read == 0)
+                        {
+                            _doneReading = true;
```
davidfowl: Do we need _doneReading at all or can we just break here?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,469 @@
...
+        // There are two modes for reading and writing to the request/response bodies.
+        // 1. Await all reads and try to read from the Output pipe
+        // 2. Done reading and await all writes.
+        private async Task ConsumeAsync()
+        {
+            await ReadAndWriteLoopAsync();
+
+            // If we are done writing, complete the output pipe and return
+            // Input Pipe will be closed when ReadAndWriteLoopAsync returns
+            await WriteLoopAsync();
```
davidfowl: Can we just move the call to WriteLoopAsync at the end of ReadAndWriteLoopAsync

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
davidfowl: Can we just break here? Do we need _doneReading at all?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -58,6 +58,7 @@ public IISHttpContextOfT(MemoryPool memoryPool, IHttpApplication<TContext> appli
             if (Volatile.Read(ref _requestAborted) == 0)
             {
                 await ProduceEnd();
+
```
davidfowl: nit: remove new line

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,476 @@
...
+        /// <returns></returns>
+        public Task FlushAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            // Issue: On first flush for websockets, we need to flush the headers such that
+            // IIS will know that an upgrade occured.
+            // If we don't have anything on the Output pipe, the TryRead in ReadAndWriteLoopAsync
+            // will fail and we will signal the upgradeTcs that we are upgrading. However, we still
+            // didn't flush.
+            // There are a few ways to fix it:
+            // 1. Swap the order to Output.FlushAsync and InitializeResponse (which will start the read/write loop)
```
davidfowl: Say which one we're doing.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,476 @@
...
+                    var readResult = await Output.Reader.ReadAsync();
+
+                    // Get data from pipe
+                    var buffer = readResult.Buffer;
+                    var consumed = buffer.End;
+
+                    try
+                    {
+                        // If the output pipe is termniated, cancel the request.
+                        if (readResult.IsCancelled)
```
davidfowl: Remove this, nobody will trigger this.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -0,0 +1,476 @@
...
+            {
+                while (!_doneReading)
+                {
+                    // First we check if there is anything to write from the Output pipe
+                    // If there is, we call WriteToIISAsync
+                    // Check if Output pipe has anything to write to IIS.
+                    if (Output.Reader.TryRead(out var readResult))
+                    {
+                        var buffer = readResult.Buffer;
+                        var consumed = buffer.End;
```
davidfowl: Nobody uses consumed.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -25,16 +25,23 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
     internal abstract partial class IISHttpContext : NativeRequestContext, IDisposable
     {
         private const int MinAllocBufferSize = 2048;
+        private const int PauseWriterThreshold = 20;
```
davidfowl: I think you forgot this?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -25,16 +25,23 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
...
+        private const int ResumeWriterTheshold = 20;
 
         private static bool UpgradeAvailable = (Environment.OSVersion.Version >= new Version(6, 2));
 
         protected readonly IntPtr _pInProcessHandler;
 
-        private bool _wasUpgraded;
+        private bool _reading; // To know whether we are currently in a read operation.
+        private bool _doneReading;
```
davidfowl: I think you can get rid of this.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -367,10 +276,24 @@ private void SetErrorResponseHeaders(int statusCode)
...
+            if (_hasResponseStarted)
+            {
+                return;
+            }
 
-            // Force flush
-            await Output.FlushAsync();
+            _hasResponseStarted = true;
+
+            SendResponseHeaders(true);
```
davidfowl: `appCompleted: true`

davidfowl: Add a comment explaining what this is doing.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -367,10 +276,24 @@ private void SetErrorResponseHeaders(int statusCode)
...
+            StartProcessingRequestAndResponseBody();
```
davidfowl: I thought you had to flush before starting this?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -416,288 +339,21 @@ public unsafe void SendResponseHeaders(bool appCompleted)
                     }
                 }
             }
+
```
davidfowl: nit: remove whitespace

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -810,26 +466,16 @@ public void IndicateCompletion(NativeMethods.REQUEST_NOTIFICATION_STATUS notific
...
 
-        internal void CompleteWriteWebSockets(int hr, int cbBytes)
-        {
-            _writeWebSocketsOperation.Complete(hr, cbBytes);
-        }
-
-        internal void CompleteReadWebSockets(int hr, int cbBytes)
-        {
-            _readWebSocketsOperation.Complete(hr, cbBytes);
+            continuation.Invoke();
```
davidfowl: Can this ever be null?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
jkotalik: Yeah it was always this way. Suggestions to fix in the future?

jkotalik:
@davidfowl 🆙 📅

Issue #536, opened 2018-01-22
ANCM: Inprocess: environmentVariables does not work 


jhkimnew:
1. Configure environmentVariables  for inprocess mode as the following example web.config:

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <system.webServer>
    <handlers>
      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
    </handlers>
    <aspNetCore processPath="c:\program files\dotnet\dotnet.exe" arguments=".\Test123.dll" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" hostingModel="inprocess">
            <environmentVariables>
                <environmentVariable name="MyEnv" value="foo" />
            </environmentVariables>
        </aspNetCore>
  </system.webServer>
</configuration>
```

2. Send a request to the aspnetcore app

EXPECTED:

The configured environment variables are available in the application process

ACTUAL:

The configured environment variables are not available in the application process

jkotalik:
This should be a quick fix, @pan-wang we decided on appending environment variables rather than overwriting them, correct?

Tratcher:
Yes please

jkotalik:
@muratg let's put this in preview 2. There are work arounds by setting the environment variables directly on the w3wp process. The question is if we are okay with the web.config environment variable settings affecting the rest of the IIS pipeline's environment variables.

Tratcher:
Vs. having consistent behavior for in and out of proc.

muratg:
@shirhatti FYI

jhkimnew:
BTW, the work-around is only available recent OS build. For example, on Windows 2012 R2, users can add additional environment variable to IIS worker process.

shirhatti:
@jkotalik / @pan-wang Has it been verified that appending the environment variables will work? @BillHiebert is relying on this feature to work for tooling

muratg:
@BillHiebert @jkotalik will be working on this soon.

jkotalik:
@pan-wang @Tratcher we need to confirm if any environment variables need to be modified from their original values. The out of process code does manipulation on the auth environment variables, but features like hostingstartup may need to be modified.

Tratcher:
List the variables and values you're concerned about.

jkotalik:
Here are the ones I can find.
```
ASPNETCORE_IIS_AUTH_ENV_STR
[...]

PR #583, opened 2018-02-13
Adds environment variable for inprocess

jkotalik:
For #536. This ended up being a little more difficult because:
- ENV_VAR_HASH doesn't have an iterator (have to call apply with a function pointer).
- SetEnvironmentVariable doesn't append by default, so we need to call GetEnvironmentVariable.

We can do a bit of refactoring between the inprocess and out of process environment variable logic (some can be refactored into a helper class).


test/IISTestSite/web.config
```diff
@@ -4,6 +4,10 @@
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
     </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" hostingModel="inprocess"/>
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" hostingModel="inprocess">
+      <environmentVariables>
+        <environmentVariable name="ASPNETCORE_INPROCESS_TESTING_VALUE" value="foobar" />
```
Tratcher: how about a test for adding to an existing variable?

src/CommonLib/environmentvariablehash.h
```diff
@@ -149,6 +149,95 @@ class ENVIRONMENT_VAR_HASH : public HASH_TABLE<ENVIRONMENT_VAR_ENTRY, PWSTR>
...
+        if (dwResult != -1)
+        {
+            struNameBuffer.QueryStr()[dwResult] = L'\0';
+            if (FAILED(hr = struNameBuffer.SyncWithBuffer()))
+            {
+                goto Finished;
+            }
+        }
+
+        if (FAILED(hr = struValueBuffer.Resize(300)))
```
pan-wang: you can use STACK_STRU(struValueBuffer, 300) to avoid this Resize call

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -507,6 +507,45 @@ IN_PROCESS_APPLICATION::ExecuteAspNetCoreProcess(
...
+    //
+    // few environment variables expected, small bucket size for hash table
+    //
+    if (FAILED(hr = pEnvironmentVarTable->Initialize(37 /*prime*/)))
+    {
+        goto Finished;
+    }
+
+    m_pConfig->QueryEnvironmentVariables()->Apply(ENVIRONMENT_VAR_HASH::CopyToTable, pEnvironmentVarTable);
+    pEnvironmentVarTable->FindKey(ASPNETCORE_IIS_AUTH_ENV_STR, &pIISAuthEntry);
```
pan-wang: how about others such as ASPNETCORE_PORT, ASPNETCORE_APPL_PATH? Does customer code use them?

src/CommonLib/environmentvariablehash.h
jkotalik: Are you allowed to resize a STACK_STRU after initialization?

test/IISTestSite/web.config
jkotalik: Done, and a few other tests.

src/RequestHandler/inprocess/inprocessapplication.cpp
jkotalik: I modified the logic to now include the ASPNETCORE_AUTH_ENV string.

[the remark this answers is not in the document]
jkotalik: I'll confirm, but I think the code path is now the same, so the hosting_assembly should be added too.

PR #585, opened 2018-02-13
Update Microsoft.AspNetCore.AspNetCoreModule to use the correct license URL and copyright

natemcmaster:
Part of https://github.com/aspnet/Universe/issues/873


shirhatti:
Uhh. I'm not sure about this

@ChrisSfanos @eilon Is this okay to change?

natemcmaster:
For reference, this matches the rest of the aspnetcore packages. https://github.com/aspnet/BuildTools/blob/dev/src/Internal.AspNetCore.Sdk/build/Common.props#L13-L23

nuget/AspNetCore.nuspec
```diff
@@ -6,10 +6,10 @@
     <version>$VERSION$</version>
     <authors>Microsoft</authors>
     <owners>Microsoft</owners>
-    <licenseUrl>http://www.microsoft.com/web/webpi/eula/net_library_eula_ENU.htm</licenseUrl>
```
natemcmaster: It's not clear to me whether apache or MIT is correct, but either way, this URL is wrong.

jkotalik: ANCM and IISIntegration used different licenses (ANCM == MIT, IISIntegration == Apache). I added both to the LICENSE.txt file after discussing with @Eilon . We may want a better solution than this.

Eilon: This package is deliberately different from other packages in ASP.NET Core, so we shouldn't change what's already there (and vetted). However, we *must* change the URLs to all be HTTPS.

natemcmaster: ok, reverted to original URL (plus https)

natemcmaster:
:up: :date:

PR #600, opened 2018-02-21
grant the config setting on 502.5 page setting

pan-wang:
During refactoring, we missed the part of checking configuration setting on flowing 502.5 page.


src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -330,7 +330,7 @@ FORWARDING_HANDLER::OnExecuteRequestHandler()
     {
         pResponse->SetStatus(400, "Bad Request", 0, hr);
     }
-    else
+    else if (fFailedToStartKestrel && !m_pApplication->QueryConfig()->QueryDisableStartUpErrorPage())
```
jkotalik: why do we need to check if we failed to start kestrel here?

pan-wang: this is a flag indicating we failed to get process (for out-of-process case)

PR #601, opened 2018-02-21
Increment version numbers on aspnetcore.dll and aspnetcorerh.dll, remove unnecessary files

jkotalik:
- The aspnetcore.dll was originally version 7.1.xxxx.0, so I needed to increment it to something larger such that the MSI will replace it. I chose 8.1.0.xxxx to somewhat match the current dotnet version. 
- Currently aspnetcorerh.dll matches the aspnetcore.dll version, but I think we should make the aspnetcorerh.dll match the dotnet version. I can use the VersionPrefix defined in version.props.
- I wanted to get the AssemblyRevision from the internal SDK (https://github.com/aspnet/BuildTools/blob/5e70208f6d4862de72a3dab2aa74e82cb3b0481f/src/Internal.AspNetCore.Sdk/build/Common.props#L38) but I don't believe it is available when building vcxprojs. @pranavkm I didn't see a great way of adding AssemblyInfo for C++ projects (ex https://stackoverflow.com/questions/1126880/how-can-i-auto-increment-the-c-sharp-assembly-version-via-our-ci-platform-hudso is only for .net)


build/Build.Settings
```diff
@@ -84,4 +84,10 @@
      </Link>
    </ItemDefinitionGroup>
  
+  <PropertyGroup>
+    <_TwoDigitYear>$([MSBuild]::Subtract($([System.DateTime]::UtcNow.Year), 2000))</_TwoDigitYear>
+    <_ThreeDigitDayOfYear>$([System.DateTime]::UtcNow.DayOfYear.ToString().PadLeft(3, '0'))</_ThreeDigitDayOfYear>
+    <AssemblyRevision>$(_TwoDigitYear)$(_ThreeDigitDayOfYear)</AssemblyRevision>
```
natemcmaster: So, when we produce more than one build in the same day, won't this lead to duplicate file revisions?

src/RequestHandler/RequestHandler.vcxproj
```diff
@@ -231,13 +231,31 @@
     <ClCompile Include="outofprocess\websockethandler.cxx" />
     <ClCompile Include="outofprocess\winhttphelper.cxx" />
   </ItemGroup>
+  <Target Name="CreateVersionHeader" BeforeTargets="PrepareForBuild">
```
natemcmaster: Does this break incremental compilation? I'm not sure if this is a thing in vcsproj.

build/Build.Settings
jkotalik: Yeah it does. Were we okay with doing this for other dlls because we use nuget package versions?

src/RequestHandler/RequestHandler.vcxproj
jkotalik: This is what used to be in the ANCM repo: https://github.com/aspnet/AspNetCoreModule/blob/e714153782a9fde9a529804efb0bcbf015af20b1/src/AspNetCore/AspNetCore.vcxproj#L193. So it is probably available.

src/RequestHandler/RequestHandler.vcxproj
```diff
@@ -231,13 +231,31 @@
...
+      <VersionHeaderContents Include="%0a" />
+      <VersionHeaderContents Include="// This file is auto-generated" />
+      <VersionHeaderContents Include="%0a" />
+      <VersionHeaderContents Include="#define FileVersion $(AspNetCoreModuleVersionMajor),$(AspNetCoreModuleVersionMinor),$(AspNetCoreModuleVersionPatch),$(AssemblyRevision)" />
+      <VersionHeaderContents Include="#define FileVersionStr &quot;$(AspNetCoreModuleVersionMajor).$(AspNetCoreModuleVersionMinor).$(AspNetCoreModuleVersionPatch).$(AssemblyRevision)\0&quot;" />
+      <VersionHeaderContents Include="#define ProductVersion $(AspNetCoreModuleVersionMajor),$(AspNetCoreModuleVersionMinor),$(AspNetCoreModuleVersionPatch),$(AssemblyRevision)" />
+      <VersionHeaderContents Include="#define ProductVersionStr &quot;$(AspNetCoreModuleVersionMajor).$(AspNetCoreModuleVersionMinor).$(AspNetCoreModuleVersionPatch).$(AssemblyRevision)\0&quot;" />
+      <VersionHeaderContents Include="#define PlatformToolset &quot;$(PlatformToolset)\0&quot;" />
+    </ItemGroup>
+    <WriteLinesToFile File="version.h" Lines="@(VersionHeaderContents)" OverWrite="true" />
```
pranavkm: Consider adding `WriteOnlyWhenDifferent="True"` to this. Should avoid writing it and breaking increments when the file content hasn't changed. In addition, add the file to writes

```xml
<ItemGroup>
  <FileWrites Include="version.h" />
</ItemGroup>
```

Prevents msbuild from cleaning up the file.

src/RequestHandler/RequestHandler.vcxproj
```diff
@@ -231,13 +231,31 @@
...
+  </Target>
   <ItemGroup>
     <ProjectReference Include="..\CommonLib\CommonLib.vcxproj">
       <Project>{55494e58-e061-4c4c-a0a8-837008e72f85}</Project>
     </ProjectReference>
     <ProjectReference Include="..\IISLib\IISLib.vcxproj">
       <Project>{4787a64f-9a3e-4867-a55a-70cb4b2b2ffe}</Project>
     </ProjectReference>
+  </ItemGroup>
+    <ItemGroup>
```
pranavkm: Formatting

src/RequestHandler/requesthandler.rc
```diff
@@ -0,0 +1,117 @@
...
+BEGIN
+    BLOCK "StringFileInfo"
+    BEGIN
+        BLOCK "040904b0"
+        BEGIN
+            VALUE "CompanyName", "Microsoft"
+            VALUE "FileDescription", "IIS ASP.NET Core Module"
+            VALUE "FileVersion", FileVersionStr
+            VALUE "InternalName", "aspnetcorer.dll"
+            VALUE "LegalCopyright", "Copyright (C) 2016"
```
pranavkm: Has @Eilon verified these?

build/Build.Settings
pranavkm: AFAIK, we use this pattern elsewhere

src/RequestHandler/requesthandler.rc
```diff
@@ -0,0 +1,117 @@
...
+ FILESUBTYPE 0x0L
+BEGIN
+    BLOCK "StringFileInfo"
+    BEGIN
+        BLOCK "040904b0"
+        BEGIN
+            VALUE "CompanyName", "Microsoft"
+            VALUE "FileDescription", "IIS ASP.NET Core Module"
+            VALUE "FileVersion", FileVersionStr
+            VALUE "InternalName", "aspnetcorer.dll"
```
natemcmaster: Typo: aspnetcorerh.dll

src/RequestHandler/requesthandler.rc
jkotalik: @Eilon these are the same as the AspNetCoreModule values except I added "Request Handler" and changed the copyright to 2018.

jkotalik: cc/ @shirhatti @muratg for this.

build/Build.Settings
natemcmaster: Where do we use this? It means our builds are not reproducible, and that two builds with different sources will have the same revision number. Although it may not be a huge problem, it's certainly not a good thing.

src/RequestHandler/requesthandler.rc
Eilon: Yeah this doesn't look right to me. I think it should be `Copyright (c) .NET Foundation`.

src/RequestHandler/requesthandler.rc
```diff
@@ -0,0 +1,117 @@
...
+BEGIN
+    BLOCK "StringFileInfo"
+    BEGIN
+        BLOCK "040904b0"
+        BEGIN
+            VALUE "CompanyName", "Microsoft"
+            VALUE "FileDescription", "IIS ASP.NET Core Module Request Handler"
+            VALUE "FileVersion", FileVersionStr
+            VALUE "InternalName", "aspnetcorerh.dll"
+            VALUE "LegalCopyright", "Copyright (C) 2018"
```
Eilon: I think this should be `Copyright (c) .NET Foundation` or `Copyright (c) Microsoft Corporation`. No date.

build/Build.Settings
jkotalik: I think it is used for all of our dlls today...

jkotalik: Other dotnet dlls have the file version matching the product version.

natemcmaster: Ok. This is still bad for reproducible builds. But it is bad in the same way the rest of our stack computes file version, so we can just add this to the list of offenses to be cleaned up when we fix https://github.com/aspnet/BuildTools/issues/452

PR #610, opened 2018-02-24
change the logic on handling configuration change

pan-wang:
listen on global config change notification and check whether there is any application needs to be recycled.
Recycle will call  application shutdown first instead of rely on destructor which has issue if there is long running request (i.e., reference counter will not reach to zero and process stop will not be called). Make the application shutdown on a different thread to avoid locking the application manager hash table for long time.


src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -197,72 +222,194 @@ APPLICATION_MANAGER::RecycleApplication(
...
 
-    if(dwPreviousCounter != m_pApplicationInfoHash->Count())
+    if (dwPreviousCounter != m_pApplicationInfoHash->Count())
     {
-        // Application got recycled. Log an event
-        STACK_STRU(strEventMsg, 256);
-        if (SUCCEEDED(strEventMsg.SafeSnwprintf(
-            ASPNETCORE_EVENT_RECYCLE_CONFIGURATION_MSG,
-            pszApplicationId)))
+        if (m_hostingModel == HOSTING_IN_PROCESS)
```
bangbingsyb: For in-proc scenario, it looks like we don't actually need to perform the shallow copy of the hashtable? Since we don't need to modify m_pApplicationInfoHash at the end, all we need to know is whether the configure change affects one of the applications in the hash table, and then we only need to recycle w3wp upon that? Maybe we could have a small optimization here:?

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -197,72 +222,194 @@ APPLICATION_MANAGER::RecycleApplication(
...
 
     if (m_pApplicationInfoHash->Count() == 0)
     {
         m_hostingModel = HOSTING_UNKNOWN;
     }
 
+    ReleaseSRWLockExclusive(&m_srwLock);
 
-    if (g_fAspnetcoreRHLoadedError)
+    if(!context.MultiSz.IsEmpty() && (m_hostingModel == HOSTING_OUT_PROCESS || m_hostingModel == HOSTING_UNKNOWN))
```
bangbingsyb: For out-proc, we do update m_pApplicationInfoHash within the lock.
What would be the behavior of the requests coming to the app right after the table update?
What would be the behavior of the requests being processed in the backend app? Within shutdown limit, if their responses come, does ANCM still forward them back to clients?

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -132,6 +131,8 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
             }
         }
 
+        m_fAppOfflineFound = TRUE;
```
jkotalik: Why was this changed?

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -189,6 +189,31 @@ APPLICATION_MANAGER::GetApplicationInfo(
     return hr;
 }
 
+BOOL
+APPLICATION_MANAGER::FindConfigChangedApplication(
```
jkotalik: I'd prefer to have this as a utility function as it doesn't depend on anything else in the application manager.

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -197,72 +222,194 @@ APPLICATION_MANAGER::RecycleApplication(
...
+
+    APPLICATION* pApplication = pEntry->QueryApplication();
+
+    // Reference the application first
+    pApplication->ReferenceApplication();
+
+    // Reset application pointer to NULL
+    // The destructor of ApplictionInfo will not call ShutDown again
+    pEntry->ResetApplication();
+    HANDLE hThread = CreateThread(
```
jkotalik: Don't we need to wait on this thread to finish before calling close handle?

src/CommonLib/resources.h
```diff
@@ -28,8 +28,9 @@
...
 #define ASPNETCORE_EVENT_INPROCESS_THREAD_EXIT_STDOUT_MSG    L"Application '%s' with physical root '%s' hit unexpected managed background thread exit, ErrorCode = '0x%x. Last 4KB characters of captured stdout and stderr logs:\r\n%s"
 #define ASPNETCORE_EVENT_INPROCESS_THREAD_EXIT_MSG           L"Application '%s' with physical root '%s' hit unexpected managed background thread exit, ErrorCode = '0x%x. Please check the stderr logs for more information."
-#define ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_MSG              L"Application '%s' is recycled due to app_offline file was detected."
-#define ASPNETCORE_EVENT_RECYCLE_CONFIGURATION_MSG           L"Application '%s' is recycled due to configuration changed"
+#define ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_MSG              L"Application '%s' is recycled due to app_offline file."
+#define ASPNETCORE_EVENT_RECYCLE_CONFIGURATION_MSG           L"Application '%s' is recycled due to configuration change"
+#define ASPNETCORE_EVENT_RECYCLE_FAILURE_CONFIGURATION_MSG   L"Failed to recyle application for configuration change and had to recycle IIS worker process, Error Code = '0x%x'."
```
jkotalik: nit: "Failed to recycle application due to configuration change and had to recycle IIS worker process."

src/RequestHandler/outofprocess/outprocessapplication.cpp
```diff
@@ -57,10 +57,14 @@ __override
 VOID
 OUT_OF_PROCESS_APPLICATION::ShutDown()
 {
```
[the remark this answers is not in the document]
jkotalik: I believe that is the case.

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -197,72 +222,194 @@ APPLICATION_MANAGER::RecycleApplication(
...
+            // clean up the hash table so that the application will be informed on shutdown
+            m_pApplicationInfoHash->Apply(ShutDownApplication, NULL);
 
-    // stop filewatcher monitoring thread
-    if (m_pFileWatcher != NULL)
-    {
-        delete  m_pFileWatcher;
-        m_pFileWatcher = NULL;
+            ReleaseSRWLockExclusive(&m_srwLock);
+        }
```
jkotalik: clear the hashtable here?

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -146,10 +147,9 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
                     strEventMsg.QueryStr());
             }
 
-            m_pApplication->ShutDown();
-            m_pApplication->DereferenceApplication();
+            // shut down application in another thread to not block FCN
+            APPLICATION_MANAGER::ShutDownApplication(this, NULL);
```
jkotalik: We need to call recycle process for inprocess here.

src/AspNetCore/Src/applicationmanager.cxx
pan-wang: this method should be static. However, it depends on config and logically associated with a hashttable. Put is to into utility is not good for maintenance.

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -189,80 +191,250 @@ APPLICATION_MANAGER::GetApplicationInfo(
...
     }
 }
+
+//
+// Function used by ShutDownApplication thread to do the real shutdown
```
pan-wang: as the function got renamed, please update the comment

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -189,80 +191,250 @@ APPLICATION_MANAGER::GetApplicationInfo(
...
-                ASPNETCORE_EVENT_RECYCLE_CONFIGURATION,
-                strEventMsg.QueryStr());
+            // When we are inprocess, we need to keep the application the
+            // application manager that is being deleted. This is because we will always need to recycle the worker
+            // process and any requests that hit this worker process must be rejected (while out of process can
+            // start a new dotnet process). We will immediately call Recycle after this call.
+
+            // We need to clear the hash table here and the ref copy the original table to avoid memory leaks
+            // and keep the application referenced.
+            m_pApplicationInfoHash->Clear();
```
pan-wang: no need to call clear. For inprocess, m_pApplicationInfoHash should contain zero entry. You can do DBG_ASSERT(m_pApplicationInfoHash->Count() ==0)

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -189,80 +191,250 @@ APPLICATION_MANAGER::GetApplicationInfo(
...
+
+            // shutdown will be done in another thread
+            RecycleApplication(pRecord, NULL);
+            pRecord->DereferenceApplicationInfo();
+            path = context.MultiSz.Next(path);
+        }
+    }
 
 Finished:
+    if (table != NULL && m_hostingModel != HOSTING_IN_PROCESS)
```
pan-wang: only need to check table != NULL as already duplicated the entry back to m_pApplicationHaskTable. Otherwise, memory leak for the table

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -189,80 +191,250 @@ APPLICATION_MANAGER::GetApplicationInfo(
...
+    // Null ref here.
+    pApplication->ReferenceApplication();
+
+    pEntry->ResetApplication();
+    pApplication->ShutDown();
+    pApplication->DereferenceApplication();
+}
+
+//
+// Function used to shutdown an application
```
pan-wang: update comment

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -146,12 +155,22 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
                     strEventMsg.QueryStr());
             }
 
-            m_pApplication->ShutDown();
-            m_pApplication->DereferenceApplication();
-            m_pApplication = NULL;
+            APPLICATION_MANAGER::GetInstance()->RecycleApplicationFromManager(
```
pan-wang: you should call recycleapplication here which will just recycle specific application instance (as this is for app_offline) not for configuration change. Otherwise a subapplication may get recycled

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -146,12 +155,22 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
...
+                m_pApplication->QueryConfig()->QueryConfigPath()->QueryStr()
+            );
 
+            if (m_pConfiguration->QueryHostingModel() == APP_HOSTING_MODEL::HOSTING_OUT_PROCESS)
```
pan-wang: This can be put into RecycleApplication if you like

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -128,10 +128,13 @@ ASPNET_CORE_PROXY_MODULE::OnExecuteRequestHandler(
             FALSE
         );
 
-        DataChunk.DataChunkType = HttpDataChunkFromMemory;
-        DataChunk.FromMemory.pBuffer = (PVOID)pAppOfflineHtm->m_Contents.QueryStr();
-        DataChunk.FromMemory.BufferLength = pAppOfflineHtm->m_Contents.QueryCB();
-        pResponse->WriteEntityChunkByReference(&DataChunk);
+        if (pAppOfflineHtm != NULL)
```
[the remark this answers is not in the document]
jkotalik: There is already a debug assert now for AppOffline. I am just going to remove it.

src/CommonLib/aspnetcoreconfig.cxx
```diff
@@ -46,12 +46,16 @@ ASPNETCORE_CONFIG::GetConfig(
     _In_  IHttpServer             *pHttpServer,
     _In_  HTTP_MODULE_ID           pModuleId,
     _In_  IHttpContext            *pHttpContext,
+    _In_  HANDLE                   hEventLog,
```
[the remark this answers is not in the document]
jkotalik: I don't believe that is available in the CommonLib library.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -152,6 +160,12 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
             m_pApplication = NULL;
         }
     }
+
```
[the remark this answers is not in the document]
jkotalik: I think this is fixed.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -149,9 +158,14 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
             m_pApplication->ShutDown();
```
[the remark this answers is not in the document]
jkotalik: Believe this is handled.

src/AspNetCore/Src/applicationmanager.cxx
jkotalik: Modifying to do a pointer copy instead of a table copy as I believe it will still work.

src/AspNetCore/Src/applicationinfo.cpp
jkotalik: Not moving this for now, but I think there a few potential issues with app_offline that I would like to address first.

src/CommonLib/resources.h
```diff
@@ -28,8 +28,10 @@
 #define ASPNETCORE_EVENT_INPROCESS_THREAD_EXIT_STDERR_MSG    L"Application '%s' with physical root '%s' hit unexpected managed background thread exit, ErrorCode = '0x%x. First 4KB characters of captured stderr logs on startup:\r\n%s"
 #define ASPNETCORE_EVENT_INPROCESS_THREAD_EXIT_STDOUT_MSG    L"Application '%s' with physical root '%s' hit unexpected managed background thread exit, ErrorCode = '0x%x. Last 4KB characters of captured stdout and stderr logs:\r\n%s"
 #define ASPNETCORE_EVENT_INPROCESS_THREAD_EXIT_MSG           L"Application '%s' with physical root '%s' hit unexpected managed background thread exit, ErrorCode = '0x%x. Please check the stderr logs for more information."
-#define ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_MSG              L"Application '%s' is recycled due to app_offline file was detected."
-#define ASPNETCORE_EVENT_RECYCLE_CONFIGURATION_MSG           L"Application '%s' is recycled due to configuration changed"
+#define ASPNETCORE_EVENT_APP_IN_SHUTDOWN_MSG                 L"Application shutting down."
+#define ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_MSG              L"Application '%s' is recycled due to app_offline file."
```
Tratcher: Are there separate messages for when the app_offline is added vs when it's removed? Is there a log for the application starting up in app_offline mode?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -199,6 +201,11 @@ public string ReasonPhrase
                 _reasonPhrase = value;
             }
         }
+        
+        public IISHttpServer Server
```
Tratcher: internal?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -55,15 +60,50 @@ public Task StartAsync<TContext>(IHttpApplication<TContext> application, Cancell
 
         public Task StopAsync(CancellationToken cancellationToken)
         {
-            // TODO: Drain pending requests
+            // First call back into native saying "DON'T SEND ME ANY MORE REQUESTS"
```
Tratcher: This comment is in the wrong place, move it down

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -55,15 +60,50 @@ public Task StartAsync<TContext>(IHttpApplication<TContext> application, Cancell
...
+            void RegisterCancelation()
+            {
+                cancellationToken.Register(() =>
+                {
+                    _shutdownSignal.TrySetResult(null);
```
Tratcher: Does this need to call `lNativeMethods.http_stop_calls_into_mananged()`? Or will that be handled somewhere else in the timeout scenario?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -55,15 +60,50 @@ public Task StartAsync<TContext>(IHttpApplication<TContext> application, Cancell
...
+            {
+                _shutdownSignal.TrySetException(ex);
+            }
+
+            return _shutdownSignal.Task;
         }
 
         public void Dispose()
         {
+            // This can get triggered too early...
```
Tratcher: too early?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -55,15 +60,50 @@ public Task StartAsync<TContext>(IHttpApplication<TContext> application, Cancell
...
+            _stopping = 1;
+            _shutdownSignal.TrySetResult(null);
```
Tratcher: What about NativeMethods.http_stop_calls_into_mananged(); and http_stop_incoming_requests()?

src/CommonLib/resources.h
jkotalik: There is currently no log for when app_offline is removed. I will add one now.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
jkotalik: sorry, obsolete comment.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
jkotalik: I can just set http_stop_incoming_requests().

Tratcher: Why? Dispose may be called without StopAsync for error cases. What's to prevent requests in progress from AV'ing without stop_calls?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -183,6 +187,12 @@ public IISHttpContext CreateHttpContext(IntPtr pInProcessHandler)
                 return new IISHttpContextOfT<T>(_memoryPool, _application, pInProcessHandler, _options, _server);
             }
         }
+
+        ~IISHttpServer()
+        {
+            // If this finalize is invoked, try our best to block all requests into managed.
+            NativeMethods.http_stop_incoming_requests();
```
Tratcher: What about stop_calls?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
jkotalik: Sorry I meant NativeMethods.http_stop_calls_into_managed(). 😨

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -105,6 +150,13 @@ private static void CompleteRequest(IISHttpContext context, Task<bool> completed
             // Post completion after completing the request to resume the state machine
             context.PostCompletion(ConvertRequestCompletionResults(completedTask.Result));
 
+            if (Interlocked.Decrement(ref context.Server._outstandingRequests) == 0 && context.Server.Stopping)
```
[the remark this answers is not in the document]
jkotalik: CompleteRequest will not be skipped and PostCompletion will never fail AFAIK. @pan-wang PostCompletion can never fail, based on this: 
https://msdn.microsoft.com/en-us/library/ms692997(v=vs.90).aspx

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -160,27 +221,47 @@ IN_PROCESS_APPLICATION::OnExecuteRequest(
...
+        }
+
+        pHttpContext->GetResponse()->SetStatus(500,
+            "Internal Server Error",
+            0,
             (ULONG)E_APPLICATION_ACTIVATION_EXEC_FAILURE);
+
+        dwRequestNotificationStatus = RQ_NOTIFICATION_FINISH_REQUEST;
+    }
+    else if (m_status != APPLICATION_STATUS::RUNNING || m_fBlockCallbacksIntoManaged)
```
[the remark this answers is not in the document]
jkotalik: Would it? m_fBlockCallbacksIntoManaged is volatile so it would be atomic.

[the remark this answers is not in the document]
Tratcher: You have a (small) gap between checking the bool and calling the delegate.

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -189,80 +201,306 @@ APPLICATION_MANAGER::GetApplicationInfo(
...
+            // When we are inprocess, we need to keep the application the
+            // application manager that is being deleted. This is because we will always need to recycle the worker
+            // process and any requests that hit this worker process must be rejected (while out of process can
+            // start a new dotnet process). We will immediately call Recycle after this call.
+            DBG_ASSERT(m_pApplicationInfoHash->Count() == 0);
+            delete m_pApplicationInfoHash;
+
+            // We don't want to delete the table as m_pApplicationInfoHash = table
+            fKeepTable = TRUE;
+            m_pApplicationInfoHash = table;
```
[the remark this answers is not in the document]
jkotalik: I deleted m_pApplicationInfoHash right before this...

pan-wang:
Let's commit this one first and fix the issues parallel instead of holding it for long time

src/RequestHandler/inprocess/inprocessapplication.cpp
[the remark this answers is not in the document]
jkotalik: It's still the same race, except worse because we are guaranteed to AV. We still need to check if the handler is null or not.

PR #612, opened 2018-02-26
websocket enabled check

pan-wang:
check whether websocket is enabled in configuration. So that we will not forward the upgrade header if websocket is not enabled. Also add Release cal to avoid memory leak


src/CommonLib/aspnetcoreconfig.h
```diff
@@ -168,11 +169,17 @@ class ASPNETCORE_CONFIG : IHttpStoredContext
...
 
     BOOL
-        QueryStdoutLogEnabled()
+    QueryStdoutLogEnabled()
     {
         return m_fStdoutLogEnabled;
     }
 
+    BOOL
+    QueryWebSocketEnable()
```
bangbingsyb: QueryWebSocketEnable**d** to keep name consistency?

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -165,7 +165,7 @@ FORWARDING_HANDLER::OnExecuteRequestHandler()
     //
     // Mark request as websocket if upgrade header is present.
     //
-    if (g_fWebSocketSupported)
+    if (pApplication->QueryConfig()->QueryWebSocketEnable())
```
jkotalik: I think we should still check win8+ too?

jkotalik: Remove g_fWebSocketSupported.

src/CommonLib/aspnetcoreconfig.cxx
```diff
@@ -161,6 +162,7 @@ ASPNETCORE_CONFIG::Populate(
     BSTR                            bstrBasicAuthSection = NULL;
     BSTR                            bstrAnonymousAuthSection = NULL;
     BSTR                            bstrAspNetCoreSection = NULL;
+    BSTR                            bstrWebscoketSection = NULL;
```
nil4: Typo: `Webscoket`

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: no need to check win8 as on the websocket configuration section is empty win7 and QueryWebSocketEnable return false. g_fWebSocketedSupported is used for initialize winhttpwebsocket function pointers and static init/cleanup

src/CommonLib/aspnetcoreconfig.h
pan-wang: good point

src/CommonLib/aspnetcoreconfig.cxx
pan-wang: fixed

src/RequestHandler/outofprocess/forwardinghandler.cpp
jkotalik: Then shouldn't g_fWebSocketedSupported be equal to pApplication->QueryConfig()->QueryWebSocketEnabled()?

PR #619, opened 2018-02-27
fixing a stress issue

pan-wang:
use interlockexchange to prevent multi-threads to access m_pDisconnect simultaneously


src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -314,15 +310,7 @@ FORWARDING_HANDLER::OnExecuteRequestHandler()
     m_RequestStatus = FORWARDER_DONE;
 
     //disbale client disconnect callback
-    if (m_pDisconnect != NULL)
-    {
-        if (fHandleSet)
-        {
-            m_pDisconnect->ResetHandler();
-        }
-        m_pDisconnect->CleanupStoredContext();
```
bariscaglar: You are losing this CleanupStoredContext when switching to RemoveRequest. Is that acceptable?

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -314,15 +310,7 @@ FORWARDING_HANDLER::OnExecuteRequestHandler()
     m_RequestStatus = FORWARDER_DONE;
 
     //disbale client disconnect callback
```
bariscaglar: typo "disbale" while you're fixing around it....

src/RequestHandler/outofprocess/forwardinghandler.cpp
bangbingsyb: This looks like a bug in the original code (before the change) to me. My understanding is that the class that implements IHttpStoredContext and its CleanupStoredContext method should not call the method. Instead, CleanupStoredContext should only be called once by iiscore at the end of the lifetime of the context container, otherwise a double cleanup may cause an AV.

jkotalik:
Pending other's feedback.

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: yes. no need to do cleanstoredcontext

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: done

PR #622, opened 2018-03-01
fix missing event log descriptions

pan-wang:
aspnetcoremodule.rc missed the include of aspnetcore_msg.rc. Also update the resource include folder in order to include the compiled MSG00001.bin file


jkotalik:
This looks good to me. Have you confirmed locally?

pan-wang:
yes. I tested locally

Issue #584, opened 2018-02-13
Disable HTTP Upgrade feature when WebSockets are not available


anurse:
## Out-of-proc

* ANCM sets `ASPNETCORE_WEBSOCKETS_SUPPORTED` environment variable to `true` if WebSockets module is installed **and** OS Version > Windows 7. Or `false` otherwise
* IIS Integration middleware reads this environment variable (defaults to `true` if not present)
* If `ASPNETCORE_WEBSOCKETS_SUPPORTED` = `false`: Remove the `IHttpUpgradeFeature` from the request features
* Else: do nothing

## In-proc

* `IISHttpServer` checks if WebSockets if WebSockets module is installed **and** OS Version > Windows 7 (either in the managed code, or by reading data populated by the native code, whichever is easier/possible)
* If `true`: Install the `IHttpUpgradeFeature` as before
* Else: **Do not** install the `IHttpUpgradeFeature`

jkotalik:
@anurse for the inproc and out of proc scenarios, they are both on a per request basis? Therefore,the IISHttpContext should be the one to remove the IHttpUgradeFeature?

anurse:
IIS should be modifying the request features yes. However, it's certainly the case that the decision to allow the upgrade feature can be made once per process, since it's not possible to enable the WebSockets module or upgrade the OS while the process is running ;). So yeah, it would be in the `IISHttpContext` that you would remove the feature, but you could use a global static variable to determine it.

PR #624, opened 2018-03-02
Adds websocket feature detection, test cleanup.

jkotalik:
For #584.

- For inprocess, we check the ServerVariable WEBSOCKET_VERSION on the first request and check if it is set.
- For out of process, we set the environment variable ASPNETCORE_IIS_WEBSOCKETS_SUPPORTED and check the value in the IISMiddleware. 

I also merged the IIS test projects together and divided them based on in process and out of process as I had too many copied apphost.config files between the two.

@pan-wang I also refactored the environment_var_hash into two classes as there were many static helpers that shouldn't belong to a hashtable.


src/RequestHandler/environmentvariablehelpers.h
```diff
@@ -0,0 +1,403 @@
...
+            goto Finished;
+        }
+        if (FAILED(hr = pIISAuthEntry->Initialize(ASPNETCORE_IIS_AUTH_ENV_STR, strIisAuthEnvValue.QueryStr())) ||
+            FAILED(hr = pEnvironmentVarTable->InsertRecord(pIISAuthEntry)))
+        {
+            goto Finished;
+        }
+
+        // We only need to set the WEBSOCKET_SUPPORTED environment variable for out of process
+        if (hostingModel == HOSTING_OUT_PROCESS)
```
jkotalik: The only change made to this file from before is here @pan-wang

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -25,8 +25,7 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
     internal abstract partial class IISHttpContext : NativeRequestContext, IDisposable
     {
         private const int MinAllocBufferSize = 2048;
-
-        private static bool UpgradeAvailable = (Environment.OSVersion.Version >= new Version(6, 2));
+        private static bool? UpgradeAvailable;
```
anurse: This will likely need locking or something to make it thread-safe. Since it's static it can be accessed by multiple concurrent requests. You probably don't need locking, just to make sure that it's initialized atomically. It's OK if multiple requests read the server variable, since they should all set the same value.

jkotalik: Seem my comment on setting the UpgradeAvailable. I don't think it's needed. All requests that come into the server will return the same value.

anurse: It's not a bool though, it's a `Nullable<bool>` which has two fields and could be sheared if it's being written while another thread reads it.

Agreed, if you can make it an atomic type, it should be safe without locks, multiple threads may do the same check but they'll all come to the same conclusion. Maybe just use an int with three states: `1` Uninitialized, `2` WebSockets Available, `3` WebSockets Not Available.

src/RequestHandler/environmentvariablehelpers.h
Tratcher: It's a little odd to pass in hostingModel only for this condition. It seems like fWebsocketsEnabled should be a tri-state field: enabled, disabled, no-op.

test/IISTestSite/Startup.cs
```diff
@@ -35,6 +35,7 @@ public void Configure(IApplicationBuilder app)
             app.Map("/CheckEnvironmentLongValueVariable", CheckEnvironmentLongValueVariable);
             app.Map("/CheckAppendedEnvironmentVariable", CheckAppendedEnvironmentVariable);
             app.Map("/CheckRemoveAuthEnvironmentVariable", CheckRemoveAuthEnvironmentVariable);
+            app.Map("/UpgradeFeatureDetection", UpgradeFeatureDetection);
```
Tratcher: Why do you have both /UpgradeFeatureDetection and StartupUpgradeFeatureDetection?

src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs
```diff
@@ -74,6 +75,14 @@ public static IWebHostBuilder UseIISIntegration(this IWebHostBuilder hostBuilder
             var path = hostBuilder.GetSetting(ServerPath) ?? Environment.GetEnvironmentVariable($"ASPNETCORE_{ServerPath}");
             var pairingToken = hostBuilder.GetSetting(PairingToken) ?? Environment.GetEnvironmentVariable($"ASPNETCORE_{PairingToken}");
             var iisAuth = hostBuilder.GetSetting(IISAuth) ?? Environment.GetEnvironmentVariable($"ASPNETCORE_{IISAuth}");
+            var websocketsSupported = hostBuilder.GetSetting(IISWebSockets) ?? Environment.GetEnvironmentVariable($"ASPNETCORE_{IISWebSockets}");
+
+            bool isWebSocketsSupported;
+            if (!bool.TryParse(websocketsSupported, out isWebSocketsSupported))
+            {
+                // If the websocket support variable is not set, we will always fallback to assuming websockets are enabled.
+                isWebSocketsSupported = true;
```
Tratcher: Add an OS check as fallback.

src/RequestHandler/environmentvariablehelpers.h
jkotalik: yeah, I'll clean that up.

test/IISTestSite/Startup.cs
jkotalik: One is for inprocess and one is for out of process. I can merge them together if you would prefer.

Tratcher: Yes please, they appear to do the same thing.

jkotalik: I merged the test classes together, however I realized there are still issues with merging the test sites together between in-process and out of process. When we call dotnet publish, we provide web.config file with hostingModel=inprocess/outofprocess and the processPath/arguments. I thought I could add this to the apphostconfig file we provide when calling iisexpress, however I am running into issues there. I'll ask you tomorrow about this.

jkotalik:
@pan-wang can you give one last passover for the native side of this change?

PR #633, opened 2018-03-07
adding general logging event for future use

pan-wang:


jkotalik:
I'd say we add warning and error messages that are the same as the general message, for the sake of ease of use.

pan-wang:
we only need event ID for general info, warning, and error event. The message is just a string input. why do we need different messages?

jkotalik:
Every time we log, we do something like:

```cpp

        // Application got recycled. Log an event
        STACK_STRU(strEventMsg, 256);
            ASPNETCORE_EVENT_RECYCLE_CONFIGURATION_MSG,
            pszApplicationId)))
        {
            UTILITY::LogEvent(g_hEventLog,
                EVENTLOG_INFORMATION_TYPE,
                ASPNETCORE_EVENT_RECYCLE_CONFIGURATION,
                strEventMsg.QueryStr());
        }
```

It is nice to have the MSG and error types match even though they are the same thing.

pan-wang:
it seems we don't need the msg at all. Just event ID is sufficient enough. The input to UTULITY::LogEvent is just a string. We don't need to duplicate the string, i.e., use strEventMsg. Current code, we uses strEventMsg because we put some parameters such as hr value and et al into the string which is not available for Generic one. Also Generic one does not support localization.

jkotalik:
That's fine by me.

PR #634, opened 2018-03-07
Update RecycleApplication function

pan-wang:
Remove the input parameter hostingModel as we don't need it. We can get the hosting model info from the application's configuration. Fix a memory leak in recycling out-of-process application.


src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -480,19 +480,18 @@ APPLICATION_MANAGER::DoRecycleApplication(
...
+    DBG_ASSERT(pApplication != NULL);
 
     // Reference the application first
     pApplication->ReferenceApplication();
 
-    if (hostingModel == APP_HOSTING_MODEL::HOSTING_OUT_PROCESS)
+    if (pApplication->QueryConfig()->QueryHostingModel() == HOSTING_OUT_PROCESS)
     {
-        pEntry->ClearApplication();
+        pEntry->ClearAndDereferenceApplication();
```
jkotalik: Are we confident that we can dereference the application too here?

pan-wang: yes, as we reference it first, the reference counter become 2, we need dereference it to reduce the counter to 1 and then deference after recycle. The application then gets deleted.

jkotalik: sounds good.

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -480,19 +480,18 @@ APPLICATION_MANAGER::DoRecycleApplication(
...
 {
-
     APPLICATION* pApplication = pEntry->QueryApplication();
+    DBG_ASSERT(pApplication != NULL);
 
     // Reference the application first
     pApplication->ReferenceApplication();
 
-    if (hostingModel == APP_HOSTING_MODEL::HOSTING_OUT_PROCESS)
+    if (pApplication->QueryConfig()->QueryHostingModel() == HOSTING_OUT_PROCESS)
```
jkotalik: Actually, can you add a comment explaining why we need to clear the application only for out of process?

pan-wang: done

PR #639, opened 2018-03-07
moving recycle and shutdown to applicationinfo

pan-wang:
move recycle and shutdown into application_info for better abstraction and remove a lock


src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -533,3 +522,101 @@ APPLICATION_INFO::FindNativeAssemblyFromHostfxr(
...
+            if (pApplication->QueryConfig()->QueryHostingModel() == HOSTING_OUT_PROCESS)
+            {
+                //
+                // For inprocess, need to set m_pApplication to NULL first to
+                // avoid mapping new request to the recycled application.
+                // Outofprocess application instance will be created for new request
+                // For inprocess, as recycle will lead to shutdown later, leave m_pApplication
+                // to not block incoming requests till worker process shutdown
+                //
+                m_pApplication = NULL;
```
jkotalik: I think there is a potential race here with setting m_pApplication to NULL and getting a new request. Here is the scenario:

App_offline is dropped, and we immediately receive a request. The request calls EnsureApplicationCreated() and that succeeds. RecycleApplication is then called, causing the application to be null. We then call QueryApplication and that application is NULL. We have a debug assert there assuming the application, which is incorrect. 

I think we need to consider a few more race conditions here too.

pan-wang: yes. there is possibility of NULL ref. I changed the function of get application to reference the application first.

src/AspNetCore/Inc/applicationinfo.h
```diff
@@ -128,19 +130,29 @@ class APPLICATION_INFO
...
+    // ExtractApplication will increase the reference counter of the application
+    // Caller is responsible for dereference the application.
+    // Otherwise memory leak
+    // Open Issue: perf impact of the lock as this is in hot path per request
+    //
+    VOID
+    ExtractApplication(APPLICATION** ppApplication)
     {
-        return m_pApplication;
+        AcquireSRWLockExclusive(&m_srwLock);
```
jkotalik: offline discussion: make this a shared lock as this doesn't modify the state of the application, only gets it.

src/AspNetCore/Inc/applicationinfo.h
```diff
@@ -4,6 +4,8 @@
 #pragma once
 #define API_BUFFER_TOO_SMALL 0x80008098
 
+extern BOOL     g_fRecycleProcessCalled;
```
jkotalik: We call RecycleProcess for inprocess here: https://github.com/aspnet/IISIntegration/blob/00b1948937e8d5cdcc7559c8d8aafda1613479b4/src/RequestHandler/inprocess/inprocessapplication.cpp#L172. Will there be issues if g+fRecycleProcess is false if we have called recycle already?

pan-wang: I thought about it. This is impossible for inprocess. If we failed to create the recycle thread, the recycle function will not be called.

Issue #638, opened 2018-03-07
Address build warnings and enable warnings as errors for the C++ projects


natemcmaster:
The compiler currently emits the following warnings for a release build.
```
IISIntegration        | acache.cxx(223): warning C4189: 'pfl': local variable is initialized but not referenced [C:\b\w\33bdfc1cae7b2a38\modules\IISIntegration\src\IISLib\IISLib.vcxproj]
IISIntegration        | stringa.cpp(1059): warning C4100: 'fFailIfNoTranslation': unreferenced formal parameter [C:\b\w\33bdfc1cae7b2a38\modules\IISIntegration\src\IISLib\IISLib.vcxproj]
IISIntegration        | stringa.cpp(1698): warning C4189: 'hr': local variable is initialized but not referenced [C:\b\w\33bdfc1cae7b2a38\modules\IISIntegration\src\IISLib\IISLib.vcxproj]
IISIntegration        | stringa.cpp(1700): warning C4189: 'cchValue': local variable is initialized but not referenced [C:\b\w\33bdfc1cae7b2a38\modules\IISIntegration\src\IISLib\IISLib.vcxproj]
IISIntegration        | stringu.cpp(1124): warning C4189: 'hr': local variable is initialized but not referenced [C:\b\w\33bdfc1cae7b2a38\modules\IISIntegration\src\IISLib\IISLib.vcxproj]
IISIntegration        | stringu.cpp(1126): warning C4189: 'cchValue': local variable is initialized but not referenced [C:\b\w\33bdfc1cae7b2a38\modules\IISIntegration\src\IISLib\IISLib.vcxproj]
IISIntegration        | c:\b\w\33bdfc1cae7b2a38\modules\iisintegration\src\iislib\stringu.cpp(802): warning C4754: Conversion rules for arithmetic operations in the comparison at c:\b\w\33bdfc1cae7b2a38\modules\iisintegration\src\iislib\stringu.cpp(797) mean that one branch cannot be executed. Cast '(... + 2)' to 'ULONG64' (or similar type of 8 bytes). [C:\b\w\33bdfc1cae7b2a38\modules\IISIntegration\src\AspNetCore\AspNetCore.vcxproj]
IISIntegration        | acache.cxx(223): warning C4189: 'pfl': local variable is initialized but not referenced [C:\b\w\33bdfc1cae7b2a38\modules\IISIntegration\src\IISLib\IISLib.vcxproj]
[...]

PR #641, opened 2018-03-08
Remove build warnings and turn on warning as errors in IISLib.

jkotalik:
For #638. Seems we forgot to decrease build warning level on IISLib from 4 to 3. 
cc/ @natemcmaster


natemcmaster:
Thanks for a taking a look at this so quickly. :shipit:

Issue #629, opened 2018-03-06
Flaky test: SynchronousReadAndWriteTests


jkotalik:
Seeing this failing sporadically local and on our CI. Disabling for now as I debug.

```csharp
Failed   Microsoft.AspNetCore.Server.IIS.FunctionalTests.SynchronousReadAndWriteTests.ReadAndWriteSynchronously
Error Message:
 System.Net.Http.HttpRequestException : Error while copying content to a stream.
---- System.IO.IOException : The read operation failed, see inner exception.
-------- System.Net.Http.WinHttpException : The connection with the server was terminated abnormally
Stack Trace:
   at System.Net.Http.HttpContent.<LoadIntoBufferAsyncCore>d__48.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Net.Http.HttpClient.<FinishSendAsyncBuffered>d__58.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Microsoft.AspNetCore.Server.IIS.FunctionalTests.SynchronousReadAndWriteTests.<ReadAndWriteSynchronously>d__2.MoveNext() in C:\Users\jukotali\code\IISIntegration\test\IISIntegration.IISServerFunctionalTests\SynchronousReadAndWriteTests.cs:line 32
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #650, opened 2018-03-09
Re-adding Read/Write tests and removing unused websocket class.

jkotalik:
For #629 . For some reason, the IISExpressAppPool runs into issues with IIS, which the CLR4IntegratedAppPool doesn't.

I'd recommend looking at the PR ignoring white space: https://github.com/aspnet/IISIntegration/pull/650?w=1


pakrym:
What was the issue? What's different about new .config files?

jkotalik:
It uses the CLR4IntegratedAppPool instead of the IISExpressAppPool. I talked with @pan-wang and he said there shouldn't be any functional difference between the two. I'll chat with @BillHiebert once he comes back from vacation as he creates the app pool.

test/IISIntegration.FunctionalTests/Inprocess/SynchronousReadAndWriteTests.cs
```diff
@@ -23,7 +23,7 @@ public SynchronousReadAndWriteTests(IISTestSiteFixture fixture)
             _fixture = fixture;
         }
 
-        [ConditionalFact (Skip = "See https://github.com/aspnet/IISIntegration/issues/629")]
+        [ConditionalFact]
```
pakrym: `[Fact]` ?

jkotalik:
It uses the CLR4IntegratedAppPool instead of the IISExpressAppPool. I talked with @pan-wang and he said there shouldn't be any functional difference between the two. I'll chat with @BillHiebert once he comes back from vacation as he creates the app pool. 

Let me minimize the difference though as there as the diff was larger than I thought.

test/IISIntegration.FunctionalTests/Inprocess/SynchronousReadAndWriteTests.cs
```diff
@@ -86,6 +86,7 @@ public void ReadAndWriteSlowConnection()
                     catch (Exception)
                     {
                         // Exceptions can be thrown based on ipv6 support
+                        return;
```
pakrym: Why are we successfully returning on connection error?

test/IISIntegration.FunctionalTests/Inprocess/SynchronousReadAndWriteTests.cs
jkotalik: Everything in the repo is a conditional fact because we check if the update IIS Schema is installed or not: https://github.com/aspnet/IISIntegration/blob/dev/test/IISIntegration.FunctionalTests/Utilities/IISExpressSupportsInProcessHostingAttribute.cs

pakrym:
How much difference is there between host configs? Can we make the required changes programatically?

jkotalik:
> How much difference is there between host configs? Can we make the required changes programmatically based on some common source?

For sure, it would reduce duplication of the apphost config files. We can do the same thing we are doing with [ANCMPath], [Port], etc. and replace it with provided values in the deployment parameters.

pakrym:
I would prefer that, right now it's impossible to tell what is the difference between two files without manually diffing them.

jkotalik:
To only have one apphost.config file, we need to update the IISExpress Deployer. I'll make a PR shortly to provide options to replace certain options in the config.

jkotalik:
@pakrym So I was incorrect about there being a problem with the IISExpressAppPool. There was an ordering issue in the IIS Modules that moved up four modules. @pan-wang thinks this was done to guarantee the the modules would be activated before other modules, like IISFreb.

jkotalik:
@pakrym offline approved this.

PR #656, opened 2018-03-13
Fix two AVs with InProcess 

jkotalik:
- Calling hostfxr_get_native_search_directories multiple times was causing an AV in hostpolicy parsing arguments.
- Closing the IIS Express window forcibly caused an AV. To fix it, I check for process detach in the request handler to not call into managed.

There was also some duplicated calls to parsing arguments. 

@jhkimnew  I have done my best to verify that multiple request don't trigger an AV. Can you verify this behavior.


jhkimnew:
@jkotalik I sent a mail with the repro steps.

jkotalik:
@jhkimnew and I talked offline and we didn't see the issue reoccurring after this fix. @pan-wang can you do one final pass through as it has changed a bit due to rebasing?

PR #660, opened 2018-03-13
add port info to event log

pan-wang:


src/RequestHandler/outofprocess/serverprocess.cxx
```diff
@@ -747,6 +747,7 @@ SERVER_PROCESS::StartProcess(
 
     while (dwRetryCount > 0)
     {
+        m_dwPort = 0;
```
jkotalik: Put this in CleanUp?

pan-wang: Its better to put here as if the failure happens earlier, we don't want to log an initialized value

PR #665, opened 2018-03-14
Add support for unit testing from console

pakrym:
C++ VS Test is great!


test/AspNetCoreModuleTests/NativeTests.targets
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <PropertyGroup>
+    <VsTestConsole>$(VCIDEInstallDir)..\CommonExtensions\Microsoft\TestWindow\vstest.console.exe</VsTestConsole>
```
natemcmaster: Did you try using `dotnet vstest`?

pakrym: Dotnet fails on imports in vcprojes.

natemcmaster: Played with it a little more just b/c I wanted to understand what happens here. It looks like there is a test adapter that ships only in VS. vstest.console itself in `dotnet vstest` is capable of handling this, but the adapter is not shipped in the CLI sdk.

PR #669, opened 2018-03-14
fixing fowarding handler for AsyncCompletion

pan-wang:


src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -536,103 +555,131 @@ REQUEST_NOTIFICATION_STATUS
...
+        if (fClientError || m_fClientDisconnected)
         {
-            pResponse->SetStatus(400, "Bad Request", 0, HRESULT_FROM_WIN32(WSAECONNRESET));
+            if (!m_fResponseHeadersReceivedAndSet)
+            {
+                pResponse->SetStatus(400, "Bad Request", 0, HRESULT_FROM_WIN32(WSAECONNRESET));
+            }
+            else
+            {
+                //
```
jkotalik: Get rid of the else and just make this a comment.

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -34,7 +34,11 @@ FORWARDING_HANDLER::FORWARDING_HANDLER(
     m_BytesToReceive(0),
     m_BytesToSend(0),
     m_fWebSocketEnabled(FALSE),
-    m_pWebSocket(NULL)
+    m_pWebSocket(NULL),
+    m_dwHandlers (1), // default http handler
+    m_fDoneAsyncCompletion(FALSE),
+    m_fHttpHandleInClose(FALSE),
+    m_fWebSocketHandleInClose(FALSE)
```
jkotalik: Initialize m_fServerResetConn?

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -422,40 +426,21 @@ REQUEST_NOTIFICATION_STATUS
     if (TlsGetValue(g_dwTlsIndex) != this)
     {
         DBG_ASSERT(TlsGetValue(g_dwTlsIndex) == NULL);
-        AcquireSRWLockShared(&m_RequestLock);
+        AcquireSRWLockExclusive(&m_RequestLock);
```
jkotalik: Add a comment explaining why this acquire lock is exclusive.

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -536,103 +555,131 @@ REQUEST_NOTIFICATION_STATUS
...
         }
     }
 
-    //
-    // Finish the request on failure.
-    // Let IIS pipeline continue only after receiving handle close callback
-    // from WinHttp. This ensures no more callback from WinHttp
-    //
-    if (m_hRequest != NULL)
+    if (m_pWebSocket != NULL && !m_fWebSocketHandleInClose)
```
jkotalik: Do you need to set m_pWebsocket to NULL here?

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -2618,25 +2703,26 @@ FORWARDING_HANDLER::TerminateRequest(
...
 
-    //
-    // If the request is a websocket request, initiate cleanup.
-    //
-    if (m_pWebSocket != NULL)
+    if (!m_fHttpHandleInClose)
     {
-        m_pWebSocket->TerminateRequest();
+        m_fHandleClosedDueToClient = fClientInitiated;
+        //panwang
```
jkotalik: remove this.

jkotalik:
@pan-wang I think messed up your merge. Check if your local changes are the ones you want and force push to this branch.

pan-wang:
it's a bad rebase. I am working on it

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: nice catch

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: done

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -465,33 +450,49 @@ REQUEST_NOTIFICATION_STATUS
...
+        hr = m_pWebSocket->ProcessRequest(this, m_pW3Context, m_hRequest, &fWebSocketUpgraded);
+        if (fWebSocketUpgraded)
+        {
+            // WinHttp WebSocket handle has been created, bump the counter so that remember to close it
+            // and prevent from premature postcomplation and unexpected callback from winhttp
+            InterlockedIncrement(&m_dwHandlers);
         }
 
-        hr = m_pWebSocket->ProcessRequest(this, m_pW3Context, m_hRequest, &fWebSocketUpgrade);
         if (FAILED(hr))
```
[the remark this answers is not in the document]
pan-wang: no. We need check fwebsocketupgrade first as the upgrade may success while processrequest may fail

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -523,6 +525,23 @@ REQUEST_NOTIFICATION_STATUS
 
     default:
         DBG_ASSERT(m_RequestStatus == FORWARDER_DONE);
+        if (m_hRequest == NULL && m_pWebSocket == NULL)
```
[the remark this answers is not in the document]
pan-wang: m_fHttpHandleInClose is set before closing wihttp handle (i.e., winhttp handle closing callback has not been triggered yet). We need do this NULL check to make sure the handle closing callback was called and handle was set to null so that there will be no more callback from WinHttp.

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: yes

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: no. We need call m_pWebSocket->Terminate() in the destructor to free the memory

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -1492,76 +1575,94 @@ None
...
 
         if (fEndRequest)
         {
-            // only postCompletion after WINHTTP_CALLBACK_STATUS_HANDLE_CLOSING
-            // so that no further WinHttp callback will be called
-            // in case of websocket, m_hRequest has already been closed after upgrade
-            // websocket will handle completion
-            m_fFinishRequest = TRUE;
-            fDoPostCompletion = TRUE;
+            fDoPostCompletion = (dwHandlers == 0 && !m_fFinishRequest);
```
jkotalik: nit: you don't need to check m_fFinishRequest here; you can always just set m_fFinishRequest to true if dwHandlers == 0.

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -536,103 +568,133 @@ REQUEST_NOTIFICATION_STATUS
...
+
+        // This is just a safety guard to prevent from returning non pending status no more once
+        // which should never happen
+        if (!m_fDoneAsyncCompletion)
+        {
+            m_fDoneAsyncCompletion = TRUE;
+        }
+        else
+        {
+            retVal = RQ_NOTIFICATION_PENDING;
```
jkotalik: I don't understand why we are returning PENDING instead of CONTINUE here.

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -1252,27 +1321,41 @@ None
     if (TlsGetValue(g_dwTlsIndex) != this)
     {
         DBG_ASSERT(TlsGetValue(g_dwTlsIndex) == NULL);
+        if (m_RequestStatus != FORWARDER_RECEIVED_WEBSOCKET_RESPONSE)
+        {
+            // Webscoket has already been guarded by critical section
+            // Only require exclisive lock for non-websocket scenario which has duplex channel
+            // Otherwise, there will be a deadlock
+            AcquireSRWLockExclusive(&m_RequestLock);
```
jkotalik: Make helper functions to lock and unlock as we do it 5 times throughout the file? Like:
```c++
VOID
FORWARDING_HANDLER::LockExclusive()
 {
            AcquireSRWLockExclusive(&m_RequestLock);
            TlsSetValue(g_dwTlsIndex, this);
            fExclusiveLocked = TRUE;
            DBG_ASSERT(TlsGetValue(g_dwTlsIndex) == this);
}
```

Same for releasing and shared locks.

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: this logic is that current thread hit an error (ideally should return non-pending status to IIS). However, another thread already returned (as m_fDoneAsyncCompletion was true). Current thread cannot return non-pending anymore otherwise IIS pipe will get messed as the later may already move to another module

src/RequestHandler/outofprocess/forwardinghandler.h
```diff
@@ -174,11 +174,33 @@ class FORWARDING_HANDLER : public REQUEST_HANDLER
     BOOL                                m_fResponseHeadersReceivedAndSet;
     BOOL                                m_fResetConnection;
     BOOL                                m_fHandleClosedDueToClient;
-    BOOL                                m_fFinishRequest;
-    BOOL                                m_fHasError;
     BOOL                                m_fDoReverseRewriteHeaders;
+    BOOL                                m_fServerResetConn;
+    volatile  BOOL                      m_fClientDisconnected;
```
bangbingsyb: Cannot find where m_fClientDisconnected is set?

src/RequestHandler/outofprocess/forwardinghandler.h
```diff
@@ -174,11 +174,33 @@ class FORWARDING_HANDLER : public REQUEST_HANDLER
     BOOL                                m_fResponseHeadersReceivedAndSet;
     BOOL                                m_fResetConnection;
     BOOL                                m_fHandleClosedDueToClient;
-    BOOL                                m_fFinishRequest;
-    BOOL                                m_fHasError;
     BOOL                                m_fDoReverseRewriteHeaders;
+    BOOL                                m_fServerResetConn;
```
bangbingsyb: volatile for m_fServerResetConn as well?

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -421,77 +432,80 @@ REQUEST_NOTIFICATION_STATUS
...
             hr = HRESULT_FROM_WIN32(GetLastError());
             goto Failure;
         }
-        break;
+        retVal = RQ_NOTIFICATION_PENDING;
+        goto Finished;
+    }
 
+    //
+    // Begins normal completion handling. There is already a shared acquired
```
bangbingsyb: Do you mean "exclusive lock acquired" in the comment?

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -536,103 +568,133 @@ REQUEST_NOTIFICATION_STATUS
...
 
+    if (retVal != RQ_NOTIFICATION_PENDING)
+    {
+
+        DBG_ASSERT(m_dwHandlers == 0);
+        RemoveRequest();
+
+        // This is just a safety guard to prevent from returning non pending status no more once
+        // which should never happen
+        if (!m_fDoneAsyncCompletion)
```
bangbingsyb: Should we flip m_fDoneAsyncCompletion inside the exclusive lock? I feel that doing so after the lock released might let  certain racing condition slip the guard here.

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -1252,27 +1321,41 @@ None
...
+            fSharedLocked = TRUE;
+            DBG_ASSERT(TlsGetValue(g_dwTlsIndex) == this);
+        }
+    }
 
-        AcquireSRWLockShared(&m_RequestLock);
-        TlsSetValue(g_dwTlsIndex, this);
-        fLockAcquired = TRUE;
-        DBG_ASSERT(TlsGetValue(g_dwTlsIndex) == this);
+    if (fEndRequest)
```
bangbingsyb: Consider to give a clearer name for this flag - hard to tell the difference between fEndRequest and m_fFinishRequest until realized it just indicated winhttp handle closing.

src/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -1492,76 +1575,94 @@ None
...
 Finished:
-
-    if (fLockAcquired)
+    //
+    // Since we use TLS to guard WinHttp operation, call PostCompletion instead of
+    // IndicateCompletion to allow cleaning up the TLS before thread reuse.
+    // Never post after the request has been finished for whatever reason
+    //
+    // Only postCompletion after all WinHttp handle got closed,
+    // i.e., received WINHTTP_CALLBACK_STATUS_HANDLE_CLOSING callback for all handles
```
bangbingsyb: Consider to clarify "all handles": essentially it means two winhttp handles: http and websocket

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: good suggestion. Changed to fHandleClosing

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: done

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: fixed

src/RequestHandler/outofprocess/forwardinghandler.h
pan-wang: no need. it's inside a lock

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: fixed

src/RequestHandler/outofprocess/forwardinghandler.cpp
pan-wang: Yes.

src/RequestHandler/outofprocess/forwardinghandler.h
pan-wang: fixed

Issue #604, opened 2018-02-22
ANCM: INPROCESS: stack overflow error on Microsoft.AspNetCore.Server.IISIntegration.dll module


jhkimnew:
When I tried to download a large htm file (123 MB), I got exception error from Microsoft.AspNetCore.Server.IISIntegration.dll, stackoverflow.
@jkotalik  Please take a look at this issue.

jkotalik:
https://support.microsoft.com/en-us/help/932909/by-default-the-maximum-stack-size-of-a-thread-that-is-created-in-a-nat
Seems to be pretty small... this also may be an issue with using the send file feature.

jhkimnew:
@jkotalik Please set Milestone for this issue. I have no permission to set Milestone.

shirhatti:
@jkotalik will verify and close. Also, permalinks :trollface:

davidfowl:
@jkotalik  @davidfowl do you still think we need to chunk writes when directly writing to IIS? yes. We can look at changing how we flush as an alternative as well. Talk to me about this tomorrow.

PR #683, opened 2018-03-16
Heap allocate HTTP_DATA_CHUNK if the size of the response is too large. 

jkotalik:
Follow up to #604 . Though we added the pause and resume thresholds to the OutputProducer, if we do a single large write in application code, it can still cause a huge write, causing stack overflow. There are a few ways we can mitigate this, but the simpilest is if there is a large write, heap allocate the chunks instead of stackalloc.


src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -12,6 +12,8 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
 {
     internal partial class IISHttpContext
     {
+        private const int StackChunkLimit = 128;
```
pakrym: Seems to small, even for 256k stack 1k should be fine.

pakrym: I see, it's a stuct, how big is it?

Tratcher: Units?

test/IISTestSite/Startup.cs
```diff
@@ -628,5 +629,15 @@ private void TestInvalidWriteOperations(IApplicationBuilder app)
                 await context.Response.WriteAsync(success ? "Success" : "Failure");
             });
         }
+
+        private void LargeResponseFile(IApplicationBuilder app)
+        {
+            app.Run(async ctx =>
+            {
+                var tempFile = Path.GetTempFileName();
+                File.WriteAllText(tempFile, new string('a', 200000000));
```
Tratcher: delete the file afterwards?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
jkotalik: I think we can increase it. The struct size is only 16 bytes: https://github.com/aspnet/HttpSysServer/blob/231cbffe1a5cc0abce11e2d0685683ed74fc7ad5/shared/Microsoft.AspNetCore.HttpSys.Sources/NativeInterop/HttpApiTypes.cs#L78

pakrym: 128 is good if it's a segment count.

test/IISTestSite/Startup.cs
jkotalik: done.

jkotalik:
Verifying that the old build fails with a S.O. if the write is too large.

jkotalik:
@pakrym it definitely fails without this change.

PR #678, opened 2018-03-15
Re-Enable ANCM OutOfProcess tests 

jkotalik:
For #498

PR #684, opened 2018-03-16
Re-Enable ANCM OutOfProcess tests  (#678)

jkotalik:


test/IISIntegration.FunctionalTests/OutOfProcess/HelloWorldTest.cs
```diff
@@ -86,7 +84,10 @@ public Task HelloWorld_IISExpress_CoreClr_X64_Portable()
 
                         response = await deploymentResult.HttpClient.GetAsync("/Auth");
                         responseText = await response.Content.ReadAsStringAsync();
-                        Assert.True("backcompat;Windows".Equals(responseText) || "latest;null".Equals(responseText), "Auth");
+
+                        // We adapted the Http.config file to be used for inprocess too. We specify WindowsAuth is enabled
```
Tratcher: You don't always want WindowsAuth to be enabled, most sites won't do that.

PR #685, opened 2018-03-16
Cleanup ReadAsync calls, make IISHttpContext private. 

jkotalik:
Offline feedback for 5e74b36c53038116c6a789a1e2a9b084f59dc377 that @davidfowl had.


src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpRequestBody.cs
```diff
@@ -39,7 +39,18 @@ public override int Read(byte[] buffer, int offset, int count)
 
         public override unsafe Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            return _httpContext.ReadAsync(buffer, offset, count, cancellationToken);
+            if (buffer == null)
```
davidfowl: Can we just let the Memory constructor handle this?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpRequestBody.cs
```diff
@@ -39,7 +39,18 @@ public override int Read(byte[] buffer, int offset, int count)
...
+            {
+                throw new ArgumentNullException(nameof(buffer));
+            }
+            if (count == 0)
```
davidfowl: This shouldn't throw.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpRequestBody.cs
jkotalik: Doesn't throw with a null buffer.

davidfowl: Then maybe it shouldn't.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpRequestBody.cs
Tratcher: Stream.Read/Async should not return 0 until the end of the body. You need a non-zero buffer to do that.

PR #696, opened 2018-03-19
Cleanup ReadAsync calls, make IISHttpContext private

jkotalik:
Accidentally made a PR to dev. I reverted it and am now doing the PR to release/2.1


src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpRequestBody.cs
```diff
@@ -39,7 +39,9 @@ public override int Read(byte[] buffer, int offset, int count)
 
         public override unsafe Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            return _httpContext.ReadAsync(buffer, offset, count, cancellationToken);
+            var memory = new Memory<byte>(buffer, offset, count);
```
Tratcher: So what does happen when buffer is null or count is zero?

jkotalik: We allow it. After chatting with @davidfowl, a use case could be that someone calls ReadAsync without a buffer to know when data is available. This reduces the amount of idle allocated memory.

Tratcher: Does it actually work? You removed the tests.

jkotalik: Yeah, I'll need to verify this doesn't null ref.

davidfowl: Why would it null ref?

jkotalik:
🆙 📅

jkotalik:
@pakrym can you take a look at this? This is some general reaction that Fowler didn't like about https://github.com/aspnet/IISIntegration/commit/5e74b36c53038116c6a789a1e2a9b084f59dc377

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpRequestBody.cs
pakrym: Memory supports null buffer, it'll return empty span.

jkotalik:
I reverted it in dev to first merge into release. Will merge into dev afterwards.

PR #708, opened 2018-03-21
Add Managed to native project dependencies

pakrym:
To fix VS build ordering


pakrym:
Also made test sites that use ancm runtime specific, changed them to only pull native dll's they require and run the correct version of IIS Express.

pakrym:
I also changed every test site to be runnable on any platform setting.

build/testsite.props
```diff
@@ -0,0 +1,29 @@
+<Project>
+
+  <PropertyGroup>
+    <RuntimeIdentifiers>win7-x64;win7-x86</RuntimeIdentifiers>
+    <Platforms>x64;x86</Platforms>
+    <IISExpressAppHostConfig>$(MSBuildThisFileDirectory)applicationhost.config</IISExpressAppHostConfig>
+    <NativePlatform Condition="'$(Platform)' == 'AnyCPU'">x64</NativePlatform>
+    <NativePlatform Condition="'$(NativePlatform)' == ''">$(Platform)</NativePlatform>
+    <DotNetPath>$(userprofile)\.dotnet\$(NativePlatform)\dotnet.exe</DotNetPath>
```
jkotalik: We cool with this only pointing to the user profile?

pakrym: I'm not a fan either but couldn't figure out any other way of selecting dotnet based on bitness.
In every case I could imagine this used dotnet would be installed by `KoreBuild` and exist in the user profile.

natemcmaster: This is how our build determines where DOTNET_HOME:

https://github.com/aspnet/IISIntegration/blob/be08942b5775d0949b23596a8e0e254a64aca0b9/run.ps1#L176-L179

PR #712, opened 2018-03-22
Disable test parallelization

jkotalik:
Trying to fix issues like: https://ci.appveyor.com/project/aspnetci/iisintegration/build/1.0.517 where dotnet publish fails. The main test site is published by 3 different tests, which can cause races.


pakrym:
Fine for now, we'll look at better ways to fix in in deployer infrastructure.

PR #714, opened 2018-03-22
Add support for running test sites on IIS from VS

pakrym:
Supports both x64 and x86


build/testsite.props
```diff
@@ -4,26 +4,37 @@
...
+    <None Include="$(MSBuildThisFileDirectory)..\src\RequestHandler\bin\$(Configuration)\$(NativeFolder)\aspnetcorerh.dll" CopyToOutputDirectory="PreserveNewest" Visible="true" Link="$(NativePlatform)\%(FileName)%(Extension)" />
+    <None Include="$(MSBuildThisFileDirectory)..\src\RequestHandler\bin\$(Configuration)\$(NativeFolder)\aspnetcorerh.pdb" CopyToOutputDirectory="PreserveNewest" Visible="true" Link="$(NativePlatform)\%(FileName)%(Extension)" />
+    <None Include="$(MSBuildThisFileDirectory)..\src\AspNetCore\bin\$(Configuration)\$(NativeFolder)\aspnetcore.dll" CopyToOutputDirectory="PreserveNewest" Visible="true" Link="$(NativePlatform)\%(FileName)%(Extension)" />
+    <None Include="$(MSBuildThisFileDirectory)..\src\AspNetCore\bin\$(Configuration)\$(NativeFolder)\aspnetcore.pdb" CopyToOutputDirectory="PreserveNewest" Visible="true" Link="$(NativePlatform)\%(FileName)%(Extension)" />
+  </ItemGroup>
+
   <PropertyGroup Condition="'$(NativePlatform)' == 'x86'">
-    <IISExpress>$(MSBuildProgramFiles32)\IIS Express\iisexpress.exe</IISExpress>
+    <IISExpressPath>$(MSBuildProgramFiles32)\IIS Express\iisexpress.exe</IISExpressPath>
```
jkotalik: Nit: On a 32 bit machine, what does this return? I think it returns Program Files.

build/testsite.props
```diff
@@ -4,26 +4,37 @@
...
+    <IISPath>$(SystemRoot)\SysWOW64\inetsrv\w3wp.exe</IISPath>
```
jkotalik: Same here, I don't know what SysWoW64 is on a 32 bit machine.

build/testsite.props
pakrym: Yes, which would be correct.

build/testsite.props
pakrym: I would like to see person trying to open this solution on 32 bit windows :)

jkotalik: Yeah its minor, I don't care much.

Issue #686, opened 2018-03-16
Invalid hostfxr path/arguments should be considered when checking backup location for dotnet.


jhkimnew:
@pan-wang Can you take a look at this issue?

STEPS TO REPRO:

1. Set bogus value for processPath and arguments and set inprocess mode as the following web.config example:

```
...
<aspNetCore processPath="bogus" arguments="bogus" startupTimeLimit="30" shutdownTimeLimit="17" stdoutLogEnabled="true" stdoutLogFile=".\stdout" forwardWindowsAuthToken="true" hostingModel="inprocess">
```  

2. Start app with IISExpress
3. Send a request to the app

EXPECTED:

No crash

ACTUAL:

IISExpress process is crashed with the below AV

```
(7f8.4e18): Unknown exception - code c0020001 (first chance)
(7f8.4e18): Unknown exception - code c0020001 (!!! second chance !!!)
KERNELBASE!RaiseException+0x68:
00007ff9`01fd57f8 488b8c24c0000000 mov     rcx,qword ptr [rsp+0C0h] ss:000000c4`deefea60=0000983e3358f3f0
0:026> kn
 # Child-SP          RetAddr           Call Site
00 000000c4`deefe9a0 00007ff8`d310fd8d KERNELBASE!RaiseException+0x68 [minkernel\kernelbase\xcpt.c @ 922] 
01 (Inline Function) --------`-------- coreclr!COMPlusThrowBoot+0x21 [e:\a\_work\1606\s\src\vm\excep.cpp @ 198] 
02 000000c4`deefea80 00007ff8`d307f75e coreclr!TheUMEntryPrestubWorker+0xc2b6d [e:\a\_work\1606\s\src\vm\dllimportcallback.cpp @ 1064] 
03 000000c4`deefead0 00007ff8`fd0463c0 coreclr!TheUMEntryPrestub+0x3e [E:\A\_work\1606\s\src\vm\amd64\UMThunkStub.asm @ 67] 
04 000000c4`deefeb40 00007ff8`fd05a5e6 WINHTTP!HTTP_USER_REQUEST::_IndicateInformational+0xf0 [onecore\net\winhttp\httpcore\usrreq.cpp @ 737] 
05 000000c4`deefebc0 00007ff8`fd074a5d WINHTTP!HTTP_USER_REQUEST::OnSendingRequest+0x606 [onecore\net\winhttp\httpcore\usrreq.cpp @ 2936] 
[...]

PR #715, opened 2018-03-22
Expand environment variables processPath and arguments, reject invalid processPaths.

jkotalik:
For #686. This should have no impact on stress/perf scenarios, purely usability.


src/CommonLib/hostfxr_utility.cpp
```diff
@@ -171,10 +171,23 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
     STRU                        struEventMsg;
     std::vector<std::wstring>   vVersionFolders;
     DWORD                       dwPosition;
+    STACK_STRU(struExpandedProcessPath, 256);
```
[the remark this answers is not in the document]
jhkimnew: MAX_PATH limitation should not be used and we should support longer path. See https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx#maxpath

src/CommonLib/utility.cxx
```diff
@@ -627,3 +627,44 @@ UTILITY::LogEvent(
...
+
+    if (dwNumBytesRead == 0)
+    {
+        hr = HRESULT_FROM_WIN32(GetLastError());
+        goto Finished;
+    }
+    else if (dwNumBytesRead >= struExpandedString->QuerySizeCCH())
+    {
+        // If the buffer is too small, resize and read again.
+        if (FAILED(hr = struExpandedString->Resize(dwNumBytesRead + 1)))
```
pakrym: Are you adding a single byte?

jkotalik: This is for the null terminator.

src/CommonLib/hostfxr_utility.cpp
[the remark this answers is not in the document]
jkotalik: This is an initial size to use. We resize if the buffer is longer.

src/CommonLib/utility.cxx
pakrym: Add a test for resizing scenario

jkotalik: Done. Do you want the E2E scenario with a custom web.config here?

pakrym: I would prefer having functional tests that does it if it doesn't take too long.

jkotalik: Added two functional tests. They are a bit verbose. I would like to create a follow up PR to make a helper that creates/deploys the application and returns the deployment result.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -171,10 +171,23 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
+    STACK_STRU(struExpandedArguments, MAX_PATH);
 
-    // Convert the process path an absolute path.
+    // Expand the process path and arguments with evironment strings
+    if (FAILED(hr = UTILITY::GetExpandedEnvironmentString(pcwzProcessPath, &struExpandedProcessPath)))
+    {
+        goto Finished;
+    }
+
+    if (FAILED(hr = UTILITY::GetExpandedEnvironmentString(pcwzArguments, &struExpandedArguments)))
```
pan-wang: why do we need to expand the parameters? hostfxr should be able to resolve them

src/CommonLib/utility.cxx
```diff
@@ -627,3 +627,44 @@ UTILITY::LogEvent(
         fwprintf(stderr, L"ERROR: %s\n", pstrMsg);
     }
 }
+
+HRESULT
+UTILITY::GetExpandedEnvironmentString(
```
pan-wang: you don't need this function. STRU has a method called CopyAndExpandEnvironmentStrings. you can use it directly

test/AspNetCoreModuleTests/utility_tests.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+        TEST_METHOD(PassUnexpandedString_ExpandsResult)
+        {
+            HRESULT hr = S_OK;
+            PCWSTR unexpandedString = L"ANCM_TEST_ENV_VAR";
+            PCWSTR unexpandedStringValue = L"foobar";
+            STRU   struExpandedString;
+            SetEnvironmentVariable(L"ANCM_TEST_ENV_VAR", unexpandedStringValue);
+
+            hr = UTILITY::GetExpandedEnvironmentString(L"%ANCM_TEST_ENV_VAR%", &struExpandedString);
```
pan-wang: it will be better to put the input as "%ANCM_TEST_ENV_VAR%/test" to make sure the resolved string is still valid.
How about if user give two parameters and both contains environment variables?

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -171,10 +171,26 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
     DWORD                       dwPosition;
+    BOOL                        fIsDotnetPath;
+    STACK_STRU(struExpandedProcessPath, MAX_PATH);
+    STACK_STRU(struExpandedArguments, MAX_PATH);
 
-    // Convert the process path an absolute path.
+    // Expand the process path and arguments with evironment strings
+    struExpandedProcessPath.CopyAndExpandEnvironmentStrings(pcwzProcessPath);
+
+    if (FAILED(struExpandedProcessPath.CopyAndExpandEnvironmentStrings(pcwzProcessPath)))
```
pan-wang: missing hr value

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -204,8 +221,29 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
+                goto Finished;
+            }
+        }
+        else
+        {
+            // If the processPath file does not exist and it doesn't include dotnet.exe or dotnet
+            // then it is an invalid argument.
+            hr = E_INVALIDARG;
+            if (SUCCEEDED(struEventMsg.SafeSnwprintf(
+                L"Invalid or unknown processPath provided in web.config: processPath = %s, ErrorCode = '0x%x'.",
```
pan-wang: for now, don't hardcode the string here as it will block localization in the future.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -171,10 +171,26 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
     STRU                        struEventMsg;
     std::vector<std::wstring>   vVersionFolders;
     DWORD                       dwPosition;
+    BOOL                        fIsDotnetPath;
+    STACK_STRU(struExpandedProcessPath, MAX_PATH);
+    STACK_STRU(struExpandedArguments, MAX_PATH);
 
-    // Convert the process path an absolute path.
+    // Expand the process path and arguments with evironment strings
+    struExpandedProcessPath.CopyAndExpandEnvironmentStrings(pcwzProcessPath);
```
pan-wang: please remove this duplication

src/CommonLib/hostfxr_utility.cpp
jkotalik: Do these all still go in resources.h?

pan-wang: yes

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -164,78 +164,93 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
-                goto Finished;
-            }
         }
         else
         {
+            //
             // If the processPath file does not exist and it doesn't include dotnet.exe or dotnet
             // then it is an invalid argument.
+            //
             hr = E_INVALIDARG;
```
pan-wang: set error code to ERROR_FILE_NOT_FOUND will be more helpful than E_INVALIDARG

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -164,78 +164,93 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
             if (SUCCEEDED(struEventMsg.SafeSnwprintf(
-                L"Invalid or unknown processPath provided in web.config: processPath = %s, ErrorCode = '0x%x'.",
+                L"Invalid or unknown processPath provided in web.config: processPath = %s, ErrorCode = '0x%x'.", // TODO put in resource
```
pan-wang: this is just a quick in resource.h  no need to wait for next pr

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -495,41 +387,216 @@ HOSTFXR_UTILITY::ParseHostfxrArguments(
...
+        goto Finished;
+    }
+
+    // At this point, we are calling where.exe to find dotnet.
+    hr = InvokeWhereToFindDotnet(struAbsolutePathToDotnet);
+
+    // If where failed, we search program files for dotnet as a backup location.
+    if (struAbsolutePathToDotnet->IsEmpty())
+    {
+        GetAbsolutePathToDotnetFromProgramFiles(struAbsolutePathToDotnet);
```
pan-wang: hresult

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -588,108 +658,146 @@ HOSTFXR_UTILITY::FindDotnetExePath(
     // 
     if (!GetExitCodeProcess(processInformation.hProcess, &dwExitCode))
     {
-        goto Fallback;
+        goto Finished;
```
pan-wang: should set hr to error?

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -588,108 +658,146 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
+    if (dwExitCode != 0)
     {
-        // Where succeeded. 
-        // Reset file pointer to the beginning of the file. 
-        dwFilePointer = SetFilePointer(hStdOutReadPipe, 0, NULL, FILE_BEGIN);
-        if (dwFilePointer == INVALID_SET_FILE_POINTER)
-        {
-            goto Fallback;
-        }
+        goto Finished;
```
pan-wang: should set hr to some error code?

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -588,108 +658,146 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
-            // This shouldn't ever be this large. We could continue to call ReadFile in a loop,
-            // however I'd rather error out here and report an issue.
-            goto Fallback;
-        }
+    // Where succeeded. 
+    // Reset file pointer to the beginning of the file. 
+    dwFilePointer = SetFilePointer(hStdOutReadPipe, 0, NULL, FILE_BEGIN);
+    if (dwFilePointer == INVALID_SET_FILE_POINTER)
+    {
+        goto Finished;
```
pan-wang: set hr to some error

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -588,108 +658,146 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
-        {
-            goto Finished;
-        }
+    //
+    // As the call to where.exe succeeded (dotnet.exe was found), ReadFile should not hang.
+    // TODO consider putting ReadFile in a separate thread with a timeout to guarantee it doesn't block.
+    //
+    if (!ReadFile(hStdOutReadPipe, pzFileContents, READ_BUFFER_SIZE, &dwNumBytesRead, NULL))
+    {
+        goto Finished;
```
pan-wang: error code

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -588,108 +658,146 @@ HOSTFXR_UTILITY::FindDotnetExePath(
...
+    }
+    if (dwNumBytesRead >= READ_BUFFER_SIZE)
+    {
+        // This shouldn't ever be this large. We could continue to call ReadFile in a loop,
+        // however if someone had this many dotnet.exes on their machine, I'd rather notify an error.
```
pan-wang: error code

jkotalik:
@Tratcher do you know if there are issues with doing a standalone publish from the net461? Seems to consistently be failing.

Tratcher:
Not that I'm aware of. These these have been disabled so something may have regressed. Get what you have merged and we'll look at net451 separately.

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -457,41 +387,215 @@ HOSTFXR_UTILITY::ParseHostfxrArguments(
...
 
-//
-// Invoke where.exe to find the location of dotnet.exe
-// Copies contents of dotnet.exe to a temp file
-// Respects path ordering.
 HRESULT
-HOSTFXR_UTILITY::FindDotnetExePath(
-    _Out_ STRU* struDotnetPath
+HOSTFXR_UTILITY::GetAbsolutePathToDotnet(
+    _Inout_ STRU* struAbsolutePathToDotnet
```
pan-wang: it's better to rename it as pstruAbsolutePathToDotnet as it is a pointer

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -457,41 +387,215 @@ HOSTFXR_UTILITY::ParseHostfxrArguments(
...
+        goto Finished;
+    }
+
+Finished:
+    return hr;
+}
+
+
+BOOL
+HOSTFXR_UTILITY::InvokeWhereToFindDotnet(
```
pan-wang: you may want to add some comments about the goal of this function

src/CommonLib/hostfxr_utility.h
```diff
@@ -51,8 +51,28 @@ class HOSTFXR_UTILITY
...
     HRESULT
-    FindDotnetExePath(
+    GetAbsolutePathToDotnet(
         STRU*   struDotnetLocation
     );
+
+    static
+    HRESULT
+    GetAbsolutePathToHostFxr(
+        _In_ STRU* struAbsolutePathToDotnet,
```
pan-wang: coding guideline on parameter naming. pointer should start with p***
similar to other functions

src/CommonLib/hostfxr_utility.h
```diff
@@ -51,8 +51,28 @@ class HOSTFXR_UTILITY
...
+    GetAbsolutePathToDotnet(
         STRU*   struDotnetLocation
     );
+
+    static
+    HRESULT
+    GetAbsolutePathToHostFxr(
+        _In_ STRU* pStruAbsolutePathToDotnet,
+        _In_ HANDLE hEventLog,
+        _Out_ STRU* struAbsolutePathToHostfxr
```
pan-wang: pstruAbsolutePathToHostfxr

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -457,41 +387,220 @@ HOSTFXR_UTILITY::ParseHostfxrArguments(
...
+    }
+
+Finished:
+
+    return hr;
+}
+
+HRESULT
+HOSTFXR_UTILITY::GetAbsolutePathToHostFxr(
+    STRU* struAbsolutePathToDotnet,
```
pan-wang: input parameter does not match with the header file

src/CommonLib/hostfxr_utility.h
```diff
@@ -51,8 +51,28 @@ class HOSTFXR_UTILITY
 
     static
     HRESULT
-    FindDotnetExePath(
+    GetAbsolutePathToDotnet(
```
pan-wang: pstruDotnetLocation

PR #718, opened 2018-03-22
IIS ANCM stress test application

bangbingsyb:
Add ANCM stress test app to IISIntegration so that our stress machines can directly pull the latest source from the repo, build the app locally with the latest dependencies.

Removed the dependency of the app to WebSockets middleware based on the suggestion from Chris and Justin to avoid circular dependency.


samples/ANCMStressTestSample/Startup.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        private void WebSocket(IApplicationBuilder app)
+        {
+            app.Run(async context =>
+            {
+                var upgradeFeature = context.Features.Get<IHttpUpgradeFeature>();
+
+                // Generate WebSocket response headers
+                string key = string.Join(", ", context.Request.Headers[Constants.Headers.SecWebSocketKey]);
```
Tratcher: `string key = context.Request.Headers[Constants.Headers.SecWebSocketKey].ToString();`

samples/ANCMStressTestSample/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+                    factory.AddConsole();
+                })
+                .UseKestrel()
+                .UseStartup<Startup>()
+                .Build();
+
+            AppLifetime = (IApplicationLifetime)host.Services.GetService(typeof(IApplicationLifetime));
+            AppLifetime.ApplicationStopping.Register(
+                () => {
+                    AppLifetimeStopping = true;
```
jkotalik: Having this been a bool seems racy to me. Provide a cancellation token that will be canceled instead? Also you can pass in that cancellation token to Send/ReceiveAsync

samples/ANCMStressTestSample/WebSockets/Constants.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            public const string Upgrade = "Upgrade";
+            public const string UpgradeWebSocket = "websocket";
+            public const string Connection = "Connection";
+            public const string ConnectionUpgrade = "Upgrade";
+            public const string SecWebSocketKey = "Sec-WebSocket-Key";
+            public const string SecWebSocketVersion = "Sec-WebSocket-Version";
+            public const string SecWebSocketProtocol = "Sec-WebSocket-Protocol";
+            public const string SecWebSocketAccept = "Sec-WebSocket-Accept";
+            public const string SupportedVersion = "13";
```
jkotalik: Why do we need SupportedVersion?

Tratcher: Some of these were from request headers, the test code assumes the request headers are correct. Purge as needed.

samples/ANCMStressTestSample/Startup.cs
bangbingsyb: Good catch! Fixed.

samples/ANCMStressTestSample/Program.cs
bangbingsyb: This was copied from Jeong Hwan's original test app for functional test. Yes, for stress test purpose, we need to handle application shutdown more carefully. Made a change to how we handle close handshake.

samples/ANCMStressTestSample/WebSockets/Constants.cs
bangbingsyb: Purged the unused constants.

samples/ANCMStressTestSample/WebSockets/Constants.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace ANCMStressTestApp
+{
+    public static class Constants
+    {
+        public static class Headers
+        {
+            public const string Upgrade = "Upgrade";
+            public const string UpgradeWebSocket = "websocket";
+            public const string Connection = "Connection";
+            public const string ConnectionUpgrade = "Upgrade";
```
pan-wang: the value is same Upgrade. You can remove this one

samples/ANCMStressTestSample/Startup.cs
```diff
@@ -0,0 +1,229 @@
...
+                var buffer = new byte[1024 * 4];
+                var result = await webSocket.ReceiveAsync(new ArraySegment<byte>(buffer), token);
+                bool closeFromServer = false;
+                string closeFromServerCmd = "CloseFromServer";
+                int closeFromServerLength = closeFromServerCmd.Length;
+
+                while (!result.CloseStatus.HasValue && !token.IsCancellationRequested && !closeFromServer)
+                {
+                    if (result.Count == closeFromServerLength &&
+                        System.Text.Encoding.ASCII.GetString(buffer).Substring(0, result.Count) == closeFromServerCmd)
```
pan-wang: do you  want it be case sensitive?

bangbingsyb: Yes, it is better to be case sensitive. For a regular data frame, the stress client will generate random data with random length within the given length range. So there is a very small chance that the generated data frame can be identical to the special "server close request" frame. Making it case sensitive will further reduce such chance.

samples/ANCMStressTestSample/WebSockets/Constants.cs
bangbingsyb: Removed

samples/ANCMStressTestSample/Program.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public class Program
+    {
+        public static IApplicationLifetime AppLifetime;
+        public static CancellationTokenSource Cts = new CancellationTokenSource();
+
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
+                .ConfigureLogging((_, factory) =>
```
jkotalik: Don't you need to call UseIISIntegration()?

samples/ANCMStressTestSample/Startup.cs
jkotalik: using System.Text; and then condense this to Encoding.ASCII.GetString();

bangbingsyb: Changed.

samples/ANCMStressTestSample/Program.cs
bangbingsyb: I think we need it. Looks like the IISSample doesn't call it, either. What's the reason?

test/ANCMStressTestApp/Startup.cs
```diff
@@ -171,7 +171,9 @@ private void WebSocket(IApplicationBuilder app)
                 // Get the WebSocket object
                 var ws = WebSocketProtocol.CreateFromStream(opaqueTransport, isServer: true, subProtocol: null, keepAliveInterval: TimeSpan.FromMinutes(2));
 
-                await Echo(ws, Program.Cts.Token);
+                var appLifetime = app.ApplicationServices.GetRequiredService<IApplicationLifetime>();
+
+                await Echo(ws, appLifetime.ApplicationStopping);
```
bangbingsyb: Looks good. So the IApplicationLifetime.ApplicationStopping itself is a cancellation token.

pakrym:
We should add correct launchSettings and `  <Import Project="..\..\build\testsite.props" />` to every test app so it is F5 runable.

Issue #591, opened 2018-02-15
[In-proc] Unhandled request exception results in an empty 200 response


Tratcher:
I intentionally commented out some of my auth settings in appsettings.json which causes the auth middleware to throw an exception per request. The exception handler middleware sets the status to 500, retries, and throws again. Hosting logs the request completed with a 500 status, but the browser reports it as an empty 200.

It appears that when in-proc catches an unhandled request exception it does not set the status code to 500, completes the response with the default 200 status.

Logs:
```
Hosting environment: Production
Content root path: C:\inetpub\wwwroot\
Application started. Press Ctrl+C to shut down.
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET https://localhost/  
fail: Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware[1]
      An unhandled exception has occurred while executing the request.
System.ArgumentException: Options.ClientId must be provided
Parameter name: ClientId
   at Microsoft.AspNetCore.Authentication.OpenIdConnect.OpenIdConnectOptions.Validate()
   at Microsoft.AspNetCore.Authentication.RemoteAuthenticationOptions.Validate(String scheme)
   at Microsoft.AspNetCore.Authentication.AuthenticationHandler`1.<InitializeAsync>d__42.MoveNext()
   at Microsoft.AspNetCore.Authentication.AuthenticationHandlerProvider.<GetHandlerAsync>d__5.MoveNext()
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.<Invoke>d__6.MoveNext()
   at Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware.<Invoke>d__6.MoveNext()
fail: Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware[3]
      An exception was thrown attempting to execute the error handler.
System.ArgumentException: Options.ClientId must be provided
Parameter name: ClientId
   at Microsoft.AspNetCore.Authentication.OpenIdConnect.OpenIdConnectOptions.Validate()
   at Microsoft.AspNetCore.Authentication.RemoteAuthenticationOptions.Validate(String scheme)
   at Microsoft.AspNetCore.Authentication.AuthenticationHandler`1.<InitializeAsync>d__42.MoveNext()
[...]

PR #719, opened 2018-03-22
Do not return 200 for exceptions

pakrym:
https://github.com/aspnet/IISIntegration/issues/591

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
@@ -332,10 +332,12 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
         public unsafe void SendResponseHeaders(bool appCompleted)
         {
             // Verifies we have sent the statuscode before writing a header
-            var reasonPhraseBytes = Encoding.UTF8.GetBytes(ReasonPhrase ?? ReasonPhrases.GetReasonPhrase(StatusCode));
+            var reasonPhrase = string.IsNullOrWhiteSpace(ReasonPhrase) ? ReasonPhrases.GetReasonPhrase(StatusCode) : ReasonPhrase;
+            var reasonPhraseBytes = Encoding.UTF8.GetBytes(reasonPhrase);
 
             fixed (byte* pReasonPhrase = reasonPhraseBytes)
             {
+                Debug.Assert((IntPtr)pReasonPhrase != IntPtr.Zero);
                 // This copies data into the underlying buffer
                 NativeMethods.http_set_response_status_code(_pInProcessHandler, (ushort)StatusCode, pReasonPhrase);
             }
--- a/test/IISIntegration.FunctionalTests/Inprocess/ResponseHeaderTests.cs
+++ b/test/IISIntegration.FunctionalTests/Inprocess/ResponseHeaderTests.cs
@@ -40,5 +40,13 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
             Assert.Equal("1", headerValues.First());
             Assert.Equal("2", headerValues.Last());
         }
+
+        [ConditionalFact]
+        public async Task ErrorCodeIsSetForExceoptionDuringRequest()
+        {
+            var response = await _fixture.Client.GetAsync("Throw");
+            Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
+            Assert.Equal("Internal Server Error", response.ReasonPhrase);
+        }
     }
 }
--- a/test/IISTestSite/Startup.cs
+++ b/test/IISTestSite/Startup.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Net;
+using System.Reflection;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -332,10 +332,12 @@ private void SetErrorResponseHeaders(int statusCode)
         public unsafe void SendResponseHeaders(bool appCompleted)
         {
             // Verifies we have sent the statuscode before writing a header
-            var reasonPhraseBytes = Encoding.UTF8.GetBytes(ReasonPhrase ?? ReasonPhrases.GetReasonPhrase(StatusCode));
+            var reasonPhrase = string.IsNullOrWhiteSpace(ReasonPhrase) ? ReasonPhrases.GetReasonPhrase(StatusCode) : ReasonPhrase;
```
Tratcher: never use OrWhiteSpace is ASP.NET libraries.... it does some strange unicode stuff. Stick with IsNullOrEmpty

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -332,10 +332,12 @@ private void SetErrorResponseHeaders(int statusCode)
...
+            var reasonPhraseBytes = Encoding.UTF8.GetBytes(reasonPhrase);
```
Tratcher: If I set a status code that had an undefined reason phrase would I have the same issue? E.g. GetReasonPhrase can return null or empty (?)

test/IISTestSite/Startup.cs
```diff
@@ -23,33 +24,16 @@ public class Startup
...
-            app.Map("/ReadAndWriteCopyToAsync", ReadAndWriteCopyToAsync);
-            app.Map("/ReadAndWriteEchoTwice", ReadAndWriteEchoTwice);
-            app.Map("/ReadAndWriteSlowConnection", ReadAndWriteSlowConnection);
-            app.Map("/WebsocketRequest", WebsocketRequest);
-            app.Map("/UpgradeFeatureDetection", UpgradeFeatureDetection);
-            app.Map("/TestInvalidReadOperations", TestInvalidReadOperations);
-            app.Map("/TestInvalidWriteOperations", TestInvalidWriteOperations);
-            app.Map("/TestReadOffsetWorks", TestReadOffsetWorks);
-            app.Map("/LargeResponseFile", LargeResponseFile);
+            foreach (var method in GetType().GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance))
```
Tratcher: 🐌 👏

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
jkotalik: Was the reason this wasn't returning 500 is because we never sent the reason phrase?

Tratcher: It sounds like http_set_response_status_code threw because of a bad reason phrase.

pakrym: It didn't threw exactly but `pInProcessHandler->QueryHttpContext()->GetResponse()->SetStatus(statusCode, pszReason);` call did not succeed.

test/IISIntegration.FunctionalTests/IISIntegration.FunctionalTests.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Server.IISIntegration\Microsoft.AspNetCore.Server.IISIntegration.csproj" />
+    <ProjectReference Include="..\IISTestSite\IISTestSite.csproj" />
```
jkotalik: Any reason to keep this?

pakrym: Forces the rebuild of test site when running test.

test/IISIntegration.FunctionalTests/Inprocess/ResponseHeaderTests.cs
```diff
@@ -40,5 +40,25 @@ public ResponseHeaders(IISTestSiteFixture fixture)
             Assert.Equal("1", headerValues.First());
             Assert.Equal("2", headerValues.Last());
         }
+        
+        [ConditionalFact]
+        public async Task ErrorCodeIsSetForExceoptionDuringRequest()
```
Tratcher: Exceoption?

test/IISIntegration.FunctionalTests/Inprocess/ResponseHeaderTests.cs
```diff
@@ -40,5 +40,25 @@ public ResponseHeaders(IISTestSiteFixture fixture)
...
+        {
+            var response = await _fixture.Client.GetAsync("Throw");
+            Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
+            Assert.Equal("Internal Server Error", response.ReasonPhrase);
+        }
+
+        [ConditionalTheory]
+        [InlineData(1, "custom", "custom")]
```
Tratcher: 👀 Why does it allow a status code below 101?

PR #724, opened 2018-03-23
Revert version to preview2

JunTaoLuo:
I'm guessing this was messed up during branching? Shouldn't the version be preview2?


JunTaoLuo:
@pakrym it's because of https://github.com/aspnet/IISIntegration/commit/88d40dacf93c9fbb5dfca8326363511c467e66f0

davidfowl:
Is this right https://github.com/aspnet/IISIntegration/blob/29343cc81627289a849c1399542902b67a8ec941/korebuild.json#L3?

pakrym:
You should've reverted that entire commit. I'm not sure how that happend but I used Github merge conflict resolution and it created that commit. Never again.

Issue #698, opened 2018-03-19
Disable loading RH from inetsrv for in-proc


shirhatti:

jkotalik:
I'm going to first investigate the viability of doing a deps file rewrite before running the application such that we don't need this feature flag.

Eilon:
@shirhatti / @jkotalik / @muratg - is this a must-have for Preview2? If not, let's move to RC1.

PR #732, opened 2018-03-27
Disable global loading for in process.

jkotalik:
Pending work items:
- Create tests in ServerTests to use Microsoft.AspNetCore.Server.IIS package (and re-enable out of process tests).
- Update IISExpressDeployer to modify web.config file for inprocess.


pakrym:
How does the error look when there is no aspnetcorerh in app folder?

jkotalik:
A generic 500 with nothing informative put in the event log. I'll fix that.

jkotalik:
For #698

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -264,16 +264,21 @@ APPLICATION_INFO::FindRequestHandlerAssembly()
...
-            // TODO eventually make this fail for in process loading.
-            hr = FindNativeAssemblyFromGlobalLocation(&struFileName);
-            if (FAILED(hr))
+            if (FAILED(hr = FindNativeAssemblyFromHostfxr(&struFileName)))
             {
                 goto Finished;
             }
         }
-
+        else
```
pan-wang: We don't want to block the option of deploying the new aspnetcorerh.dll at runtime store which contains new features or changes for out of process. It also blocks customer's aspnetcorerh.dll. I would image we should try hostfxr first and then global for out of process

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -264,16 +264,21 @@ APPLICATION_INFO::FindRequestHandlerAssembly()
             goto Finished;
         }
 
-        if (FAILED(hr = FindNativeAssemblyFromHostfxr(&struFileName)))
+        if (m_pConfiguration->QueryHostingModel() == APP_HOSTING_MODEL::HOSTING_IN_PROCESS)
```
pan-wang: we need log a windows event in case of failure. Nothing was logged currently

src/AspNetCore/Src/applicationinfo.cpp
jkotalik: You can still install out of process changes globally. Are you suggesting to allow for this with out of process?

src/AspNetCore/Src/applicationinfo.cpp
jkotalik: done.

pan-wang: this is critical error. should be logged as error instead of info.
please also make sure this log is not pre request. It should be logged only once per applicationinfo

jkotalik: It can't be logged more than once because we set g_fAspnetcoreRHLoadedError if this function failed.

jkotalik:
🆙 📅

jkotalik:
@pan-wang it will take a reasonable amount of work to support loading the request handler for out of proc. I can file a separate issue for it.

pakrym:
@pan-wang do you have additional comments? We didn't correctly support loading out of proc request handler from appbase before this PR. If we want this feature lets file and track it separately.

pan-wang:
The change looks good to me.
However if we plan to support multiple versions of ANCM side by side, I think we should allow loading request handler from the application folder or shared framework, i.e, claiming request handler is part of the application even for out-of-process.   This is business decision.

pan-wang:
I opened issue #758 to track the new feature ask.

jkotalik:
This good to go then? (seeking approval)

PR #735, opened 2018-03-28
Add deps file RH injection. Do not build targets if assembly exists to prevent test errors 

pakrym:


build/testsite.props
```diff
@@ -34,7 +34,29 @@
...
   </PropertyGroup>
 
+  <PropertyGroup>
+    <TestTargetsAssembly>$(MSBuildThisFileDirectory)..\test\TestTasks\bin\$(Configuration)\netstandard2.0\TestTasks.dll</TestTargetsAssembly>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MSBuildThisFileDirectory)..\test\TestTasks\TestTasks.csproj">
+       <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+       <BuildReference Condition="Exists('$(TestTargetsAssembly)')">False</BuildReference>
```
pakrym: This is the real change.

jkotalik:
@pakrym I'm doing some investigation on why AppVeyor tests are flaky right now.

build/testsite.props
jkotalik: Makes sense to me.

Issue #725, opened 2018-03-23
Weird AppDomain.CurrentDomain.BaseDirectory path in "InProcess" mode


Tornhoof:
If you run your application "inprocess" in 2.1-preview1-final, ``AppDomain.CurrentDomain.BaseDirectory`` return ``\\?\C:\InProcess`` which apparently breaks file loading in e.g. xmldocument. In my Code there is a File.Exist before that, which works, but then it fails in the doc.Load.

In outofprocess mode it still returns ``c:\InProcess``.

Repro:
run the following snippet in an inprocess web application and copy a test.xml into the folder
```csharp
var fileName = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "test.xml");
var doc = new XmlDocument();
doc.Load(fileName);
```
Exception is:
```ini
System.UriFormatException
  HResult=0x80131537
  Message=Invalid URI: The hostname could not be parsed.
  Source=System.Private.Uri
  StackTrace:
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   at System.Uri..ctor(String uriString, UriKind uriKind)
   at System.Xml.XmlResolver.ResolveUri(Uri baseUri, String relativeUri)
   at System.Xml.XmlUrlResolver.ResolveUri(Uri baseUri, String relativeUri)
   at System.Xml.XmlTextReaderImpl..ctor(String url, XmlNameTable nt)
   at System.Xml.XmlDocument.Load(String filename)
```

Note: This might as well be some bug in corefx, but since it first showed up in the inprocess mode, I'll post it here first. I don't know this path format and apparently not every part of .net core can handle this.

Tratcher:
Sure, but is there a way to suppress the `\\?\` prefix?

jkotalik:
Oh, we can remove the \\?\ portion.

Tornhoof:
Yeah, please remove the prefix, that's all I need I think, I did that manually and at least my app started and loaded the files.
I still have some issues I need to figure out, but I'm one step further now.

jkotalik:
@pan-wang

pan-wang:
the prefix is for  extended-length path and UNC. https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx.

muratg:
This is tracking the core-setup bug that @davidfowl linked above.

Tornhoof:
[...]

PR #736, opened 2018-03-28
Do not canonicalize short paths

pakrym:


src/IISLib/util.cxx
```diff
@@ -61,13 +61,17 @@ Return Values:
             return hr;
         }
     }
-    else
+    else if (wcslen(pszName) > MAX_PATH)
```
jkotalik: I feel like it would be better to check this in https://github.com/aspnet/IISIntegration/blob/18799f6d4847dbad82efce25657cada221fa309f/src/CommonLib/utility.cxx#L427.
Check:
```c++
if (wcslen(pszFullPath) > MAX_PATH)
{
    hr = MakePathCanonicalizationProof( pszFullPath, pStruFullPath );
}
```

pakrym: I don't want to touch network path (`\\share\bla`) logic at all and that would affect it.

jkotalik: Kk. @pan-wang this will be modifying IISLib code.

jkotalik:
I am okay with this change, however @pan-wang should have final say.

jkotalik:
#725

jkotalik:
https://github.com/aspnet/KestrelHttpServer/issues/2440

PR #738, opened 2018-03-28
out of process fix

pan-wang:
fix an AV in case user give ASPNETCORE_PORT env as we double dereferenced
avoid retry and log if the env set invalid port value
create jobobject in retry as we delete it in cleanup


jkotalik:
Can you add a test where we set the ASPNETCORE_PORT in the environment variable section and verify it sets the right port/doesn't AV?

pan-wang:
I verified manually. Will add a test case as another PR.

jkotalik:
I don't see why it would be thatttt hard to add a test for this.

pan-wang:
Simple PR,  easy to review

pan-wang:
Test will be added tomorrow. Another pr

src/RequestHandler/outofprocess/serverprocess.cxx
```diff
@@ -887,6 +905,12 @@ SERVER_PROCESS::StartProcess(
         goto Finished;
 
     Failure:
+        if (fCriticalError)
+        {
+            // Critical error, no retry need to avoid wasting resource and polluting log
+            dwRetryCount = 0;
```
jkotalik: Is there any other failure cases we would consider "critical" here? If we are going to pass in a bool for setting up the port, should we pass it in to other functions?

PR #742, opened 2018-03-30
Check native HRESULTs

pakrym:


src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
```diff
@@ -7,22 +7,14 @@
 
 namespace Microsoft.AspNetCore.Server.IISIntegration
 {
-    internal class NativeMethods
+    internal static class NativeMethods
     {
-#if DOTNET5_4
```
jkotalik: @Tratcher this was here for out of process. Is it safe to remove now?

src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs
```diff
@@ -117,7 +117,7 @@ public class IISMiddleware
                 // IIS already limits this, no need to do it twice.
                 bodySizeFeature.MaxRequestBodySize = null;
             }
-
+
```
jkotalik: bruh

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
```diff
@@ -37,85 +29,212 @@ public enum REQUEST_NOTIFICATION_STATUS
...
+        }
+
+        public static unsafe int HttpWebsocketsWriteBytes(IntPtr pInProcessHandler, HttpApiTypes.HTTP_DATA_CHUNK* pDataChunks, int nChunks, PFN_WEBSOCKET_ASYNC_COMPLETION pfnCompletionCallback, IntPtr pvCompletionContext, out bool fCompletionExpected)
+        {
+            return http_websockets_write_bytes(pInProcessHandler, pDataChunks, nChunks, pfnCompletionCallback, pvCompletionContext, out fCompletionExpected);
+        }
+
+        public static void HttpEnableWebsockets(IntPtr pInProcessHandler)
+        {
+            Validate(http_enable_websockets(pInProcessHandler));
```
pan-wang: if someone call this methods on Win7, I think there will an AV. the native method  in \src\RequestHandler\managedexports.cxx should do null check and condition check

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
```diff
@@ -37,85 +29,212 @@ public enum REQUEST_NOTIFICATION_STATUS
...
-        public static extern int http_get_server_variable(IntPtr pInProcessHandler, [MarshalAs(UnmanagedType.AnsiBStr)] string variableName, [MarshalAs(UnmanagedType.BStr)] out string value);
+        private static extern int http_get_server_variable(IntPtr pInProcessHandler, [MarshalAs(UnmanagedType.AnsiBStr)] string variableName, [MarshalAs(UnmanagedType.BStr)] out string value);
 
         [DllImport(AspNetCoreModuleDll)]
-        public unsafe static extern bool http_shutdown();
+        private static extern bool http_shutdown();
 
         [DllImport(AspNetCoreModuleDll)]
-        public unsafe static extern int http_websockets_read_bytes(IntPtr pInProcessHandler, byte* pvBuffer, int cbBuffer, PFN_WEBSOCKET_ASYNC_COMPLETION pfnCompletionCallback, IntPtr pvCompletionContext, out int dwBytesReceived, out bool fCompletionExpected);
+        private static extern unsafe int http_websockets_read_bytes(IntPtr pInProcessHandler, byte* pvBuffer, int cbBuffer, PFN_WEBSOCKET_ASYNC_COMPLETION pfnCompletionCallback, IntPtr pvCompletionContext, out int dwBytesReceived, out bool fCompletionExpected);
```
jkotalik: Nit: can you reformat the websocket method decls to multiple lines.

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
```diff
@@ -37,85 +29,212 @@ public enum REQUEST_NOTIFICATION_STATUS
...
         public static bool is_ancm_loaded()
         {
             return GetModuleHandle(AspNetCoreModuleDll) != IntPtr.Zero;
         }
+
+        private static void Validate(int hr)
+        {
+            if (hr != 0)
+            {
+                throw Marshal.GetExceptionForHR(hr);
```
jkotalik: We should wrap these exceptions somewhere with a custom exception because most of these will be ugly COM exceptions.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -225,7 +225,7 @@ private unsafe IISAwaitable WriteToIISAsync(ReadOnlySequence<byte> buffer)
                     chunk.DataChunkType = HttpApiTypes.HTTP_DATA_CHUNK_TYPE.HttpDataChunkFromMemory;
                     chunk.fromMemory.pBuffer = (IntPtr)pBuffer;
                     chunk.fromMemory.BufferLength = (uint)buffer.Length;
-                    hr = NativeMethods.http_write_response_bytes(_pInProcessHandler, pDataChunks, nChunks, out fCompletionExpected);
+                    NativeMethods.HttpWriteResponseBytes(_pInProcessHandler, pDataChunks, nChunks, out fCompletionExpected);
```
jkotalik: check hr.

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
```diff
@@ -37,85 +29,212 @@ public enum REQUEST_NOTIFICATION_STATUS
...
+        }
+
+        public static bool HttpTryGetServerVariable(IntPtr pInProcessHandler, string variableName, out string value)
+        {
+            return http_get_server_variable(pInProcessHandler, variableName, out value) == 0;
+        }
+
+        public static bool HttpShutdown()
+        {
+            return http_shutdown();
```
pan-wang: could not find this http_shutdown() native  method

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
Tratcher: yes, we never built that. It was for nano (?)

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
jkotalik: We can remove it; I wanted to be liberal with the native methods exposed as I didn't want to break the API.

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
pakrym: We can't really add more information to it, HRESULT and stack trace is all we have.

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
pakrym: Can you please file a separate issue for that

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -191,7 +191,7 @@ public IISHttpContext CreateHttpContext(IntPtr pInProcessHandler)
         ~IISHttpServer()
         {
             // If this finalize is invoked, try our best to block all calls into managed.
-            NativeMethods.http_stop_calls_into_managed();
+            NativeMethods.HttpStopCallsIntoManaged();
```
davidfowl: This can't throw right?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -145,11 +145,13 @@ internal unsafe IISHttpContext(MemoryPool<byte> memoryPool, IntPtr pInProcessHan
                 // server variables a few extra times if a bunch of requests hit the server at the same time.
                 if (_websocketAvailability == WebsocketAvailabilityStatus.Uninitialized)
                 {
-                    NativeMethods.http_get_server_variable(pInProcessHandler, WebSocketVersionString, out var webSocketsSupported);
-                    var webSocketsAvailable = !string.IsNullOrEmpty(webSocketsSupported);
-                    _websocketAvailability = webSocketsAvailable ?
-                        WebsocketAvailabilityStatus.Available :
-                        WebsocketAvailabilityStatus.NotAvailable;
+                    if (NativeMethods.HttpTryGetServerVariable(pInProcessHandler, WebSocketVersionString, out var webSocketsSupported))
```
davidfowl: Shouldn't this set _websocketAvailability to something even if this fails?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
pakrym: It doesn't return HRESULT and does no validation.

pakrym:
The broken test is real, will investigate.

jkotalik:
I would check reading/cancel_io as the point of failures for those. Probably something to do with going async?

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
```diff
@@ -207,7 +223,8 @@ public static void HttpEnableWebsockets(IntPtr pInProcessHandler)
 
         public static void HttpCancelIO(IntPtr pInProcessHandler)
         {
-            Validate(http_cancel_io(pInProcessHandler));
+            //TODO: Validation is disabled here because it randomly returns 0x80070490 Element not found
+            http_cancel_io(pInProcessHandler);
```
jkotalik: @pakrym my guess is this returns an HRESULT if there is nothing to cancel.

pakrym: Aren't we keeping track of when we are reading?

jkotalik: Yes, but we don't know what the native state of reading is. If a read completed in native and is propagating up to managed, in managed we still think the reading is pending. If we call CancelIo, it returns a failure saying there is nothing to cancel. The read will still complete correctly though.

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
jkotalik: Don't think this is an issue, we will never call this method if websockets are disabled. We can check in native but it seems redundant.

src/Microsoft.AspNetCore.Server.IISIntegration/NativeMethods.cs
pakrym: True, I'll verify and add it back then

src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs
```diff
@@ -43,6 +43,7 @@ public static IWebHostBuilder UseIISIntegration(this IWebHostBuilder hostBuilder
             // Check if in process
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && NativeMethods.IsAspNetCoreModuleLoaded())
             {
+
```
davidfowl: nit: space

PR #752, opened 2018-04-03
Explore usage of std::mutex and CHandle

pakrym:
This is exploratory PR which tries to explore improving our error handling/resource management patterns in native code. Main goal is to lower entry level and modernize the codebase.

Basic ideas:
1. Use RAII when possible (std::mutex, CHandle are two examples used in this PR) - provides easier, less error prone resource management and greatly simplifies logic by allowing to just `return` out of methods instead of doing goto finalization section.
2. Catch all exceptions on the border between shim and inprocess handler to prevent them from leaking up the stack.

Next possible steps:
Use exceptions for flowing HRESULTS - a lot of our methods just receive hresult check if it's failed and return it up the stack - this is exactly what exceptions are created for.

Easier to review with ?w=1 enabled:` https://github.com/aspnet/IISIntegration/pull/752?w=1`

### DIFF
--- a/src/CommonLib/SRWLockWrapper.cpp
+++ b/src/CommonLib/SRWLockWrapper.cpp
@@ -1,0 +1,13 @@
+#include "stdafx.h"
+#include "SRWLockWrapper.h"
+
+SRWLockWrapper::SRWLockWrapper(const SRWLOCK& lock)
+    : m_lock(lock)
+{
+    AcquireSRWLockExclusive(const_cast<SRWLOCK*>(&m_lock));
+}
+
+SRWLockWrapper::~SRWLockWrapper()
+{
+    ReleaseSRWLockExclusive(const_cast<SRWLOCK*>(&m_lock));
+}
--- a/src/RequestHandler/inprocess/inprocessapplication.cpp
+++ b/src/RequestHandler/inprocess/inprocessapplication.cpp
@@ -6,6 +6,7 @@ IN_PROCESS_APPLICATION::IN_PROCESS_APPLICATION(
     IHttpServer*        pHttpServer,
     ASPNETCORE_CONFIG*  pConfig) :
     APPLICATION(pHttpServer, pConfig),
+    m_pHttpServer(pHttpServer),
     m_ProcessExitCode(0),
     m_hLogFileHandle(INVALID_HANDLE_VALUE),
     m_hErrReadPipe(INVALID_HANDLE_VALUE),
@@ -15,7 +16,8 @@ IN_PROCESS_APPLICATION::IN_PROCESS_APPLICATION(
     m_fBlockCallbacksIntoManaged(FALSE),
     m_fInitialized(FALSE),
     m_fShutdownCalledFromNative(FALSE),
-    m_fShutdownCalledFromManaged(FALSE)
+    m_fShutdownCalledFromManaged(FALSE),
+    m_srwLock()
 {
     // is it guaranteed that we have already checked app offline at this point?
     // If so, I don't think there is much to do here.
@@ -57,8 +59,8 @@ IN_PROCESS_APPLICATION::ShutDown(
     VOID
 )
 {
-    HANDLE   hThread = NULL;
-    HRESULT  hr = S_OK;
+    HRESULT hr = S_OK;
+    CHandle  hThread;
     DWORD    dwThreadStatus = 0;
     DWORD    dwTimeout = m_pConfig->QueryShutdownTimeLimitInMS();
 
@@ -67,15 +69,15 @@ IN_PROCESS_APPLICATION::ShutDown(
         dwTimeout = INFINITE;
     }
 
-    hThread = CreateThread(
+    hThread.Attach(CreateThread(
         NULL,       // default security attributes
         0,          // default stack size
         (LPTHREAD_START_ROUTINE)DoShutDown,
         this,       // thread function arguments
         0,          // default creation flags
-        NULL);      // receive thread identifier
+        NULL));      // receive thread identifier
 
-    if (hThread == NULL)
+    if ((HANDLE)hThread == NULL)
     {
         hr = HRESULT_FROM_WIN32(GetLastError());
         goto Finished;
@@ -94,12 +96,6 @@ IN_PROCESS_APPLICATION::ShutDown(
     }
 
 Finished:
-
-    if (hThread != NULL)
-    {
-        CloseHandle(hThread);
-    }
-    m_hThread = NULL;
 
     if (FAILED(hr))
     {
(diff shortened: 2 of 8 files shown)


pan-wang:
IIS OOB code is supposed to run on multiple version of Windows and cannot assume the C++ runtime redistribution package is installed, so it historically uses windows APIs only instead of some std package. Switch to std, you may want to static link all libs and this make the assembly much bigger.

Tratcher:
Except the C++ Redist is an install prereq?

pakrym:
@pan-wang where can I find this guideline?

pakrym:
https://osgwiki.com/wiki/WIL_Error_Handling_Helpers interesting read about handling exceptions on ABI interfaces.

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -57,49 +56,54 @@ IN_PROCESS_APPLICATION::ShutDown(
...
+            this,       // thread function arguments
+            0,          // default creation flags
+            NULL));      // receive thread identifier
 
-    if (hThread == NULL)
-    {
-        hr = HRESULT_FROM_WIN32(GetLastError());
-        goto Finished;
-    }
+        if ((HANDLE)hThread == NULL)
```
jkotalik: I think this should have been checking for ERR_INVALID_HANDLE to being with...

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -57,49 +56,54 @@ IN_PROCESS_APPLICATION::ShutDown(
...
-
-    if (hThread != NULL)
+    catch (const std::system_error& ex)
     {
-        CloseHandle(hThread);
+        hr = ex.code().value();
+    }
+    catch (...)
+    {
+        hr = ERROR_EXCEPTION_IN_SERVICE;
```
jkotalik: Anything that starts with ERROR_* needs to be wrapped HRESULT_FROM_WIN32(). One good macro we have XD

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -132,70 +136,77 @@ IN_PROCESS_APPLICATION::ShutDownInternal()
...
     HANDLE   handle = NULL;
     WIN32_FIND_DATA fileData;
-    BOOL     fLocked = FALSE;
 
-    if (IsDebuggerPresent())
+    try
     {
-        dwTimeout = INFINITE;
-    }
+        if (IsDebuggerPresent())
```
jkotalik: Move up to the lock out to the try?

src/RequestHandler/inprocess/inprocesshandler.cpp
```diff
@@ -20,7 +20,19 @@ IN_PROCESS_HANDLER::OnExecuteRequestHandler()
...
+    {
+        hr = ((IN_PROCESS_APPLICATION*)m_pApplication)->LoadManagedApplication();
+    }
+    catch (const std::system_error& ex)
+    {
+        hr = ex.code().value();
+    }
+    catch (...)
+    {
+        hr = ERROR_EXCEPTION_IN_SERVICE;
```
jkotalik: HRESULT_FROM_WIN32.

pakrym:
I'm going to undo exception part of this and make a custom RAII wrapped for mutex.

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -227,58 +238,49 @@ IN_PROCESS_APPLICATION::ShutDownInternal()
...
+        // multiple times can lead to AVs. 
+        if (m_fRecycleCalled)
+        {
+            return;
+        }
 
-    AcquireSRWLockExclusive(&m_srwLock);
-    fLockAcquired = TRUE;
+        {
+            std::unique_lock<std::mutex> lock(m_srwLock);
```
pan-wang: this is the part I personally don't prefer. The code depends on the destructor of unique_lock to do unlock  once the object is out of scope. This makes the code no that easy to be read.

pakrym: Yeah, but C++ guarantees that it would run vs a person having to remember to add `Finished` label and undo it there.

pan-wang: my understanding is that SWRLOCK is faster than C++ unique_lock as later requires more resource.
Also even performance is not a concern, I would prefer doing unlock specifically for easy code readability and maintenance. The rule is that optimize code for the reader instead of the writer.  For instance, it is not easy to understand why IISExpress scenario is not in else  section for if (!m_pHttpServer->IsCommandLineLaunch())

pakrym: I removed unique_lock replacing it with SWR wrapper. I'll try improve readability.

pakrym: I rewrote the method a bit, `m_pHttpServer->RecycleProcess` call doesn't require lock because there is nothing else calling method on it and first part would prevent double call.
Notice that we never actually set `m_fRecycleCalled` value to true before so the code had a bug.

pakrym: I wan't us to use RAII in cases where resource is being aquired and returned in a scope of one method, it would greatly simplify the logic and allows to use returns instead goto's and labels.

src/CommonLib/SRWLockWrapper.cpp
```diff
@@ -0,0 +1,13 @@
+#include "stdafx.h"
+#include "SRWLockWrapper.h"
+
+SRWLockWrapper::SRWLockWrapper(const SRWLOCK& lock)
+    : m_lock(lock)
+{
+    AcquireSRWLockExclusive(const_cast<SRWLOCK*>(&m_lock));
```
jkotalik: It seems to be excessive to be downcasting to SRWLOCK just so the call to SRWLockWrapper accepts const. I would drop it here.

pan-wang: what's the benefit of this wrapper? it introduce additional overhead by just wrapping Windows API. Why should we waste CPU?

pakrym: The benefit is that you don't need to think about releasing resources in methods and is a widely used practice in C++ (https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) that simplifies code.

The overhead of this wrapper is minimal if any at all, compiler knows now to inline constructor and method calls so I'm pretty sure generated code wouldn't change at all:

image

pan-wang: I ran  AcquireSRWLockExclusive() &  ReleaseSRWLockExclusive() vs. SRWLockWrapper lock(srwLock) million times, the first takes 15 ticks while the other takes  140 ticks.

pan-wang: after applying that, the perf is same. Compiler does put the method inline :)

jkotalik:
🚢 🇮🇹

PR #756, opened 2018-04-03
[WIP] Create abstraction to call into WindowsApis

jkotalik:
I would like to create an abstraction for calling into windows apis for testing. To do this, we need to create a wrapper class and destaticfy each call, as we need to override the implementation with a mock. As a PoC, I did this by making the HOSTFXR_UTILITY use a few methods from a WindowsFileApi class, which will call ReadFile, WriteFile, etc. The "mock" class itself doesn't do anything, as we should probably switch to gTest before writing proper tests.

Few questions with this:
- Should we put all windows apis into the same class?
- Should all utility functions eventually be non-static such that we can override their implementation?

Don't nit style or anything yet pls.


pan-wang:
I am not sure about the goal of this change. This will make the readability worse and increase the maintenance cost for the future.

test/AspNetCoreModuleTests/stdafx.h
```diff
@@ -51,6 +51,8 @@
 #include "..\..\src\CommonLib\debugutil.h"
 #include "..\..\src\CommonLib\requesthandler.h"
 #include "..\..\src\CommonLib\resources.h"
+#include "WindowsFileApiMock.h"
```
pakrym: Is it the right ordering?

jkotalik:
The main goal is allow us to mock windows calls without actually doing them. For example, if I want call a function which calls CreateFile, but not actually create a file on disk, I can change the implementation of the windows call to mock it and do some expected behavior.

pakrym:
I wouldn't start by abstracting out win32 calls but instead would've done higher level functional entities: `where runner`, `dotnet exe finder`, `env variable expander` etc.
So we can test higher level logic like that we try to search in right places for right mode on ANCM or that we try to replace environment variables.

When we would have that we can abstract win32 calls from under those implementations and test they do expected things with the environment.

PR #759, opened 2018-04-04
Chunk writes to file

jkotalik:
Apparently it is a bad idea to allocate 200MB in mem at the same time <_<. Fixes some AppVeyor failures.

### DIFF
--- a/test/WebSites/InProcessWebSite/Startup.cs
+++ b/test/WebSites/InProcessWebSite/Startup.cs
@@ -652,7 +652,15 @@ namespace IISTestSite
             app.Run(async ctx =>
             {
                 var tempFile = Path.GetTempFileName();
-                File.WriteAllText(tempFile, new string('a', 200000000));
+                var fileContent = new string('a', 200000);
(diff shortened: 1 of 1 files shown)


pakrym:
Does it repro the bug it was supposed to test?

jkotalik:
No, it was returning System.OutOfMemoryException due to allocating the string.

pakrym:
What I'm saying is that we wrote this test for a bug with stackoverflow when trying to stackalloc a lot of segments for writing. Changing the test to chunking might stop reproducing that bug. (https://github.com/aspnet/IISIntegration/commit/cf4874997eec8ce11a66b03856772737dec5a4c4)

jkotalik:
I see. I'll do some investigation.

jkotalik:
The bug will still reproduce because the SendFileAsync still has the same size for the file. There was no issue with calling write on a large string.

pakrym:
Oh, it's about reading the file. But does this mean that we are still loading the entire file into memory just not twice?

jkotalik:
Yep.

PR #763, opened 2018-04-06
Fixing couple issues

pan-wang:
1) memory leak of file watcher entry in case of the configuration changed and then shutdown. the configuration changed and then shutdown. 
2) in out of process, the port in use check may fail due to a race condition. That is we first query the table size, allocate a buffer and then query again to retrieve the data. If there is another process opens a listening port simultaneously, the table size acquired from previous query is out of data and leads to a smaller buffer. This may cause process start failure. Call GetExitCodeProcess again to make sure the process really exit as we sometime get false value. 
3) adding more log for out of process scenario, i.e., the id of the process which is listening on the assigned port, and the event of process shutdown.


src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -563,6 +564,11 @@ APPLICATION_INFO::RecycleApplication()
                 g_pHttpServer->RecycleProcess(L"On Demand by AspNetCore Module for recycle application failure");
             }
         }
+        else
```
pakrym: Use CThread? :)

src/RequestHandler/outofprocess/serverprocess.cxx
```diff
@@ -528,15 +528,20 @@ SERVER_PROCESS::PostStartCheck(
...
-                hr = E_APPLICATION_ACTIVATION_EXEC_FAILURE;
-                strEventMsg.SafeSnwprintf(
-                    ASPNETCORE_EVENT_PROCESS_START_STATUS_ERROR_MSG,
-                    m_struAppFullPath.QueryStr(),
-                    m_struPhysicalPath.QueryStr(),
-                    m_struCommandLine.QueryStr(),
-                    hr,
-                    processStatus);
-                goto Finished;
+                // double check
```
pakrym: Why? Do we expect it to change in the millisecond since last check?

pan-wang: as I stated, sometime the call returned 0, while the process is still alive per the log. Double check to avoid false alert. This issue was found in VS team test.

src/AspNetCore/Src/applicationinfo.cpp
pan-wang: will Cthread  be heavy due to constructor and destructor?

src/RequestHandler/outofprocess/serverprocess.cxx
```diff
@@ -1222,6 +1236,12 @@ SERVER_PROCESS::CheckIfServerIsUp(
         if (iResult == SOCKET_ERROR)
         {
             hr = HRESULT_FROM_WIN32(WSAGetLastError());
+            if (hr == HRESULT_FROM_WIN32(WSAECONNREFUSED))
```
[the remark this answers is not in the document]
pan-wang: WSAECONNREFUSED is winNT error code not hresult. We need to introduce another viable to hold the value. Not sure which one has better perf. In the past, we just did this pattern

src/RequestHandler/outofprocess/serverprocess.cxx
pakrym: Do we know that second would return correct value?

src/AspNetCore/Src/applicationinfo.cpp
pakrym: Yes

src/AspNetCore/Inc/applicationmanager.h
```diff
@@ -79,18 +79,18 @@ class APPLICATION_MANAGER
 
     ~APPLICATION_MANAGER()
     {
+        if (m_pFileWatcher != NULL)
```
jkotalik: Why was this moved up?

src/AspNetCore/Src/applicationinfo.cpp
jkotalik: I am fine with using Cthread on any non-request code path. Recycling the application is per application, and I prefer the benefits of using CThread s.t. we don't need to close the handle.

src/AspNetCore/Src/filewatcher.cxx
```diff
@@ -380,15 +379,17 @@ FILE_WATCHER_ENTRY::StopMonitor(VOID)
     //
     InterlockedExchange(&_lStopMonitorCalled, 1);
 
-    AcquireSRWLockExclusive(&_srwLock);
-
     if (_hDirectory != INVALID_HANDLE_VALUE)
     {
-        CloseHandle(_hDirectory);
-        _hDirectory = INVALID_HANDLE_VALUE;
+        AcquireSRWLockExclusive(&_srwLock);
```
jkotalik: Once @pakrym PR is merged, we should use the SRWLockWrapper here.

src/RequestHandler/outofprocess/serverprocess.cxx
```diff
@@ -1159,6 +1170,9 @@ SERVER_PROCESS::CheckIfServerIsUp(
             goto Finished;
         }
 
+        // More entries may be created between the two GetExtendedTcpTable calls
+        // Give a bigger buffer, 200 byte is big enough to hold additional entries
+        dwSize += 200;
```
jkotalik: I don't like the approach here. Why can't we just call GetExtendedTcpTable with an initial size for the pTCPInfo, and keep expanding it until the call succeeds?

src/AspNetCore/Src/applicationinfo.cpp
pan-wang: I really want to do some performance test to see the perf impact before using standard libraries. NGinx is purely based on C and has much better performance.

src/RequestHandler/outofprocess/serverprocess.cxx
pan-wang: We don't know. We can only do double check. With those change, VS tests seems stable now.

src/AspNetCore/Inc/applicationmanager.h
pan-wang: Stop file watcher first is safer trying to avoid the race that FCN comes to a shutting down applicaiton

src/RequestHandler/outofprocess/serverprocess.cxx
pan-wang: we don't want to be in the situation of infinite loop (multiple tries) if processes continue open ports.

jkotalik: 1. We can limit the number of retries we do to a fixed number.
2. Can processes opening ports out grow like a 2x buffer size increase? I doubt it.

pakrym: Yeah, let's avoid the magic const and do the loop

pan-wang: You don't want waste CPU cycles on those retries. Why not give a bigger enough buffer ？

pan-wang: You don't want waste CPU cycles on those retries. Why not give a bigger enough buffer ？

pakrym: Make a retry count 2, just write it as a loop and double the buffer size. I would be much cleaner.

jkotalik: Because this can still fail.  I'd rather guarantee reliability here rather than worry about why a process didn't start randomly.

pakrym: 200 would only fit 10 entries. `MIB_TCPROW` is 20 bytes.

src/RequestHandler/outofprocess/serverprocess.cxx
```diff
@@ -1123,36 +1134,36 @@ SERVER_PROCESS::CheckIfServerIsUp(
...
-
-        if (dwResult != NO_ERROR && dwResult != ERROR_INSUFFICIENT_BUFFER)
+        while (dwResult == ERROR_INSUFFICIENT_BUFFER)
         {
-            hr = HRESULT_FROM_WIN32(dwResult);
-            goto Finished;
-        }
+            // Increase the buffer size with additional space, MIB_TCPROW 20 bytes
+            // New entries may be added by other processes before calling GetExtendedTcpTable
+            dwSize += 200;
```
jkotalik: make this *2 instead of + 200.

src/RequestHandler/outofprocess/serverprocess.cxx
```diff
@@ -1123,36 +1134,36 @@ SERVER_PROCESS::CheckIfServerIsUp(
...
+            }
+
+            dwResult = GetExtendedTcpTable(pTCPInfo,
+                &dwSize,
+                FALSE,
+                AF_INET,
+                TCP_TABLE_OWNER_PID_LISTENER,
+                0);
+
+            if (dwResult != NO_ERROR && dwResult != ERROR_INSUFFICIENT_BUFFER)
```
jkotalik: HeapFree at the end of the method if there is a failure? Also why not have the HeapFree at the end of the loop?

pan-wang: it's safer as when retry happens we make sure the previous resource gets released first to avoid memory leak. if doing this at the end of loop, I have to do another condition check.

src/RequestHandler/outofprocess/serverprocess.cxx
pan-wang: double the size is too aggressive. on a regular box with couple IIS sites, the tcptable size is only about 600-800 bytes. Claiming a larger buffer will waste resource.

src/CommonLib/utility.cxx
```diff
@@ -642,7 +642,7 @@ UTILITY::LogEventF(
 
     STACK_STRU ( strEventMsg, 256 );
 
-    if (SUCCEEDED(strEventMsg.SafeSnwprintf(
+    if (SUCCEEDED(strEventMsg.SafeVsnwprintf(
```
jkotalik: Use LogEventF? And why was this changed?

pan-wang: ISafeSnwPrintf internally use argslist. If we call SafeSnwprints and pass in va_list, va_list will be treated as one input and values will not be resolved correctly. We need call SafeVsnwprintf which accepts va_list. See our STRU implementation.

src/AspNetCore/Src/applicationinfo.cpp
pan-wang: @pakrym please open another working issue so that we can switch to CTHREAD cross the projects. I don't want partially switch to CThread in this change.

PR #764, opened 2018-04-09
Cleanup LogEvent calls

pakrym:
Remove duplicated logic

### DIFF
--- a/src/AspNetCore/Src/applicationinfo.cpp
+++ b/src/AspNetCore/Src/applicationinfo.cpp
@@ -112,7 +112,7 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
         // if it was, log that app_offline has been dropped.
         if (m_fAppOfflineFound)
         {
-            UTILITY::LogEventF(g_hEventLog,
+            UTILITY::LogEvent(g_hEventLog,
                 EVENTLOG_INFORMATION_TYPE,
                 ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_REMOVED,
                 ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_REMOVED_MSG);
@@ -258,7 +258,7 @@ APPLICATION_INFO::FindRequestHandlerAssembly()
         {
             if (FAILED(hr = FindNativeAssemblyFromHostfxr(&struFileName)))
             {
-                UTILITY::LogEventF(g_hEventLog,
+                UTILITY::LogEvent(g_hEventLog,
                         EVENTLOG_INFORMATION_TYPE,
                         ASPNETCORE_EVENT_INPROCESS_RH_MISSING,
                         ASPNETCORE_EVENT_INPROCESS_RH_MISSING_MSG);
@@ -268,7 +268,7 @@ APPLICATION_INFO::FindRequestHandlerAssembly()
         {
             if (FAILED(hr = FindNativeAssemblyFromGlobalLocation(&struFileName)))
             {
-                UTILITY::LogEventF(g_hEventLog,
+                UTILITY::LogEvent(g_hEventLog,
                     EVENTLOG_INFORMATION_TYPE,
                     ASPNETCORE_EVENT_OUT_OF_PROCESS_RH_MISSING,
                     ASPNETCORE_EVENT_OUT_OF_PROCESS_RH_MISSING_MSG);
--- a/src/AspNetCore/Src/applicationinfo.cpp
+++ b/src/AspNetCore/Src/applicationinfo.cpp
@@ -15,7 +15,7 @@ APPLICATION_INFO::~APPLICATION_INFO()
     {
         // Mark the entry as invalid,
         // StopMonitor will close the file handle and trigger a FCN
-        // the entry will delete itself when processing this FCN 
+        // the entry will delete itself when processing this FCN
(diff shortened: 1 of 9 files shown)


src/CommonLib/utility.cxx
```diff
@@ -627,3 +627,30 @@ UTILITY::LogEvent(
         fwprintf(stderr, L"ERROR: %s\n", pstrMsg);
     }
 }
+
+VOID
+UTILITY::LogEventF(
```
jkotalik: LogFailureEvent? I'm not familiar with the F suffix.

src/CommonLib/utility.cxx
```diff
@@ -627,3 +627,30 @@ UTILITY::LogEvent(
...
+    _In_ WORD    dwEventInfoType,
+    _In_ DWORD   dwEventId,
+    _In_ LPCWSTR pstrMsg,
+    ...
+)
+{
+    STACK_STRU ( strEventMsg, 256 );
+
+    va_list argsList;
+    va_start(argsList, pstrMsg);
```
jkotalik: Comment saying "va_start needs to know the last argument of the function call is to get the arguments.

src/CommonLib/utility.cxx
pakrym: [F]ormat? Like in printf

jkotalik: Ahh. Fine with me then.

src/CommonLib/utility.cxx
pan-wang: I like this LogEventF. It avoids duplicated code.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -112,15 +112,10 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
...
-            {
-                UTILITY::LogEvent(g_hEventLog,
-                    EVENTLOG_INFORMATION_TYPE,
-                    ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_REMOVED,
-                    strEventMsg.QueryStr());
-            }
+            UTILITY::LogEventF(g_hEventLog,
+                EVENTLOG_INFORMATION_TYPE,
+                ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_REMOVED,
+                ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_REMOVED_MSG);
```
pan-wang: for this one, please try to add the application info to the log. if there are lots applications, it will be hard to tell which one's app_offlinefile was removed based on the log.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -268,32 +258,20 @@ APPLICATION_INFO::FindRequestHandlerAssembly()
...
-                {
-                    UTILITY::LogEvent(g_hEventLog,
+                UTILITY::LogEventF(g_hEventLog,
                         EVENTLOG_INFORMATION_TYPE,
                         ASPNETCORE_EVENT_INPROCESS_RH_MISSING,
-                        strEventMsg.QueryStr());
-                }
-
-                goto Finished;
+                        ASPNETCORE_EVENT_INPROCESS_RH_MISSING_MSG);
```
pan-wang: please add application info.

src/AspNetCore/Src/applicationinfo.cpp
pakrym: I'm not changing messages themselves in this PR. Just refactoring the LogEvent usages. We can file it and do it the future. I'd like to keep this mechanical.

src/CommonLib/utility.cxx
pan-wang: one optimization is to avoid calling strEventMsg.SafeSnwprintf (to avoid build stru and string copy) if less than 4 input parameters

src/AspNetCore/Src/applicationinfo.cpp
pan-wang: In such case, we should call LogEvent instead of LogEventF to avoid creating stru and string copy

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -268,32 +258,20 @@ APPLICATION_INFO::FindRequestHandlerAssembly()
...
-                {
-                    UTILITY::LogEvent(g_hEventLog,
-                        EVENTLOG_INFORMATION_TYPE,
-                        ASPNETCORE_EVENT_OUT_OF_PROCESS_RH_MISSING,
-                        strEventMsg.QueryStr());
-                }
+                UTILITY::LogEventF(g_hEventLog,
+                    EVENTLOG_INFORMATION_TYPE,
+                    ASPNETCORE_EVENT_OUT_OF_PROCESS_RH_MISSING,
+                    ASPNETCORE_EVENT_OUT_OF_PROCESS_RH_MISSING_MSG);
```
pan-wang: should we add more info to this error message?

src/AspNetCore/Src/applicationinfo.cpp
pakrym: I'll change LogEventF to short circuit on empty args.

pakrym: Turns out it's hard, changed to LogEvent

PR #767, opened 2018-04-09
Gain some perf

pakrym:
Goes from 133k to 154k

Check bytes left before reading or IIS would try to read from network.

Also lazy features, a bit less tasks.


src/RequestHandler/managedexports.cxx
```diff
@@ -206,18 +206,26 @@ http_read_request_bytes(
...
-
-    hr = pHttpRequest->ReadEntityBody(
-        pvBuffer,
-        dwCbBuffer,
-        fAsync,
-        pdwBytesReceived,
-        pfCompletionPending);
 
-    if (hr == HRESULT_FROM_WIN32(ERROR_HANDLE_EOF))
+    if (pHttpRequest->GetRemainingEntityBytes() > 0)
```
jkotalik: How does this affect read performance? Is GetRemainingEntityBytes inexpensive when there is data available? Doing this on every read could be costly.

pakrym: Shouldn't affect read performance it returns 0 for GET requests and some value for Content-Lenght, for everything else `INFINITE` is returned.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -122,11 +144,12 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleRequest(IntPtr pI
             var server = (IISHttpServer)GCHandle.FromIntPtr(pvRequestContext).Target;
             Interlocked.Increment(ref server._outstandingRequests);
 
-            var context = server._iisContextFactory.CreateHttpContext(pInProcessHandler);
-
-            var task = Task.Run(() => context.ProcessRequestAsync());
-
-            task.ContinueWith((t, state) => CompleteRequest((IISHttpContext)state, t), context);
+            _ = Task.Run(
```
jkotalik: @davidfowl did we want to put this back on the IIS Request thread?

src/RequestHandler/managedexports.cxx
```diff
@@ -206,18 +206,26 @@ http_read_request_bytes(
...
+            pvBuffer,
+            dwCbBuffer,
+            fAsync,
+            pdwBytesReceived,
+            pfCompletionPending);
+
+        if (hr == HRESULT_FROM_WIN32(ERROR_HANDLE_EOF))
+        {
+            // We reached the end of the data
+            hr = S_OK;
```
jkotalik: Set hr = S_OK initially so we don't need the else. Also, don't we need to set pdwBytesReceived to 0 and pfCompletionPending to False?

pakrym: Done

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
davidfowl: No closures...

pakrym: It looks better and taking ctor off iis thread gains ~4k. I say lets take it and iterate more later.

davidfowl: Get rid of the closure and use QUWI?

pakrym: There is a task created for state machine anyway

pakrym: My bad, you are right. Will change.

davidfowl: But you can avoid the per request allocation....

pakrym: Also, cross compile for nca2.1 and compile out inproc from netstandard

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpServer.cs
```diff
@@ -144,16 +144,25 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleRequest(IntPtr pI
...
-                });
-
+#if NETCOREAPP2_1
+            ThreadPool.QueueUserWorkItem<(IISHttpServer, IntPtr)>(
+                state => _ = HandleRequest(state.Item1, state.Item2),
+                (server, pInProcessHandler),
+                preferLocal: false);
+#else
+            ThreadPool.QueueUserWorkItem(
+                state => _ = HandleRequest(server, pInProcessHandler));
```
davidfowl: But why?

pakrym:
@davidfowl any other concerns?

PR #776, opened 2018-04-13
reset some timeout values to infinite if debugger is attached

pan-wang:
for VS debugging scenario


src/AspNetCoreModuleV2/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -1138,11 +1139,16 @@ FORWARDING_HANDLER::CreateWinHttpRequest(
         goto Finished;
     }
 
+    if (!pServerProcess->IsDebuggerAttached())
```
jkotalik: Port this to V1?

pan-wang: yes. I also need port couple of other changes to V1. will create another PR for porting.

pakrym:
Why are we using CheckRemoteDebuggerPresent instead of IsDebuggerPresent ?

> The "remote" in CheckRemoteDebuggerPresent does not imply that the debugger necessarily resides on a different computer; instead, it indicates that the debugger resides in a separate and parallel process. Use the IsDebuggerPresent function to detect whether the calling process is running under the debugger.

Issue #775, opened 2018-04-13
ANCM SxS schema files 


pan-wang:
AS part of ANCM SxS, different version of schema files may get dropped.  I did some manual tests 
1)	Dropping  another version of ANCM schema file with different name but same content. Asp.Net core application still works.
2)	Dropping another version of  ANCM schema file with reduced content (no matter the file name). Asp.Net core application still works
3)	Dropping another version of ANCM schema file with conflicted elements. Asp.Net core application may fail. 

**Conclusion:** IIS DOES merge the schema files and tolerate duplications. In the case of conflicted elements, the order of files (it seems to be alphabet order) does matter. The last one wins.

**Suggestion:** Rename ANCM V2 schema file as aspnetcoreV2_schema.xml

**FollowUp**: Official validation once the MSIs are available

rockerinthelocker:
Why not simply using a single schema file? Since IIS merges schema files, the outcome would be the same.

pan-wang:
@rockerinthelocker  Schema file does not have version control. Considering the scenario that user has V2 installed which has some new schema elements first and then install V1, the schema file will be replaced by V1
s and cause issue for V2 runtime. Similar uninstallation could also be problematic.

rockerinthelocker:
@pan-wang , Well, only the existing v1 installer is not smart enough and customers (server administrators) simply repair the latest post-v1 ANCM module, or restore the latest schema file from backup, to resolve any such issue caused by the v1 installer. Going forward, post-v1 installers should simply not replace the schema file on install/repair if a newer ANCM version is installed. On uninstall, post-v1 installers should remove the schema file only if there's no other ANCM module installed. Leaving the latest schema file in place works because ANCM modules ignore unknown settings and IIS doesn't care either. Agreed, the schema file must be growing-only, newer ANCM versions cannot remove settings (adding a comment to settings no longer supported should be good to go). In a perfect world... but this approach is certainly better than confusing anyone with a number of ANCM schema files that have almost the same content anyway -- and server administrators will applaude you.

shhsu:
@pan-wang 
IIS schema processing ultimately calls method: FindNextFileA (https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-findnextfilea) which does not assume any order.

pakrym:
@shirhatti did we decide to update V1 schema to whatever we have in V2?

PR #777, opened 2018-04-13
Rename aspnetcore_schema -> aspnetcoreV2_schema

jkotalik:
For #775


pakrym:
Do we need to change any nuget packages/archives/targets/scripts etc.?

jkotalik:
Going to make this change as part of splitting the installers completely.

Issue #780, opened 2018-04-18
w3wp.exe crashed when rename app_offline.htm on Win10 X86 machine.


Mgoul:
There are many w3wp.exe unhanding exception when running test on X86 machine. I check the event logs, there are many Application error about W3WP.

Machine build: 10.0.17133.1.x86fre.rs4_release.180323-1312 X86 
Error messages:
     Faulting application name: w3wp.exe, version: 10.0.17109.1000, time stamp: 0xd6e38087
Faulting module name: ntdll.dll, version: 10.0.17109.1000, time stamp: 0x9cc51555
Exception code: 0xc0000005

Repro steps:
1. Build AspNetCore application with cmd “dotnet publish D:\gitroot\AspNetCoreModule\test\AspNetCoreModule.TestSites.Standard --framework netcoreapp2.0”
2. Convert the test AspNetCore application to web application under Default Web Site
3. Create an 1App_offline.htm under AspNetCoreApp folder.
4. Send a request to AspNetCoreApp
5. Rename 1App_offline.htm to App_offline.htm, the crash issue repro.

This crash was caused below AV: 
(bc4.3c4): Access violation - code c0000005 (first chance)
First chance exceptions are reported before any exception handling.
This exception may be expected and handled.
eax=00000000 ebx=0082a310 ecx=77ade062 edx=00000000 esi=d4022ba9 edi=00000000
eip=77b39863 esp=031bf6e0 ebp=031bf6f4 iopl=0         nv up ei pl nz na po nc
cs=001b  ss=0023  ds=0023  es=0023  fs=003b  gs=0000             efl=00010202
ntdll!RtlCaptureStackContext+0x13e23:
77b39863 ff773c          push    dword ptr [edi+3Ch]  ds:0023:0000003c=????????
0:010> kn
ChildEBP RetAddr  
WARNING: Stack unwind information not available. Following frames may be wrong.
00 031bf6f4 77adfb72 ntdll!RtlCaptureStackContext+0x13e23
01 031bf710 77acaf7a ntdll!TpSetWaitEx+0x382
02 031bf8d8 75b2a1a4 ntdll!TpCallbackIndependent+0x6ca
03 031bf8ec 77aa343e KERNEL32!BaseThreadInitThunk+0x24
04 031bf934 77aa3412 ntdll!PssNtCaptureSnapshot+0x79e
05 031bf944 00000000 ntdll!PssNtCaptureSnapshot+0x77

muratg:
[...]

PR #783, opened 2018-04-18
Add LPTHREAD_START_ROUTINE due to compiler change (fixing issue #780)

pan-wang:


pan-wang:
fixing issue #780

pan-wang:
@pakrym  https://msdn.microsoft.com/en-us/library/windows/desktop/ms682453(v=vs.85).aspx requires the third input parameter is LPTHREAD_START_ROUTINE. Compilers seem tolerating the missing of the force conversion. However, the binary generated by the new compiler works fine only for x64. For x86,  the function address cannot be resolved correctly at runtime and causes AV.

pakrym:
It's suspicious that we have to do it and also that thread functions are not static like they should be (see https://stackoverflow.com/questions/1372967/how-do-you-use-createthread-for-functions-which-are-class-members?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)

Can you also fix https://github.com/aspnet/IISIntegration/blob/589a672521727c84bd33d136292da05501a93bcf/src/AspNetCoreModuleV2/RequestHandler/inprocess/inprocessapplication.cpp#L75

https://msdn.microsoft.com/en-us/library/windows/desktop/ms686736(v=vs.85).aspx

states 
> Do not declare this callback function with a void return type and cast the function pointer to LPTHREAD_START_ROUTINE when creating the thread. Code that does this is common, but it can crash on 64-bit Windows.

pakrym (approved):

PR #792, opened 2018-04-24
Make exchange classes pure virtual

pakrym:
We shouldn't have non-virtual members as part of our ABI


src/AspNetCoreModuleV2/CommonLib/application.h
```diff
@@ -30,24 +26,23 @@ class APPLICATION
     Recycle() = 0;
 
     virtual
-    ~APPLICATION();
+    ~APPLICATION() = 0 { };
```
pan-wang: why not have IAPPLICATION pure abstract class and then have APPLICATION::IAPPLICATION as abstract class which can avoid duplicate code at in-process and out-of-process sub classes. Same for REQEUST_HANDLER

src/AspNetCoreModuleV2/RequestHandler/dllmain.cxx
```diff
@@ -303,7 +303,7 @@ CreateApplication(
             goto Finished;
         }
 
-        pApplication = new OUT_OF_PROCESS_APPLICATION(pServer, pConfig);
```
pan-wang: we passed pServer to out-of-process application and in-process application as we want to have the ability to read the configuration inside the class instead of using the passed the configuration instance., e.g., requesthandler  needs to read configuration setting from handlerSettings

src/AspNetCoreModuleV2/CommonLib/application.h
pakrym: There isn't too much code duplication, I would like to avoid base classes for things that ship with different schedule and live in different repos because it increases probability of us accidentaly breaking things.

src/AspNetCoreModuleV2/RequestHandler/dllmain.cxx
pakrym: We can still do it, but there is no point in keeping unused parameter around until then

pakrym:
I also moved CreateHandler to APPLICATION and removed `QueryConfig` from it.

src/AspNetCoreModuleV2/CommonLib/application.h
bariscaglar: I agree with Pan here. If there's a strong reason to define an interface, you can do so with an IAPPLICATION. You should still have a base class so you don't have to implement all the methods in all derived classes.

pakrym: Common part is 3 members two of which are ref counting which we don't try to extract into common type anywhere (https://github.com/aspnet/IISIntegration/search?utf8=%E2%9C%93&q=cRefs+%3D+InterlockedDecrement%28%26m_cRefs%29%29&type=), and the third one is a trivial field access. 

We are additionally planning to split current code base into two repositories that would be built and released at a different time, I don't want accidental changes to a base class to break the product.

pakrym:
No, by assigning I meant you are the main reviewer and have the last say.

pakrym:
I'll add a base class today.

src/AspNetCoreModuleV2/AspNetCore/src/precomp.hxx
```diff
@@ -104,7 +104,7 @@ inline bool IsSpace(char ch)
 #include "..\..\CommonLib\environmentvariablehash.h"
 #include "..\..\CommonLib\aspnetcoreconfig.h"
 #include "..\..\CommonLib\hostfxr_utility.h"
-#include "..\..\CommonLib\application.h"
+#include "..\..\CommonLib\IAPPLICATION.h"
```
jkotalik: Nit: lowercase.

src/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
```diff
@@ -171,24 +171,24 @@
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClInclude Include="application.h" />
+    <ClInclude Include="IAPPLICATION.h" />
```
jkotalik: Nit: lowercase.

PR #793, opened 2018-04-24
React to splitting memory pool

pakrym:
https://github.com/aspnet/Home/issues/3040


pakrym:
Oops, accidental merge.

PR #794, opened 2018-04-24
Honor DotNetAssetRootUrl when installing .NET Core x86 runtimes

natemcmaster:
Required to make 2.1 builds pass.


natemcmaster:
Found a simpler solution, so never mind.

PR #811, opened 2018-05-03
Split ANCM into two dlls.

jkotalik:


pakrym:
Make sure it's green on CIs before merging.

Issue #812, opened 2018-05-03
Reduce configuration parsing in shim.


jkotalik:
Instead of parsing the entire aspnetcore_config in the shim, we should parse only the configuration necessary to the shim and then parse the rest of the configuration in the request handler.

PR #814, opened 2018-05-04
Split config and hostfxr parsing between shim and request handler.

jkotalik:
For #812


src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
```diff
@@ -0,0 +1,175 @@
...
+    CONST
+    PCWSTR
+    QueryHostFxrFullPath(
+        VOID
+    )
+    {
+        return m_struHostFxrLocation.QueryStr();
+    }
+
+        VOID
```
pakrym: nit: formatting

src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
```diff
@@ -0,0 +1,175 @@
...
+    SetHostFxrArguments(
+    DWORD dwArgc,
```
pakrym: nit: formatting

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
```diff
@@ -10,7 +10,7 @@ typedef
 HRESULT
 (WINAPI * PFN_ASPNETCORE_CREATE_APPLICATION)(
     _In_  IHttpServer        *pServer,
-    _In_  ASPNETCORE_CONFIG  *pConfig,
+    _In_  IHttpContext       *pHttpContext,
```
pakrym: Doesn't seem right, isn't `IHttpContext` per request? Why would app need it in addition to `IHttpServer`?

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
```diff
@@ -0,0 +1,260 @@
...
+    STRU                            strApplicationFullPath;
+    IAppHostAdminManager           *pAdminManager = NULL;
+    IAppHostElement                *pAspNetCoreElement = NULL;
+    DWORD                           dwCounter = 0;
+    DWORD                           dwPosition = 0;
+    WCHAR*                          pszPath = NULL;
+    BSTR                            bstrAspNetCoreSection = NULL;
+
+    pAdminManager = pHttpServer->GetAdminManager();
+    hr = m_struConfigPath.Copy(pHttpContext->GetApplication()->GetAppConfigPath());
```
pakrym: Take application in instead of context.

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    hr = m_struApplicationPhysicalPath.Copy(pHttpContext->GetApplication()->GetApplicationPhysicalPath());
+    if (FAILED(hr))
+    {
+        goto Finished;
+    }
+
+    pszPath = m_struConfigPath.QueryStr();
+    while (pszPath[dwPosition] != NULL)
```
pakrym: What is this?

src/AspNetCoreModuleV2/CommonLib/iapplication.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+class IREQUEST_HANDLER;
```
pakrym: How did it work before?

src/AspNetCoreModuleV2/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -995,12 +1001,17 @@ IN_PROCESS_APPLICATION::QueryConfig() const
...
 IN_PROCESS_APPLICATION::CreateHandler(
     _In_  IHttpContext       *pHttpContext,
-    _In_  HTTP_MODULE_ID     *pModuleId,
     _Out_ IREQUEST_HANDLER   **pRequestHandler)
 {
     HRESULT hr = S_OK;
     IREQUEST_HANDLER* pHandler = NULL;
-    pHandler = new IN_PROCESS_HANDLER(pHttpContext, pModuleId, this);
+
+    if (FAILED(hr))
```
pakrym: Hm?

src/AspNetCoreModuleV2/RequestHandler/dllmain.cxx
```diff
@@ -276,16 +276,19 @@ HRESULT
...
 )
 {
     HRESULT      hr = S_OK;
     IAPPLICATION *pApplication = NULL;
+    REQUESTHANDLER_CONFIG *pConfig = NULL;
 
     // Initialze some global variables here
     InitializeGlobalConfiguration(pServer);
 
+    hr = REQUESTHANDLER_CONFIG::CreateRequestHandlerConfig(pServer, pHttpContext, g_hEventLog, &pConfig);
```
pakrym: Check it.

src/AspNetCoreModuleV2/RequestHandler/inprocess/inprocessapplication.h
```diff
@@ -10,10 +10,15 @@ typedef REQUEST_NOTIFICATION_STATUS(WINAPI * PFN_MANAGED_CONTEXT_HANDLER)(void *
 class IN_PROCESS_APPLICATION : public APPLICATION
 {
 public:
-    IN_PROCESS_APPLICATION(IHttpServer* pHttpServer, ASPNETCORE_CONFIG* pConfig);
+    IN_PROCESS_APPLICATION(IHttpServer* pHttpServer, REQUESTHANDLER_CONFIG *pConfig);
 
     ~IN_PROCESS_APPLICATION();
 
+	HRESULT
```
pakrym: nit: Formatting

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
pan-wang: you need IHttpServer to be able to read configuration. You also need IIS application info which is available at httpcontext. I agree it looks wired to pass httpcontext directly. We should only pass necessary info

src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
```diff
@@ -0,0 +1,175 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
```
pan-wang: I personally don't the name aspnetcore_shim_config. As you already renamed another one to request_handler_config, you can keep current one

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
jkotalik: I'll need to modify my websocket PR to set whether websockets are enabled later on as I use the HttpContext for it. Otherwise, I can change the IHttpContext to IHttpApplication.

src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
jkotalik: I prefer including shim in the name. Keeps it separate from the old aspnetcore_config and makes it clear that the config is smaller.

src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
```diff
@@ -0,0 +1,175 @@
...
+    STRU*
+    QueryArguments(
+        VOID
+    )
+    {
+        return &m_struArguments;
+    }
+
+    HRESULT
+    SetHostFxrFullPath(
```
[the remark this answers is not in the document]
jkotalik: I'll wait for your PR to change that.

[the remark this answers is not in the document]
jkotalik: For now I'll remove some hostfxr stuff from the request handler configuration.

src/AspNetCoreModuleV2/RequestHandler/requesthandler_config.cpp
```diff
@@ -2,10 +2,10 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 #include "stdafx.h"
-#include "aspnetcoreconfig.h"
+#include "requesthandler_config.h"
```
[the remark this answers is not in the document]
jkotalik: I'd rather keep the configurations completely separate. The shim configuration is fairly small already. The amount of duplication is minimal.

jkotalik:
🆙 📅

pan-wang:
could you please measure the delay introduced by the two configuration readings

jkotalik:
Performance change is worse than I thought. I'll do a bit of investigation for what I can affect versus what Pan will be changing with hostfxr stuff. I'll run a perf view and see where the bottlenecks are.
```
           Method |     Mean |    Error |   StdDev |  Op/s | Allocated |
----------------- |---------:|---------:|---------:|------:|----------:|
Before config change
 SendFirstRequest | 913.4 ms | 170.4 ms | 112.7 ms | 1.095 |   5.45 KB |
 SendFirstRequest | 863.4 ms | 105.6 ms | 69.88 ms | 1.158 |       0 B |
 SendFirstRequest | 845.0 ms | 69.47 ms | 45.95 ms | 1.183 |   5.45 KB |
After Config change
 SendFirstRequest | 992.1 ms | 115.8 ms | 76.59 ms | 1.008 |   5.45 KB |
 SendFirstRequest | 963.4 ms | 105.2 ms | 69.55 ms | 1.038 |   5.45 KB |
 SendFirstRequest | 999.2 ms | 127.5 ms | 84.32 ms | 1.001 |   5.45 KB |
```

jkotalik:
The main perf regression was due to finding dotnet twice, which calls where.exe. I went ahead and refactored it locally and the perf is the same as in dev, but hesitate to push to this branch as it will cause a much larger diff.

src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
```diff
@@ -0,0 +1,175 @@
...
+class ASPNETCORE_SHIM_CONFIG : IHttpStoredContext
+{
+public:
+    virtual
+    ~ASPNETCORE_SHIM_CONFIG();
+
+    static
+    HRESULT
+    GetConfig(
+        _In_  IHttpServer             *pHttpServer,
```
pan-wang: I think we can get ride of this method as now shim_config is owned by applicationinfo object. Can be done in next PR

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
```diff
@@ -0,0 +1,259 @@
...
+    DBG_ASSERT(m_cRefs != 0);
+    LONG cRefs = 0;
+    if ((cRefs = InterlockedDecrement(&m_cRefs)) == 0)
+    {
+        delete this;
+    }
+}
+
+HRESULT
+ASPNETCORE_SHIM_CONFIG::GetConfig(
```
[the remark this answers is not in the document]
jkotalik: What benefits do you see in removing it?

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
pan-wang: apppath (pszPath) is not used by shim-config and can be removed.

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+    else
+    {
+        DebugPrintf(ASPNETCORE_DEBUG_FLAG_INFO,
+            "ASPNETCORE_SHIM_CONFIG::GetConfig, set config to ModuleContext");
+
+        // set appliction info here instead of inside Populate()
+        // as the destructor will delete the backend process
+        hr = pAspNetCoreShimConfig->QueryApplicationPath()->Copy(pHttpApplication->GetApplicationId());
```
pan-wang: i don't think we need this

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationmanager.h
```diff
@@ -65,7 +65,7 @@ class APPLICATION_MANAGER
     HRESULT
     GetOrCreateApplicationInfo(
```
pan-wang: we can pass ihttpcontext here. It makes sense to get the application for a request. Can be done in next PR

src/AspNetCoreModuleV2/CommonLib/iapplication.h
jkotalik: Not sure, but I get compilation errors saying the request handler cannot be found.

jkotalik: Maybe ordering of includes in the stdafx.h file. I'll mess around with that.

src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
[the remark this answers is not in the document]
jkotalik: This pr now has the hostfxr changes.

src/AspNetCoreModuleV2/RequestHandler/dllmain.cxx
```diff
@@ -21,7 +21,8 @@ IHttpServer *       g_pHttpServer = NULL;
 HINSTANCE           g_hWinHttpModule;
 HINSTANCE           g_hAspNetCoreModule;
 HANDLE              g_hEventLog = NULL;
-
+PCSTR               g_szDebugLabel = "ASPNET_CORE_MODULE_REQUEST_HANDLER";
+std::map<std::wstring, std::wstring> requestHandlerConfigurationMap;
```
pakrym: Don't use the dictionary, compare name and stick value into a variable.

src/AspNetCoreModuleV2/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -16,15 +16,15 @@ IN_PROCESS_APPLICATION::IN_PROCESS_APPLICATION(
...
+    m_srwLock()
 {
     // is it guaranteed that we have already checked app offline at this point?
     // If so, I don't think there is much to do here.
     DBG_ASSERT(pHttpServer != NULL);
     DBG_ASSERT(pConfig != NULL);
-    InitializeSRWLock(&m_srwLock);
 
+    InitializeSRWLock(&m_srwLock);
+    m_pConfig = pConfig;
```
pakrym: Did this have to go under the lock?

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
```diff
@@ -9,9 +9,16 @@ extern BOOL     g_fRecycleProcessCalled;
 typedef
 HRESULT
 (WINAPI * PFN_ASPNETCORE_CREATE_APPLICATION)(
-    _In_  IHttpServer        *pServer,
-    _In_  ASPNETCORE_CONFIG  *pConfig,
-    _Out_ IAPPLICATION       **pApplication
+    _In_  IHttpServer    *pServer,
+    _In_  IHttpContext   *pHttpContext,
```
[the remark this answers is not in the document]
jkotalik: Yes. I'll add it as part of the SetKeyValue api so we can drop it.

src/AspNetCoreModuleV2/RequestHandler/dllmain.cxx
jkotalik: 😢

src/AspNetCoreModuleV2/RequestHandler/outofprocess/winhttphelper.cxx
```diff
@@ -26,7 +26,7 @@ WINHTTP_HELPER::StaticInitialize(
 {
     HRESULT hr = S_OK;
 
-    if (!g_fWebSocketSupported)
+    if (g_fWebSocketSupported != WebSocketStatus::WEBSOCKET_SUPPORTED)
```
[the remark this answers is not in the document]
jkotalik: I'm removing the websocket check from OutOfProcess for now. May revisit in another PR.

src/AspNetCoreModuleV2/RequestHandler/inprocess/inprocesshandler.cpp
```diff
@@ -123,6 +124,24 @@ IN_PROCESS_HANDLER::SetAsyncCompletionStatus(
...
+VOID
+IN_PROCESS_HANDLER::SetWebsocketSupportedVariable(
+    VOID
+)
+{
+    PCSTR pzWebsocketVar = NULL;
+    DWORD dwVersionLength;
+    HRESULT hr;
+
+    if (g_fWebSocketSupported == WebSocketStatus::WEBSOCKET_UNKNOWN)
```
pan-wang: for naming conversion, g_XX means global variable. As you check websocket per application, it should use g_xxx

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -221,8 +224,13 @@ APPLICATION_INFO::EnsureApplicationCreated()
...
                 goto Finished;
             }
+            hr = m_pfnAspNetCoreSetKeyValue(L"ExeLocation", exeLocation.c_str());
+            if (FAILED(hr))
+            {
+                goto Finished;
+            }
 
-            hr = m_pfnAspNetCoreCreateApplication(m_pServer, m_pConfiguration, &pApplication);
+            hr = m_pfnAspNetCoreCreateApplication(m_pServer, pHttpContext->GetApplication(), &pApplication);
```
pan-wang: why not pass in the exe location string here so that you don't need to introduce another API?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -314,6 +322,14 @@ APPLICATION_INFO::FindRequestHandlerAssembly()
             hr = HRESULT_FROM_WIN32(GetLastError());
             goto Finished;
         }
+
```
pan-wang: I don't think we need this API

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        if (FAILED(hr = pAspNetCoreShimConfig->SetHostFxrFullPath(struHostFxrDllLocation.QueryStr())))
+        {
+            goto Finished;
+        }
+
+        pAspNetCoreShimConfig->SetHostFxrArguments(dwArgCount, pwzArgv);
+
+        *exeLocation = std::wstring(struExeAbsolutePath.QueryStr());
```
pan-wang: avoid using stdlib here as you duplicate the string

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    hr = pAspNetCoreShimConfig->Populate(pHttpServer, pHttpApplication);
+    if (FAILED(hr))
+    {
+        goto Finished;
+    }
+
+    // Modify Inprocess specific configuration here. 
+    if (pAspNetCoreShimConfig->QueryHostingModel() == APP_HOSTING_MODEL::HOSTING_IN_PROCESS)
```
pan-wang: this piece of code seems wired here as it changes the configuration. It either should be part of ShimConfig::Populate or other method where use the HostFxr

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
```diff
@@ -145,7 +137,10 @@ class APPLICATION_INFO
     ShutDownApplication();
 
     HRESULT
-    EnsureApplicationCreated();
+    EnsureApplicationCreated(
+        IHttpContext   *pHttpContext,
+        std::wstring    exeLocation
```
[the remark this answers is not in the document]
jkotalik: Nope. I'll drop it for now so we so we don't need to debate it.

src/AspNetCoreModuleV2/CommonLib/iapplication.h
```diff
@@ -42,6 +44,20 @@ class IAPPLICATION
...
+HRESULT
+(WINAPI * PFN_ASPNETCORE_CREATE_APPLICATION)(
+    _In_  IHttpServer    *pServer,
+    _In_  IHttpApplication   *pHttpContext,
+    _Out_ IAPPLICATION  **pApplication
+    );
+
+typedef
+HRESULT
+(WINAPI * PFN_ASPNETCORE_SET_KEY_VALUE)(
```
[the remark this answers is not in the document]
jkotalik: Dropping this.

src/AspNetCoreModuleV2/RequestHandler/inprocess/inprocesshandler.cpp
jkotalik: I don't quite understand the difference you are specifying.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
jkotalik: We don't want to have the executable path be a required parameter for creating the application. I'm adding a SetProperties function to the IApplication, which takes in arbitrary KVPs to be set on the application.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
jkotalik: Removed.

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
```diff
@@ -0,0 +1,240 @@
...
+    DBG_ASSERT(m_cRefs != 0);
+    LONG cRefs = 0;
+    if ((cRefs = InterlockedDecrement(&m_cRefs)) == 0)
+    {
+        delete this;
+    }
+}
+
+HRESULT
+ASPNETCORE_SHIM_CONFIG::GetConfig(
```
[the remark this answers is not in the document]
jkotalik: I'm still using the module context for now. I'd prefer to change that in a separate PR as this has gotten too large.

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
jkotalik: I'd prefer to not store state on the config at all from hostfxr. I can do this in a separate PR though.

jkotalik:
Few things that will be cleaned up in future PRs:
- Make ASPNETCORE_SHIM_CONFIG belong to the application info rather than module container
- Remove setting hostfxr configuration in the aspnetcore_shim_config or requesthandler_config.
- Remove IHttpContext and ExePath from CreateApplication Api once #839 is complete. This will move config parsing to another stage.

src/AspNetCoreModuleV2/AspNetCore/src/dllmain.cpp
```diff
@@ -20,7 +20,7 @@ SRWLOCK             g_srwLock;
 DWORD               g_dwDebugFlags = 0;
 PCSTR               g_szDebugLabel = "ASPNET_CORE_MODULE";
 PCWSTR              g_pwzAspnetcoreRequestHandlerName = L"aspnetcorerh.dll";
-PFN_ASPNETCORE_CREATE_APPLICATION      g_pfnAspNetCoreCreateApplication;
```
pakrym: Undo

src/AspNetCoreModuleV2/RequestHandler/outofprocess/outprocessapplication.cpp
```diff
@@ -101,7 +99,7 @@ OUT_OF_PROCESS_APPLICATION::CreateHandler(
     //add websocket check here
     SetWebsocketStatus(pHttpContext);
 
-    pHandler = new FORWARDING_HANDLER(pHttpContext, m_fOutOfProcessWebSocketSupported, this);
+    pHandler = new FORWARDING_HANDLER(pHttpContext, m_fWebSocketSupported, this);
```
pan-wang: you can tell whether websocket is supported based on pHttpContext, you don't need to pass in m_fWebSocketSupported

jkotalik: That is inefficient. We would check if websockets are enabled every request with a server variable.

src/AspNetCoreModuleV2/RequestHandler/outofprocess/forwardinghandler.cpp
```diff
@@ -18,9 +18,9 @@ PROTOCOL_CONFIG             FORWARDING_HANDLER::sm_ProtocolConfig;
 RESPONSE_HEADER_HASH *      FORWARDING_HANDLER::sm_pResponseHeaderHash = NULL;
 
 FORWARDING_HANDLER::FORWARDING_HANDLER(
-    _In_ IHttpContext                   *pW3Context,
-    _In_  HTTP_MODULE_ID                *pModuleId,
-    _In_ OUT_OF_PROCESS_APPLICATION     *pApplication
+    _In_ IHttpContext                  *pW3Context,
+    _In_ BOOL                           fOutOfProcessWebSocketSupported,
```
pan-wang: the name can be as simple as fWebSocketSupported. No one cares about OutOfProcess here

jkotalik:
Prior test failures were due to flakiness in appveyor tests <_<. Reran and they passed.

pakrym:
What test are flaky?

jkotalik:
Like all tests on appveyor occasionally fail. It wasn't specific to the inprocess websocket tests. My only guess is that on AppVeyor, we need to update the schema file in the IISExpress directory, and there could be a race condition between the file update and running iisexpress.

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -142,15 +142,29 @@ HOSTFXR_UTILITY::GetStandaloneHostfxrParameters(
...
+HOSTFXR_UTILITY::IsDotnetExecutable(PCWSTR executablePath)
+{
+    STRU struDotnetExecutablePath;
+    if (FAILED(struDotnetExecutablePath.Copy(executablePath)))
+    {
+        return FALSE;
+    }
+    return struDotnetExecutablePath.EndsWith(L"dotnet.exe") || struDotnetExecutablePath.EndsWith(L"dotnet");
+}
+
```
pakrym: nit: newlines

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -182,7 +196,7 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
     }
 
     // Check if the absolute path is to dotnet or not.
-    if (struAbsolutePathToDotnet.EndsWith(L"dotnet.exe") || struAbsolutePathToDotnet.EndsWith(L"dotnet"))
+    if (HOSTFXR_UTILITY::IsDotnetExecutable(struAbsolutePathToDotnet.QueryStr()))
```
pakrym: You already have STRU why turn it into cstring and back?

src/AspNetCoreModuleV2/RequestHandler/inprocess/inprocessapplication.cpp
pakrym: then undo the change.

src/AspNetCoreModuleV2/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -807,7 +813,9 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
 
     DBG_ASSERT(m_status == APPLICATION_STATUS::STARTING);
 
-    hModule = LoadLibraryW(m_pConfig->QueryHostFxrFullPath());
+    // hostfxr should already be loaded by the shim. If not, then we will need
+    // to load it ourselves by finding hostfxr again.
+    hModule = LoadLibraryW(L"hostfxr.dll");
```
pakrym: const?

src/AspNetCoreModuleV2/RequestHandler/outofprocess/outprocessapplication.cpp
```diff
@@ -1,12 +1,13 @@
...
 OUT_OF_PROCESS_APPLICATION::OUT_OF_PROCESS_APPLICATION(
-    ASPNETCORE_CONFIG*  pConfig) :
-    m_pConfig(pConfig)
+    REQUESTHANDLER_CONFIG  *pConfig) :
+    m_fWebSocketSupported(WEBSOCKET_STATUS::WEBSOCKET_UNKNOWN)
 {
     m_status = APPLICATION_STATUS::RUNNING;
     m_pProcessManager = NULL;
-    InitializeSRWLock(&rwlock);
+    m_pConfig = pConfig;
```
pakrym: move back to initializers

src/AspNetCoreModuleV2/RequestHandler/outofprocess/winhttphelper.cxx
```diff
@@ -26,11 +26,6 @@ WINHTTP_HELPER::StaticInitialize(
 {
     HRESULT hr = S_OK;
 
-    if (!g_fWebSocketSupported)
```
pakrym: We were shortcitcuiting here, is removing it fine?

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
jkotalik: Mostly for the sake of a better api for outside functions calling IsDotnetExecutable. I'll change it though as all other calls use STRU.

src/AspNetCoreModuleV2/RequestHandler/outofprocess/winhttphelper.cxx
jkotalik: Reverting.

src/AspNetCoreModuleV2/RequestHandler/inprocess/inprocessapplication.cpp
jkotalik: Which portion do you want to be const? The string itself is const already, I can make HMODULE const too.

jkotalik:
🆙 📅

src/AspNetCoreModuleV2/CommonLib/iapplication.h
```diff
@@ -42,6 +45,14 @@ class IAPPLICATION
     HRESULT
     CreateHandler(
         _In_  IHttpContext       *pHttpContext,
-        _In_  HTTP_MODULE_ID     *pModuleId,
-        _Out_ IREQUEST_HANDLER   **pRequestHandler) = 0;
+        _Out_ IREQUEST_HANDLER  **pRequestHandler) = 0;
 };
+
+typedef
```
pakrym: This type is only required on the shim side.

src/AspNetCoreModuleV2/RequestHandler/outofprocess/outprocessapplication.cpp
```diff
@@ -99,3 +112,32 @@ OUT_OF_PROCESS_APPLICATION::CreateHandler(
     *pRequestHandler = pHandler;
     return hr;
 }
+
+VOID
+OUT_OF_PROCESS_APPLICATION::SetWebsocketStatus(
+    IHttpContext* pHttpContext
+)
+{
+    // TODO
```
pakrym: Is this TODO still true?

src/AspNetCoreModuleV2/RequestHandler/environmentvariablehelpers.h
```diff
@@ -367,7 +367,7 @@ class ENVIRONMENT_VAR_HELPERS
     AddWebsocketEnabledToEnvironmentVariables
     (
         _Inout_ ENVIRONMENT_VAR_HASH*       pInEnvironmentVarTable,
-        _In_ BOOL                           fWebsocketsEnabled
+        _In_ WEBSOCKET_STATUS               fWebsocketsEnabled
```
pakrym: Why pass WEBSOCKET_STATUS   instead of bool? You don't try to initialize it in this method anyway.

pakrym: And if it's not initialized at this point it's an error.

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -142,15 +142,26 @@ HOSTFXR_UTILITY::GetStandaloneHostfxrParameters(
...
 }
 
+BOOL
+HOSTFXR_UTILITY::IsDotnetExecutable(STRU *struExecutablePath)
+{
+    if (struExecutablePath == NULL)
+    {
+        return FALSE;
+    }
+    return struExecutablePath->EndsWith(L"dotnet.exe") || struExecutablePath->EndsWith(L"dotnet");
```
pan-wang: how about the scenarion of mydotnet.exe or mydotnet? We can fix it in next PR

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -224,14 +239,18 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
         //
         if (UTILITY::CheckIfFileExists(struAbsolutePathToDotnet.QueryStr()))
         {
-            hr = GetStandaloneHostfxrParameters(
+            hr = HOSTFXR_UTILITY::GetStandaloneHostfxrParameters(
```
pan-wang: need check the returned hr

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
jkotalik: Yeah next PR.

Issue #798, opened 2018-04-26
Websocket connection fails on in-process mode with SDK 2.1.300-preview3-008646 build 


MogulHuang:
Repro Steps:
1.	Connect the test machine, create the ANCM test aspnetcoreapp named aspnetcoreapprc
2.	Open IE and browse http://localhost/aspnetcoreapprc/chatplus.html
3.	Type ws://localhost/aspnetcorerc/websocket  onto “WebSocket Server URL:” text box
4.	Click “Connect” button

EXPECTED: Websocket connection succeeded.
ACTAUL:You will see it fails. 

NOTE:
This issue happens only with in-process mode. 
You will see there is no problem if you try the repro steps again after switching hostingModel from in-process to out-process with updating D:\inetpub\wwwroot\aspnetcoreapprc\web.config file and recycling w3wp.exe process.

Tratcher:
Can you share the app code?

jhkimnew:
Here you are . https://github.com/aspnet/AspNetCoreModule/tree/dev/test/AspNetCoreModule.TestSites.Standard

jhkimnew:
@shhsu Would you verify this? After that, our test should be updated with newer SDK version which contains this bug fix.

Tratcher:
The scheme is supposed to be http/s. ws/s is only used on the client side, it's not transmitted as part of the request. IsWebSocketRequest must be failing for another reason, such as incorrect headers.

Tratcher:
Can you capture a fiddler or wireshark trace of the request?

shhsu:
Turns out it's IIS configuration issue. sorry for the false alarm

PR #818, opened 2018-05-04
Extract a layer of IIS Async IO handling

pakrym:
The initial goal was to fix https://github.com/aspnet/IISIntegration/issues/798 which was broken by removing empty flush from the pipe.

I ended up extracting IIS IO specifics into two engines:
1. AsyncIOEngine - for normal message body - it allows one IO operation happening at a time
2. WebSocketsAsyncIOEngine - for upgraded requests

IO operations were extracted as seperate classes that allow pooling and implement IValueTaskSource.

I'll post benchmarks result after server comes online.


pakrym:
Ready for review

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncWriteOperationBase.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            if (_buffer.Length > int.MaxValue)
+            {
+                throw new InvalidOperationException($"Writes larger then {int.MaxValue} are not supported.");
+            }
+
+            bool completionExpected;
+            int hr;
+            var chunkCount = GetChunkCount();
+
```
jkotalik: nit space

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncWriteOperationBase.cs
```diff
@@ -0,0 +1,125 @@
...
+            if (_buffer.IsSingleSegment)
+            {
+                return 1;
+            }
+
+            var count = 0;
+
+            foreach (var _ in _buffer)
+            {
+                count++;
```
jkotalik: ReadOnlySequence should have a Chunk count api <_<

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/WebSocketsAsyncIOEngine.Initialize.cs
```diff
@@ -0,0 +1,51 @@
...
+            }
+
+            public void Initialize(IntPtr requestHandler)
+            {
+                _requestHandler = requestHandler;
+            }
+
+            public override bool InvokeOperation()
+            {
+                var hr = NativeMethods.HttpFlushResponseBytes(_requestHandler, out var fCompletionExpected);
```
jkotalik: This feels a bit overengineered for a single operation. Not against it, but it is a lot of code just to call HttpFlushResponseBytes.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/WebSocketsAsyncIOEngine.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            if (!_isInitialized)
+            {
+                throw new InvalidOperationException("Socket IO not initialized yet");
+            }
+        }
+
+        public void Stop()
+        {
+            // TODO
```
jkotalik: Make this throw too?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.Flush.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Server.IISIntegration
+{
+    internal partial class AsyncIOEngine
```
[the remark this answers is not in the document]
pakrym: And private member access (return operation), they are very tightly coupled.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
```diff
@@ -0,0 +1,164 @@
...
+                    // operation went async
+                    if (nextContinuation == null)
+                    {
+                        _runningOperation = next;
+                    }
+                }
+            }
+
+            continuation.Invoke();
+            nextContinuation?.Invoke();
```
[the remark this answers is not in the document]
pakrym: It's for a case when next completed synchronously so there would be no NotifyCompletion

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
```diff
@@ -0,0 +1,164 @@
...
+            Interlocked.Exchange(ref _cachedAsyncWriteOperation, null) ??
+            new AsyncWriteOperation(this);
+
+        private AsyncFlushOperation GetFlushOperation() =>
+            Interlocked.Exchange(ref _cachedAsyncFlushOperation, null) ??
+            new AsyncFlushOperation(this);
+
+        private void ReturnOperation(AsyncReadOperation operation)
+        {
+            Volatile.Write(ref _cachedAsyncReadOperation, operation);
```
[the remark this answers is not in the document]
pakrym: So the reference assignment is not reordered with resetting properties by CPU.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOOperation.cs
```diff
@@ -0,0 +1,143 @@
...
+                Continuation = continuation;
+                State = state;
+            }
+
+            public void Invoke()
+            {
+                if (Continuation != null)
+                {
+                    // TODO: use generic overload when code moved to be netcoreapp only
+                    var continuation = Continuation;
```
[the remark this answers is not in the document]
pakrym: You can not close around struct properties in lambda.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/WebSocketsAsyncIOEngine.Initialize.cs
pakrym: Kinda, but it still has to be awaitable and queuable

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncWriteOperationBase.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+
+        protected override void ResetOperation()
+        {
+            base.ResetOperation();
+
+            _requestHandler = default;
+            _buffer = default;
+            _handles.AsSpan().Clear();
```
davidfowl: Why AsSpan?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/IAsyncIOEngine.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Microsoft.AspNetCore.Server.IISIntegration
+{
+    internal interface IAsyncIOEngine
+    {
+        ValueTask<int> ReadAsync(Memory<byte> memory);
+        ValueTask<int> WriteAsync(ReadOnlySequence<byte> data);
+        ValueTask FlushAsync();
+        void NotifyCompletion(int hr, int bytes);
+        void Stop();
```
davidfowl: This one doesn't fit. Maybe Dispose?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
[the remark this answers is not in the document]
Tratcher: Concurrent reads and writes are not supported, correct? Should that be checked here or on Get?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        private void Run(AsyncIOOperation ioOperation)
+        {
+            lock (this)
+            {
+                if (_runningOperation != null)
+                {
+                    if (_nextOperation == null)
+                    {
+                        _nextOperation = ioOperation;
```
Tratcher: This appears to allow for two reads or two writes?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -22,7 +20,10 @@ internal partial class IISHttpContext
         /// <returns></returns>
         internal async Task<int> ReadAsync(Memory<byte> memory, CancellationToken cancellationToken)
         {
-            StartProcessingRequestAndResponseBody();
+            if (!_hasResponseStarted)
+            {
+                await InitializeResponseAwaited();
```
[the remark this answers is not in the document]
Tratcher: You're missing a test for the default scenario: Read the request body, set the status code and headers, write the response body.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -57,18 +58,13 @@ internal partial class IISHttpContext
...
         internal Task WriteAsync(ReadOnlyMemory<byte> memory, CancellationToken cancellationToken = default(CancellationToken))
         {
-
-            // Want to keep exceptions consistent,
-            if (!_hasResponseStarted)
+            async Task WriteFirstAsync()
             {
-                return WriteAsyncAwaited(memory, cancellationToken);
+                await InitializeResponseAwaited();
+                await Output.WriteAsync(memory, cancellationToken);
```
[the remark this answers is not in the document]
pakrym: No, it's a local method not a lambda.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
pakrym: It allows for one operation to be running and another queued. This is required so we can wait until current read cancellation notification would arrive and then pick up the next write.

test/WebSites/InProcessWebSite/Startup.cs
```diff
@@ -424,6 +445,110 @@ private void WebsocketRequest(IApplicationBuilder app)
...
+
+        private static async Task SendMessages(WebSocket webSocket, params string[] messages)
+        {
+            foreach (var message in messages)
+            {
+                await webSocket.SendAsync(new ArraySegment<byte>(Encoding.ASCII.GetBytes(message)), WebSocketMessageType.Text, true, CancellationToken.None);
+            }
+        }
+
+        private static async Task<WebSocket> Upgrade(HttpContext context)
```
jkotalik: Move websocket stuff into a separate class, such that it can be used by other tests in the future.

test/WebSites/InProcessWebSite/Startup.cs
```diff
@@ -381,6 +385,23 @@ private void ReadAndWriteEcho(IApplicationBuilder app)
             });
         }
 
+        private void ReadAndWriteEchoLines(IApplicationBuilder app)
+        {
+            app.Run(async context => {
```
jkotalik: few other potential tests (some may not be applicable)
- Write to response body, then upgrade the connection
- Flush after upgrade
- Reading and Writing concurrently with websockets (two separate tasks reading and writing).

test/IISIntegration.FunctionalTests/Utilities/TestConnections.cs
```diff
@@ -0,0 +1,212 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
```
jkotalik: Was this copied from Kestrel?

test/IISIntegration.FunctionalTests/Inprocess/WebSocketTests.cs
```diff
@@ -0,0 +1,71 @@
...
+    [Collection(IISTestSiteCollection.Name)]
+    public class WebSocketsTests
+    {
+        private readonly string _webSocketUri;
+
+        public WebSocketsTests(IISTestSiteFixture fixture)
+        {
+            _webSocketUri = fixture.BaseUri.Replace("http:", "ws:");
+        }
+
```
jkotalik: nit extra new lines.

test/IISIntegration.FunctionalTests/Inprocess/SynchronousReadAndWriteTests.cs
```diff
@@ -70,55 +67,73 @@ public SynchronousReadAndWriteTests(IISTestSiteFixture fixture)
...
+                foreach (var message in messages)
                 {
-                    stringBuilder.Append(Encoding.ASCII.GetString(buffer, 0, size));
+                    await connection.Send(message);
+                    await connection.Receive(message);
                 }
 
-                Assert.Contains(new StringBuilder().Insert(0, "hello world", 100).ToString(), stringBuilder.ToString());
+                await connection.Send("\r\n");
+                await connection.WaitForConnectionClose();
```
jkotalik: This test doesn't assert anything.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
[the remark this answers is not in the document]
pakrym: No, it's a cache that might race between return and aquire and that's fine.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.FeatureCollection.cs
```diff
@@ -250,16 +250,21 @@ void IHttpResponseFeature.OnCompleted(Func<object, Task> callback, object state)
             {
                 throw new InvalidOperationException("CoreStrings.UpgradeCannotBeCalledMultipleTimes");
             }
+
+            if (AsyncIO != null)
```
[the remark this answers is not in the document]
pakrym: Not anymore, I've split request and response starting.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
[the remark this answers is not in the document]
pakrym: Added.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -97,378 +91,94 @@ private void StartProcessingRequestAndResponseBody()
...
-                            // Always Advance the data pointer to the end of the buffer.
-                            Output.Reader.AdvanceTo(buffer.End);
-                        }
-                    }
+                    var read = await AsyncIO.ReadAsync(memory);
 
-                    // Check if there was an upgrade. If there is, we will replace the request and response bodies with
-                    // two seperate loops. These will still be using the same Input and Output pipes here.
-                    if (_upgradeTcs?.TrySetResult(null) == true)
+                    if (read == -1)
```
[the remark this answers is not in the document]
pakrym: Done.

test/IISIntegration.FunctionalTests/Utilities/TestConnections.cs
pakrym: Yep

test/IISIntegration.FunctionalTests/Inprocess/SynchronousReadAndWriteTests.cs
pakrym: Yes it does, `Receive` takes an argument and asserts that it received what expected.

test/WebSites/InProcessWebSite/Startup.cs
pakrym: I'd like to get this PR in, there are always more things we can test.

pakrym:
:up::date:

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.ReadWrite.cs
```diff
@@ -97,378 +91,94 @@ private void StartProcessingRequestAndResponseBody()
...
-                                await FlushToIISAsync();
-                            }
-                        }
-                        finally
-                        {
-                            // Always Advance the data pointer to the end of the buffer.
-                            Output.Reader.AdvanceTo(buffer.End);
-                        }
-                    }
+                    var read = await AsyncIO.ReadAsync(memory);
```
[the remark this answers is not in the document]
pakrym: We would hit pipe backpressure eventually.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.FeatureCollection.cs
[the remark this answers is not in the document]
Tratcher: So if I ever call request.Body.Read then Upgrade will fail? That's not a restriction imposed by the other servers. Read returns 0 for upgrade requests and does not affect the upgrade.

What happens if I call request.Body.Read after an upgrade? The request body should be disconnected from the upgrade stream and continue to return 0.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -204,46 +202,54 @@ internal IISHttpServer Server
...
+
+            EnsureIOInitialized();
+
+            await AsyncIO.FlushAsync();
+
+            _writeBodyTask = WriteBody();
         }
 
-        private async Task ProduceStart(bool appCompleted)
+        private void InitializeRequest()
```
Tratcher: InitializeRequestIO?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContextOfT.cs
```diff
@@ -80,18 +80,18 @@ public IISHttpContextOfT(MemoryPool<byte> memoryPool, IHttpApplication<TContext>
             }
             finally
             {
-                // The app is finished and there should be nobody writing to the response pipe
                 Output.Dispose();
 
-                // The app is finished and there should be nobody reading from the request pipe
-                Input.Reader.Complete();
```
Tratcher: Where did this go? How do you terminate the request body read loop early?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
```diff
@@ -0,0 +1,164 @@
...
+            lock (this)
+            {
+                if (_runningOperation is AsyncReadOperation && _nextOperation != null)
+                {
+                    NativeMethods.HttpTryCancelIO(_handler);
+                }
+            }
+        }
+
+        public ValueTask FlushAsync()
```
Tratcher: Is this flush the one you always call after write? Or is it the public one? Any need for a combined write & flush operation to avoid cancelling the read twice?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+
+        private void CancelPendingRead()
+        {
+            lock (this)
+            {
+                if (_runningOperation is AsyncReadOperation && _nextOperation != null)
+                {
+                    NativeMethods.HttpTryCancelIO(_handler);
```
Tratcher: Is there a possible race condition between the async read becoming the _runningOperation and it becoming cancellable natively? E.g. could you try to cancel it natively before it had started natively and then get blocked when it didn't actually cancel?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncWriteOperationBase.cs
```diff
@@ -0,0 +1,119 @@
...
+                var chunks = stackalloc HttpApiTypes.HTTP_DATA_CHUNK[chunkCount];
+                hr = WriteSequence(chunkCount, _buffer, chunks, out completionExpected);
+            }
+            else
+            {
+                // Otherwise allocate the chunks on the heap.
+                var chunks = new HttpApiTypes.HTTP_DATA_CHUNK[chunkCount];
+                fixed (HttpApiTypes.HTTP_DATA_CHUNK* pDataChunks = chunks)
+                {
+                    hr = WriteSequence(chunkCount, _buffer, pDataChunks, out completionExpected);
```
Tratcher: Is this ever async?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
pakrym: No, `_runningOperation`  get's assigned only when already started.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncWriteOperationBase.cs
pakrym: Yes, why wouldn't it be ?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.FeatureCollection.cs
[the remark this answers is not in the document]
pakrym: <> What happens if I call request.Body.Read after an upgrade? The request body should be disconnected from the upgrade stream and continue to return 0.

Same things as it did before this change. I'm not trying to fix all our issues in one PR.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncWriteOperationBase.cs
Tratcher: So how is `fixed` supposed to keep it pinned async?

pakrym: It doesn't. AFAIK chunks do not need to be async. We keep memory itself pinned in handles.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
pakrym: This is called by `ReadBody()` loop after application calls `stream.FlushAsync()`.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncWriteOperationBase.cs
```diff
@@ -0,0 +1,119 @@
...
+            bytes = bufferLength;
+            return !completionExpected;
+        }
+
+        public override void FreeOperationResources(int hr, int bytes)
+        {
+            // Free the handles
+            foreach (var handle in _handles)
+            {
+                handle.Dispose();
```
jkotalik: Based on the above feedback, we should look into freeing the operation resources after calling WriteChunks rather than when the operation returns.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/WebSocketsAsyncIOEngine.cs
```diff
@@ -0,0 +1,124 @@
...
+        private void CheckInitialized()
+        {
+            if (!_isInitialized)
+            {
+                throw new InvalidOperationException("Socket IO not initialized yet");
+            }
+        }
+
+        public void Dispose()
+        {
```
jkotalik: Any reason to keep this?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/WebSocketsAsyncIOEngine.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            var read = GetReadOperation();
+            read.Initialize(_handler, memory);
+            read.Invoke();
+            return new ValueTask<int>(read, 0);
+        }
+
+        public ValueTask<int> WriteAsync(ReadOnlySequence<byte> data)
+        {
+            var write = GetWriteOperation();
```
jkotalik: Do we need a check Initialized here too?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/WebSocketsAsyncIOEngine.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            var continuation = init.Complete(hr, bytes);
+
+            _initializationFlush = null;
+
+            continuation.Invoke();
+        }
+
+        private void CheckInitialized()
+        {
```
jkotalik: CheckInitialize is really checking that flush was called before Read/Write, correct? I'd rename it to "ThrowIfNotInitialized" or "ThrowIfNotFlushed"

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncWriteOperationBase.cs
pakrym: This one is for pinning pipe blocks.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/WebSocketsAsyncIOEngine.cs
pakrym: Has to be implemented, it's on the interface.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.FeatureCollection.cs
```diff
@@ -267,7 +267,7 @@ void IHttpResponseFeature.OnCompleted(Func<object, Task> callback, object state)
             // Upgrade async will cause the stream processing to go into duplex mode
             AsyncIO = new WebSocketsAsyncIOEngine(_pInProcessHandler);
 
-            await InitializeResponse();
+            await InitializeResponse(true);
```
jkotalik: nit: flushHeaders:

test/IISIntegration.FunctionalTests/Inprocess/SynchronousReadAndWriteTests.cs
```diff
@@ -169,15 +169,26 @@ public SynchronousReadAndWriteTests(IISTestSiteFixture fixture)
...
-                await connection.Receive(
-                    "5",
-                    message,
-                    "");
-                await connection.Receive(
-                    "0",
-                    "",
-                    "");
+
+                // This test can return both content lenght or chunked response
```
jkotalik: nit: length

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
```diff
@@ -0,0 +1,171 @@
...
+                    }
+                    else
+                    {
+                        nextContinuation = next.Invoke();
+
+                        // operation went async
+                        if (nextContinuation == null)
+                        {
+                            _runningOperation = next;
+                        }
```
jkotalik: Do we need to null out _nextOperation if it completed synchronously?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOOperation.cs
```diff
@@ -0,0 +1,158 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks.Sources;
+
+namespace Microsoft.AspNetCore.Server.IISIntegration
+{
+    internal abstract class AsyncIOOperation: IValueTaskSource<int>, IValueTaskSource
+    {
+        private static readonly Action<object> CallbackCompleted = _ => { Debug.Assert(false, "Should not be invoked"); };
```
jkotalik: Throw instead here?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOOperation.cs
```diff
@@ -0,0 +1,158 @@
...
+                return Complete(hr, bytes);
+            }
+            return null;
+        }
+
+        protected abstract bool InvokeOperation(out int hr, out int bytes);
+
+        public AsyncContinuation Complete(int hr, int bytes)
+        {
+            if (hr != NativeMethods.ERROR_OPERATION_ABORTED)
```
jkotalik: Can we move this check to be part of the AsyncReadOperation?

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/WebSocketsAsyncIOEngine.cs
jkotalik: +1

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
pakrym: We assign both to null couple lines above.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOOperation.cs
pakrym: Nah, it's not worth throwing this in customer apps.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
jkotalik: oh im blind.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOOperation.cs
pakrym: We can but it would require slight refactoring I'm not super eager to do at this point.

jkotalik: Yeah, let's just get this in first.

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IISHttpContext.cs
```diff
@@ -275,27 +287,11 @@ private void SetErrorResponseHeaders(int statusCode)
...
-
-            Task flushAsync;
-
-            lock (_stateSync)
-            {
-                DisableReads();
-                flushAsync = Output.FlushAsync();
-            }
-            await flushAsync;
+            await ProduceStart(true);
```
Tratcher: named param

src/Microsoft.AspNetCore.Server.IISIntegration/Server/IO/AsyncIOEngine.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                if (_runningOperation != null)
+                {
+                    if (_nextOperation == null)
+                    {
+                        _nextOperation = ioOperation;
+
+                        if (_runningOperation is AsyncReadOperation)
+                        {
+                            NativeMethods.HttpTryCancelIO(_handler);
```
Tratcher: Comment on why you need to do this.

PR #839, opened 2018-05-14
Split RequestHandler into two dlls

jkotalik:
For #811

PR #845, opened 2018-05-16
Cleanup headers and enable MultiProcessorCompilation

pakrym:
Also get rid of  `    applicationinfo.obj : MSIL .netmodule or module compiled with /GL found; restarting link with /LTCG; add /LTCG to the link command line to improve linker performance` warning

Rules I used for header files:

1. precomp contains only global and very rarely changing libraries - meaning no header files from the current project.
2. All other .h files reference precomp and other header files required for definitions
3. All .cpp files referent their corresponding .h file and additional headers required for implementations.
4. Forward declarations should not be used in all cases except to break circular declaration dependency.

This way we can see what every file uses, make them self-sufficient and avoid including to much unused stuff.


pakrym:
VS decided to change all line endings, please enable ignore whitespace in review settings.

jkotalik:
Once #839 is in, we will do the same with the inprocess and outofprocess request handlers.

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
```diff
@@ -14,6 +19,9 @@ HRESULT
     _Out_ IAPPLICATION       **pApplication
     );
 
+extern BOOL     g_fRecycleProcessCalled;
```
jkotalik: We should figure out where we are externing global variables. Do we want to move all of them outside of the precomp.hxx file or not?

pakrym: We want to remove all of them ideally.

src/AspNetCoreModuleV2/AspNetCore/Inc/precomp.hxx
```diff
@@ -0,0 +1,56 @@
...
+FORCEINLINE
+HRESULT
+HRESULT_FROM_GETLASTERROR()
+{
+    return  ( GetLastError() != NO_ERROR )
+           ? HRESULT_FROM_WIN32( GetLastError() )
+           : E_FAIL;
+}
+
+extern PVOID        g_pModuleId;
```
jkotalik: Anything we can do with these global variables? Put them in the proxy module?

src/AspNetCoreModuleV2/AspNetCore/src/globalmodule.cpp
```diff
@@ -1,4 +1,4 @@
-#include "precomp.hxx"
+#include "globalmodule.h"
```
jkotalik: Copyright.

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
```diff
@@ -2,9 +2,14 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 #pragma once
-#define API_BUFFER_TOO_SMALL 0x80008098
 
-extern BOOL     g_fRecycleProcessCalled;
+#include "precomp.hxx"
```
pan-wang: why do we need to add those includes here?

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
pan-wang: what's the benefit?

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
pakrym: So we see what every file uses and can control it, also makes every header file self contaned.

>We used centralized precompiled header file to avoid adding multiple includes in each individual file and issue of orders

It cause us to have a massive amount of unneeded included headers. There are no ordering issues when every header includes only what it needs. There is a single case of circular dependency that is fixed by forwarding definition.

jkotalik: Circular dependencies are the biggest reason for this change IMO. I've battled a few not recognizing why I caused a circular dependency. It also makes it very clear what classes depend on what.

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
pakrym: It's easy to track who uses and sets global variable when it's part of some class.

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
pakrym: Another benefit is that we don't need to figure out how to order headers in precomp correctly so they compile.

pan-wang: I believe the old IIS team use precompiled header file based on 1) less and clean code and 2) less compilation time. Circular dependency is a code design issue instead of compilation, which could happen even you put include into every header files. I personally still prefer the precompiled header as it compilation efficient and help to identify code issue. This is why VS c+ template use precompiled header by default.

pakrym: Precompiled headers are not enabled in our solution at all. We just use it as an include directive dump and gain no perf.

Additionally, precompiled headers are not supposed to be used for header files from current project at all because it causes precompiled header invalidation every time the code is changed ruining the entire purpose.
Precompiled headers are designed to hold large external includes that rarely change.

> help to identify code issue

Which code issues does it help to identify?

jkotalik:
What is the status on this PR? @pakrym @pan-wang

pakrym:
Waiting for Pan to respond.

jkotalik:
Rebased on dev. Will merge once builds are green.

PR #853, opened 2018-05-22
Add Hostable Web Core based inprocess test server 

pakrym:


src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -106,22 +106,16 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
 
     ReferenceApplicationInfo();
 
-    if (INVALID_FILE_ATTRIBUTES == GetFileAttributes(strFilePath.QueryStr()) &&
-        GetLastError() == ERROR_FILE_NOT_FOUND)
+    if (INVALID_FILE_ATTRIBUTES == GetFileAttributes(strFilePath.QueryStr()))
```
jkotalik: Should we check if GetLastError is FileNotFound or InvalidPath rather than removing this check?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -213,7 +207,7 @@ APPLICATION_INFO::EnsureApplicationCreated(
             // FindRequestHandlerAssembly obtains a global lock, but after releasing the lock,
             // there is a period where we could call
 
-            hr = FindRequestHandlerAssembly();
+            hr = FindRequestHandlerAssembly(&options);
```
jkotalik: I don't like returning options from FindRequestHandlerAssembly, it is only used for the exeLocation.

src/AspNetCoreModuleV2/CommonLib/environmentvariablehelpers.h
```diff
@@ -98,9 +98,10 @@ class ENVIRONMENT_VAR_HELPERS
             dwResult = GetEnvironmentVariable(struNameBuffer.QueryStr(),
                 struValueBuffer.QueryStr(),
                 struValueBuffer.QuerySizeCCH());
-            if (struValueBuffer.IsEmpty())
+
+            if (dwResult <= 0)
```
jkotalik: How was this case being hit? Were environment variables growing/changing between each server start?

src/AspNetCoreModuleV2/CommonLib/hostfxroptions.h
```diff
@@ -0,0 +1,131 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+#pragma once
+#include "precomp.h"
+
+#include "hostfxr_utility.h"
+
+#pragma once
```
jkotalik: multiple pragmas.

src/AspNetCoreModuleV2/CommonLib/hostfxroptions.h
```diff
@@ -0,0 +1,131 @@
...
+{
+public:
+	HOSTFXR_OPTIONS(PWCHAR hostFxrLocation, PWCHAR exeLocation, DWORD argc, BSTR argv[]):
+        m_hostFxrLocation(hostFxrLocation),
+        m_exeLocation(exeLocation),
+        m_argc(argc),
+        m_argv(argv)
+    {}
+
+	~HOSTFXR_OPTIONS()
```
jkotalik: Nit formatting.

src/AspNetCoreModuleV2/CommonLib/hostfxroptions.h
```diff
@@ -0,0 +1,131 @@
...
+#pragma once
+#include "precomp.h"
+
+#include "hostfxr_utility.h"
+
+#pragma once
+class HOSTFXR_OPTIONS
+{
+public:
+	HOSTFXR_OPTIONS(PWCHAR hostFxrLocation, PWCHAR exeLocation, DWORD argc, BSTR argv[]):
```
jkotalik: PCWSTR instead of PWCHAR?

src/AspNetCoreModuleV2/CommonLib/hostfxroptions.h
```diff
@@ -0,0 +1,131 @@
...
+    static
+    HRESULT Create(
+        _In_ PCWSTR         pcwzExeLocation,
+        _In_ PCWSTR         pcwzProcessPath,
+        _In_ PCWSTR         pcwzApplicationPhysicalPath,
+        _In_ PCWSTR         pcwzArguments,
+        _In_ HANDLE         hEventLog,
+        _Out_ HOSTFXR_OPTIONS ** ppWrapper)
+    {
+        HRESULT                 hr = S_OK;
```
jkotalik: Put this in a separate cpp file. Method is too long to be in .h.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -845,7 +867,7 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
     // set the callbacks
     s_Application = this;
 
-    hr = RunDotnetApplication(m_pConfig->QueryHostFxrArgCount(), m_pConfig->QueryHostFxrArguments(), pProc);
+    hr = RunDotnetApplication(hostfxrArgc, hostfxrArgv, pProc);
```
jkotalik: Argc and Argv are never set. Use hostfxr options.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -808,33 +811,52 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
     hostfxr_main_fn     pProc;
 
     DBG_ASSERT(m_status == APPLICATION_STATUS::STARTING);
 
-    // hostfxr should already be loaded by the shim. If not, then we will need
-    // to load it ourselves by finding hostfxr again.
-    hModule = LoadLibraryW(L"hostfxr.dll");
-
-    if (hModule == NULL)
+    pProc = s_fMainCallback;
```
jkotalik: Comment on what maincallback is.

test/IISIntegration.FunctionalTests/IISIntegration.FunctionalTests.csproj
```diff
@@ -1,22 +1,38 @@
...
   <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Server.IIS\Microsoft.AspNetCore.Server.IIS.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Server.IISIntegration\Microsoft.AspNetCore.Server.IISIntegration.csproj" />
     <ProjectReference Include="..\WebSites\**\*.csproj">
       <ReferenceOutputAssembly>False</ReferenceOutputAssembly>
     </ProjectReference>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(OS)' == 'Windows_NT'">
+    <None Include="$(MSBuildThisFileDirectory)..\..\src\AspNetCoreModuleV2\InProcessRequestHandler\bin\$(Configuration)\x64\aspnetcorev2_inprocess.dll" CopyToOutputDirectory="PreserveNewest" Visible="true" Link="%(FileName)%(Extension)" />
```
jkotalik: Does this interfere with what we do for testing that the request handler is loaded via hostfxr?

test/IISIntegration.FunctionalTests/Utilities/TestServer.cs
```diff
@@ -0,0 +1,129 @@
...
+namespace IISIntegration.FunctionalTests.Utilities
+{
+    class TestServer: IDisposable, IStartup
+    {
+        private static readonly SemaphoreSlim WebCoreLock = new SemaphoreSlim(1, 1);
+
+        private static readonly int BasePort = 50690;
+        private static readonly Uri BaseUri = new Uri("http://localhost:" + BasePort);
+
+        private readonly TaskCompletionSource<object> _taskCompletionSource = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
```
jkotalik: change name of _taskCompletionSource to _started

test/IISIntegration.FunctionalTests/Utilities/TestServer.cs
```diff
@@ -0,0 +1,129 @@
...
+            var server = new TestServer(app, loggerFactory);
+            server.Start();
+            await server.HttpClient.GetAsync("/start");
+            await server._taskCompletionSource.Task;
+            return server;
+        }
+
+        private void Start()
+        {
+            LoadLibrary(Environment.ExpandEnvironmentVariables(@"%windir%\system32\inetsrv\hwebcore.dll"));
```
jkotalik: nit space.

test/IISIntegration.FunctionalTests/Utilities/TestServer.cs
```diff
@@ -0,0 +1,129 @@
...
+            LoadLibrary(@"aspnetcorev2_inprocess.dll");
```
jkotalik: we should make these constants in another class in the M.A.S.IIS namespace.

test/IISIntegration.FunctionalTests/Utilities/TestServer.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        private int Main(IntPtr argc, IntPtr argv)
+        {
+            _host = new WebHostBuilder()
+                .UseIIS()
+                .ConfigureServices(services => {
+                        services.AddSingleton<IStartup>(this);
+                        services.AddSingleton<ILoggerFactory>(_loggerFactory);
+                    })
+                .UseSetting(WebHostDefaults.ApplicationKey, typeof(TestServer).GetTypeInfo().Assembly.FullName)
```
jkotalik: Why do we need this setting? For a test?

test/IISIntegration.FunctionalTests/Utilities/TestServer.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            HttpClient = new HttpClient(new LoggingHandler(new SocketsHttpHandler(), _loggerFactory.CreateLogger<TestServer>()))
+            {
+                BaseAddress = BaseUri
+            };
+        }
+
+        public static async Task<TestServer> Create(RequestDelegate app, ILoggerFactory loggerFactory)
+        {
+            await WebCoreLock.WaitAsync();
```
jkotalik: Does this guarantee that only one server is up at a time? It's interesting that we need to do this.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pakrym: I looked through windows other source files and they don't generally check it. 
Additionally, what should we do if we get `INVALID_FILE_ATTRIBUTES` but some other LastError?

src/AspNetCoreModuleV2/CommonLib/environmentvariablehelpers.h
pakrym: Yep we kept appending ASPNETCORE_HOSTING_STARTUP

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocess/inprocessapplication.cpp
pakrym: That's the point - to avoid resolving options for callback case.

test/IISIntegration.FunctionalTests/IISIntegration.FunctionalTests.csproj
pakrym: No, because that happens in test websites, not in the functional test project.

test/IISIntegration.FunctionalTests/Utilities/TestServer.cs
pakrym: Only one HWC can run at a time.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocess/inprocessapplication.cpp
jkotalik: In the non-callback case, we are calling dotnet will a NULL buffer. This codepath is still shared between both.

pakrym: Right

test/IISIntegration.FunctionalTests/Utilities/TestServer.cs
pakrym: Where?

jkotalik: Nvrm, ignoring whitespace changed the diff.

src/AspNetCoreModuleV2/CommonLib/hostfxroptions.h
```diff
@@ -0,0 +1,61 @@
...
+#pragma once
+
+class HOSTFXR_OPTIONS
+{
+public:
+    HOSTFXR_OPTIONS(PCWSTR hostFxrLocation, PCWSTR exeLocation, DWORD argc, BSTR argv[]):
+        m_argc(argc),
+        m_argv(argv)
+    {
+        m_hostFxrLocation.Copy(hostFxrLocation);
```
jkotalik: @pan-wang will make you add a Populate method as copy returns an HR.

test/IISIntegration.FunctionalTests/Utilities/TestServer.cs
pakrym: So ApplicationName is setup correctly. Required because we register IStartup implementation.

src/AspNetCoreModuleV2/CommonLib/hostfxroptions.h
pakrym: Ugh.

jkotalik: I don't care as I think the only hr will be E_OUTOFMEM, which I want us to stop avoiding.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -178,13 +172,13 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
...
 APPLICATION_INFO::EnsureApplicationCreated(
-    IHttpContext *pHttpContext,
-    STRU* struExeLocation
+    IHttpContext *pHttpContext
 )
 {
     HRESULT             hr = S_OK;
     BOOL                fLocked = FALSE;
     IAPPLICATION       *pApplication = NULL;
+    STRU                exeLocation;
```
pan-wang: please follow the same naming conversation and rename it to struExeLocation as it helps to identify the type of this variable

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -240,10 +235,11 @@ APPLICATION_INFO::EnsureApplicationCreated(
 }
 
 HRESULT
-APPLICATION_INFO::FindRequestHandlerAssembly()
+APPLICATION_INFO::FindRequestHandlerAssembly(STRU& location)
 {
     HRESULT             hr = S_OK;
     BOOL                fLocked = FALSE;
+    PCWSTR              handlerDllName;
```
pan-wang: rename it with pstrHandlerDllName

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -267,46 +263,67 @@ APPLICATION_INFO::FindRequestHandlerAssembly()
...
-            if (FAILED(hr = FindNativeAssemblyFromHostfxr(&struFileName)))
+            handlerDllName = g_pwzAspnetcoreInProcessRequestHandlerName;
+        }
+        else
+        {
+            handlerDllName = g_pwzAspnetcoreOutOfProcessRequestHandlerName;
+        }
+
+        // Try to see if RH is already loaded
+        g_hAspnetCoreRH = LoadLibraryW(handlerDllName);
```
pan-wang: this will load the module. If somehow there is a assembly with the same name in some load path, this call will accidently loaded it. Please use GetModuleHandle https://msdn.microsoft.com/en-us/library/windows/desktop/ms683199(v=vs.85).aspx if you just want to make sure if the assembly has been loaded. Alos better perf

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -267,46 +263,67 @@ APPLICATION_INFO::FindRequestHandlerAssembly()
...
+
+        if (g_hAspnetCoreRH == NULL)
+        {
+            if (m_pConfiguration->QueryHostingModel() == APP_HOSTING_MODEL::HOSTING_IN_PROCESS)
             {
-                STACK_STRU(strEventMsg, 256);
-                if (SUCCEEDED(strEventMsg.SafeSnwprintf(
-                    ASPNETCORE_EVENT_INPROCESS_RH_MISSING_MSG)))
+                std:: unique_ptr<HOSTFXR_OPTIONS> options;
```
pan-wang: you can use MULTISZ defined in IISLIB, I prototyped the similar change

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -401,6 +420,8 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
 //
 HRESULT
 APPLICATION_INFO::FindNativeAssemblyFromHostfxr(
+    HOSTFXR_OPTIONS* hostfxrOptions,
```
pan-wang: HOSTFXR_OPTIONS can be replaced by MULTISZ

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -16,7 +19,8 @@ IN_PROCESS_APPLICATION::IN_PROCESS_APPLICATION(
     m_fInitialized(FALSE),
     m_fShutdownCalledFromNative(FALSE),
     m_fShutdownCalledFromManaged(FALSE),
-    m_srwLock()
+    m_srwLock(),
+    m_fDotnetExeLocation(pDotnetExeLocation)
```
pan-wang: m_fDotnetExeLocation is supposed to be BOOL. the naming is wrong

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pakrym: How exactly? I need all 3 fields there, how can I replace all of them with a string wrapper?

pakrym:
@pan-wang I fixed everything but `MULTISZ`, we can iterate on that later.

src/AspNetCoreModuleV2/CommonLib/hostfxroptions.cpp
```diff
@@ -0,0 +1,102 @@
...
+        goto Finished;
+    }
+
+Finished:
+
+    return hr;
+}
+
+
+HRESULT HOSTFXR_OPTIONS::Populate(PCWSTR hostFxrLocation, PCWSTR struExeLocation, DWORD argc, BSTR argv[])
```
jkotalik: hostfxr_main accepts WCHAR* for argv: https://github.com/dotnet/core-setup/blob/master/src/corehost/common/pal.h#L103. Let's file an issue to fix it.

test/IISIntegration.FunctionalTests/Inprocess/TestServerTest.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        public TestServerTest(ITestOutputHelper output = null) : base(output)
+        {
+        }
+
+        [ConditionalFact]
+        public async Task Test()
+        {
+            var helloWorld = "Hello World";
+            using (var testServer = await TestServer.Create(ctx => ctx.Response.WriteAsync(helloWorld), LoggerFactory))
```
jkotalik: Assert something in this test as that is one of the main reasons to make this change.

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
```diff
@@ -147,14 +148,13 @@ class APPLICATION_INFO
...
-        IHttpContext *pHttpContext,
-        STRU*    exeLocation
+        IHttpContext *pHttpContext
     );
 
 private:
-    HRESULT FindRequestHandlerAssembly();
-    HRESULT FindNativeAssemblyFromGlobalLocation(STRU* struFilename);
-    HRESULT FindNativeAssemblyFromHostfxr(STRU* struFilename);
+    HRESULT FindRequestHandlerAssembly(STRU& location);
```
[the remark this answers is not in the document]
pakrym: No need to take the address or dereference, cannot be null, const by default.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -1,10 +1,13 @@
 #include "..\precomp.hxx"
+#include "hostfxroptions.h"
 
 IN_PROCESS_APPLICATION*  IN_PROCESS_APPLICATION::s_Application = NULL;
+hostfxr_main_fn IN_PROCESS_APPLICATION::s_fMainCallback = NULL;
```
pan-wang: please rename it to s_pfnMainCallback, f means BOOL

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocess/inprocessapplication.h
```diff
@@ -176,10 +185,15 @@ class IN_PROCESS_APPLICATION : public APPLICATION
...
     CHAR                            m_pzFileContents[4096] = { 0 };
     DWORD                           m_dwStdErrReadTotal;
+    PCWSTR                          m_pstrDotnetExeLocation;
     static IN_PROCESS_APPLICATION*  s_Application;
 
     REQUESTHANDLER_CONFIG*          m_pConfig;
 
+    // Allows to override call to hostfxr_main with custome callback
+    // used in testing
+    static hostfxr_main_fn          s_fMainCallback;
```
pan-wang: rename the variable name  s_pfnMainCallback

src/AspNetCoreModuleV2/CommonLib/hostfxroptions.cpp
pan-wang: what do you plan to fix? This function does not guard the scenario that the  input of argc is larger than the size of argv, even though it is a private function

Issue #861, opened 2018-05-24
Stop checking for E_OUTOFMEMORY after calling new.


jkotalik:
Since VS2005, the new operator will no longer return null. We can reduce some code bloat by removing the check after any call with new.

PR #863, opened 2018-05-24
New ops can't return null

jkotalik:
For #861.


src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -324,24 +325,14 @@ HOSTFXR_UTILITY::ParseHostfxrArguments(
     }
 
     argv = new BSTR[argc + 1];
-    if (argv == NULL)
-    {
-        hr = E_OUTOFMEMORY;
-        goto Failure;
-    }
 
     argv[0] = SysAllocString(pwzExePath);
```
pakrym: SysAllocString can

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
```diff
@@ -12,10 +12,6 @@ ASPNET_CORE_PROXY_MODULE_FACTORY::GetHttpModule(
 )
 {
     ASPNET_CORE_PROXY_MODULE *pModule = new (pAllocator) ASPNET_CORE_PROXY_MODULE();
```
pakrym: `new (pAllocator) ` can return null too.

jkotalik: good catch.

jkotalik: @pan-wang why was pAllocator added to begin with?

pakrym: It's part of IIS interface:
```
HttpModuleFactory
{
public:
    virtual
    HRESULT
    GetHttpModule(
        _Outptr_ CHttpModule **  ppModule,
        _In_ IModuleAllocator *     pAllocator
    ) = 0;

    virtual
    VOID
    Terminate(
        VOID
    ) = 0;
};
```

pan-wang: This is IIS style of reusing objects instead of allocating a new one every time.  It is assumed for better performance but not sure whether it is still hold with new compiler and OS

PR #864, opened 2018-05-25
Catch all exceptions on ANCM/IIS border

pakrym:
new operator might throw, catch all exceptions before they reach IIS.

### DIFF
--- a/src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
+++ b/src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
@@ -6,8 +6,7 @@
 #include "applicationmanager.h"
 #include "applicationinfo.h"
 #include "acache.h"
-#include "hostfxroptions.h"
-
+#include "exceptions.h"
 
 __override
 HRESULT
@@ -16,13 +15,12 @@ ASPNET_CORE_PROXY_MODULE_FACTORY::GetHttpModule(
     IModuleAllocator *  pAllocator
 )
 {
-    ASPNET_CORE_PROXY_MODULE *pModule = new (pAllocator) ASPNET_CORE_PROXY_MODULE();
-    if (pModule == NULL)
+    try
     {
-        return E_OUTOFMEMORY;
+        *ppModule = THROW_IF_NULL_ALLOC(new (pAllocator) ASPNET_CORE_PROXY_MODULE());;
+        return S_OK;
     }
-    *ppModule = pModule;
-    return S_OK;
+    CATCH_RETURN();
 }
 
 __override
@@ -82,21 +80,23 @@ ASPNET_CORE_PROXY_MODULE::OnExecuteRequestHandler(
     REQUEST_NOTIFICATION_STATUS retVal = RQ_NOTIFICATION_CONTINUE;
     IAPPLICATION* pApplication = NULL;
     STRU struExeLocation;
-
-    if (g_fInShutdown)
+    try
     {
-        hr = HRESULT_FROM_WIN32(ERROR_SERVER_SHUTDOWN_IN_PROGRESS);
-        goto Finished;
-    }
 
-    pApplicationManager = APPLICATION_MANAGER::GetInstance();
+        if (g_fInShutdown)
+        {
+            hr = HRESULT_FROM_WIN32(ERROR_SERVER_SHUTDOWN_IN_PROGRESS);
+            goto Finished;
+        }
 
-    hr = pApplicationManager->GetOrCreateApplicationInfo(
-        g_pHttpServer,
+        pApplicationManager = APPLICATION_MANAGER::GetInstance();
+
+        hr = pApplicationManager->GetOrCreateApplicationInfo(
+            g_pHttpServer,
         pHttpContext,
-        &m_pApplicationInfo);
-    if (FAILED(hr))
-    {
+            &m_pApplicationInfo);
+        if (FAILED(hr))
+        {
         goto Finished;
     }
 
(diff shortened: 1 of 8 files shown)


src/AspNetCoreModuleV2/CommonLib/exceptions.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
jkotalik: copyright header.

src/AspNetCoreModuleV2/CommonLib/exceptions.h
```diff
@@ -0,0 +1,35 @@
...
+{
+    OutputDebugStringA("Unhandled exception:");
+    OutputDebugStringA(exception.what());
+}
+
+inline __declspec(noinline) HRESULT CaughtExceptionHResult()
+{
+    try
+    {
+        throw;
```
jkotalik: Tests for this function.

src/AspNetCoreModuleV2/AspNetCore/Inc/proxymodule.h
```diff
@@ -61,4 +61,4 @@ class ASPNET_CORE_PROXY_MODULE_FACTORY : public IHttpModuleFactory
 
     VOID
     Terminate();
-};
\ No newline at end of file
+};
```
jkotalik: revert

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
```diff
@@ -83,96 +84,104 @@ ASPNET_CORE_PROXY_MODULE::OnExecuteRequestHandler(
     REQUEST_NOTIFICATION_STATUS retVal = RQ_NOTIFICATION_CONTINUE;
     IAPPLICATION* pApplication = NULL;
     STACK_STRU(struFileName, 256);
-
-    if (g_fInShutdown)
+    try
```
jkotalik: Why not wrap the entire function in try catch?

pakrym: returns REQUEST_NOTIFICATION_STATUS and has to set 500 response code.

src/AspNetCoreModuleV2/CommonLib/exceptions.h
pakrym: Good point.

test/CommonLibTests/exception_handler_tests.cpp
```diff
@@ -0,0 +1,49 @@
...
+    }
+    catch(...)
+    {
+        hr = CaughtExceptionHResult();
+    }
+
+    EXPECT_EQ(E_INVALIDARG, hr);
+}
+
+TEST(CaughtExceptionHResult, ReturnsERROR_UNHANDLED_EXCEPTIONForOtherExceptions)
```
jkotalik: Cleanup these names.  ReturnsErrorUnhandledExceptionForOtherExceptions

pakrym:
No perf regression:
```
[10:35:13.768] RequestsPerSecond:           214,168
[10:35:13.768] Max CPU (%):                 8623
[10:35:13.769] WorkingSet (MB):             404
[10:35:13.769] Avg. Latency (ms):           1.2
[10:35:13.769] Startup (ms):                511
[10:35:13.770] First Request (ms):          424.8
[10:35:13.770] Latency (ms):                0.9
[10:35:13.770] Total Requests:              3,233,828
[10:35:13.770] Duration: (ms)               15,100
[10:35:13.770] Socket Errors:               0
[10:35:13.771] Bad Responses:               0
```

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
```diff
@@ -204,7 +214,14 @@ ASPNET_CORE_PROXY_MODULE::OnAsyncCompletion(
     IHttpCompletionInfo *   pCompletionInfo
 )
 {
-    return m_pHandler->OnAsyncCompletion(
-        pCompletionInfo->GetCompletionBytes(),
-        pCompletionInfo->GetCompletionStatus());
+    try
+    {
```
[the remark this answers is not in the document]
pakrym: Anything inside m_pHandler->OnAsyncCompletion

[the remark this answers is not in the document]
jkotalik: Manually try throwing many exceptions in the modules and see what happens? You can write a test by creating your own module but is probably too much work for this change.

jkotalik:
appveyor failed for a flaky test.

pakrym:
@dotnet-bot test this please

src/AspNetCoreModuleV2/CommonLib/exceptions.h
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+#pragma once
+inline VOID ReportException(std::exception& exception)
+{
```
pan-wang: I think we should only report exception for debug build and check whether debug flag enabled as the cost could be very high for release scenario. Have you tested the behavior for the Remarks scenario in https://msdn.microsoft.com/en-us/library/windows/desktop/aa363362(v=vs.85).aspx

jkotalik: The debug flag is undiscoverable and hard to set. I agree that we shouldn't call OutputDebugStringA for every request, but this is only in the failure path.

jkotalik: Maybe make it on by default would be better.

jkotalik:
@pan-wang can you list the modifications you want for this PR?

pan-wang:
Basically I want enable nothrownew in compilation so that we would not hit memory leaks caused by new failure.  Otherwise, we have to deeply scan all functions to wrap the NEW calls and make sure there is no memory leak.

pakrym:
There are better ways to address this problem then compiling against obsolete non standard compliant library. 

We have other places that might throw, and catching and reporting is a nice thing to do.

pakrym:
The other thing is, CLR would crash anyway in OOM situations.

pan-wang:
No. for web server, reliability is a key. You don't want to crash the process even in memory critical situation. Current try-catch, can leak memory and make things even worse. I was not convinced by CLR would crash as I never see it. Current try-catch code opens doors for lots of unknowns. Please following IIS coding guideline which has its good reason on avoiding try-catch.  If you really like, you can start a conversation with Bila or Drago.

pakrym:
CLR and ASP.NET Core are not hardened against OOM and don't catch OutOfMemoryExceptions anywhere so application would almost certainly crash under low memory.

As for leaking memory in when throwing exceptions: there are modern C++ primitives that prevent such leaks from happening in any situation and simplify memory management (take a look at unique_ptr, shared_ptr in C++11). We should certainly look into and start using them. Unknowns are part of the progress.

I don't think that referring to the IIS internal guidance and it's authors is a productive way to discuss technical problems. With all due respect to authors of the guideline it is 20 years old and many things have changed in C++ world including patterns for error handling and exception management.

Additionally we are not trying to contribute code to IIS and have product specifics that are not covered by the said guidance.

1. Much shorter agile release cycle.
2. Being open source and accessible to external contributors.
3. Being easily understood by new developers on the project.

We are free to select the guidance that fits our product the best.

pakrym:
AppVeyor failure: https://github.com/aspnet/IISIntegration/issues/939

PR #874, opened 2018-05-25
Use dynamic ports for Https/Ntlm tests

HaoK:


pakrym:
Let's put this into Hosting, there are other repos that do IIS Https tests.

Tratcher:
I was just thinking that, but I can't find any other examples. Can you?

pakrym:
Don't we have something in ServerTests or MusicStore? If we don't we should add some.

Tratcher:
None at the moment.

jkotalik:
Can we remove setting an explicit port for the deployers and put this in hosting. Giving the option of setting the port is a slippery slope to flaky tests.

Tratcher:
But it would still be nice to have this in Hosting, Right about here: https://github.com/aspnet/Hosting/blob/a3300a6a605dd66d79fabc297e8d8abdb8b8edf6/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs#L74

HaoK:
Is there a consensus on this?  Do you want me to add this to hosting as a shared thing, or just leave it here for now since nothing else needs it today?

Tratcher:
It belongs in hosting.

test/IISIntegration.FunctionalTests/OutOfProcess/HttpsTest.cs
```diff
@@ -30,10 +32,11 @@ public static TestMatrix TestVariants
         [MemberData(nameof(TestVariants))]
         public async Task HttpsHelloWorld(TestVariant variant)
         {
+            var port = Helpers.GetNextSSLPort();
             var deploymentParameters = new DeploymentParameters(variant)
             {
                 ApplicationPath = Helpers.GetOutOfProcessTestSitesPath(),
-                ApplicationBaseUriHint = "https://localhost:44394/",
+                ApplicationBaseUriHint = $"https://localhost:{port}/",
```
Tratcher: Now these can change to 0 :-)

HaoK:
Cool nice, updated removed the setting of the port/hint path

test/IISIntegration.FunctionalTests/OutOfProcess/HttpsTest.cs
```diff
@@ -33,7 +33,6 @@ public static TestMatrix TestVariants
             var deploymentParameters = new DeploymentParameters(variant)
             {
                 ApplicationPath = Helpers.GetOutOfProcessTestSitesPath(),
-                ApplicationBaseUriHint = "https://localhost:44394/",
```
Tratcher: Does this need to be  "https://localhost:0/" so it gets the scheme hint?

Tratcher:
Something's bugy, 1855 isn't a valid HTTPS port. It probably doesn't have your Hosting changes. Update the dependencies?

HaoK:
yeah I think so, I'll update deps again...aren't cross repo changes fun

build/dependencies.props
```diff
@@ -1,42 +1,42 @@
...
-    <MicrosoftAspNetCoreAuthenticationCorePackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreAuthenticationCorePackageVersion>
-    <MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>
-    <MicrosoftAspNetCoreHostingAbstractionsPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHostingAbstractionsPackageVersion>
-    <MicrosoftAspNetCoreHostingPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHostingPackageVersion>
-    <MicrosoftAspNetCoreHttpExtensionsPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHttpExtensionsPackageVersion>
-    <MicrosoftAspNetCoreHttpOverridesPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHttpOverridesPackageVersion>
-    <MicrosoftAspNetCoreHttpPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHttpPackageVersion>
-    <MicrosoftAspNetCoreHttpSysSourcesPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHttpSysSourcesPackageVersion>
-    <MicrosoftAspNetCoreServerIntegrationTestingPackageVersion>0.6.0-a-preview1-tratcher-exp-17050</MicrosoftAspNetCoreServerIntegrationTestingPackageVersion>
```
jkotalik: May still need to target @Tratcher branch of Integration testing?

HaoK: I pushed the hosting changes already yesterday so they should be the recent builds

Tratcher: You should rebase on my branch in this repo as well.

test/IISIntegration.FunctionalTests/OutOfProcess/HttpsTest.cs
HaoK: For some reason the https://localhost:0 does not work.  The tests will get socket aborted exception when doing it, if I explicitlly call GetNextSSLPort and feed it into the BaseUriHint (which is what should be happening anyways) it works...
 
Any ideas @Tratcher ?

Tratcher: Hmm, it should.. https://github.com/aspnet/Hosting/blob/cfab182399e2a22f3e6d87a8875396fced6078c8/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs#L145-L148
Did you rebase and update deps?

HaoK: Yes it works fine for the non https case in Ntlm for example, so there's something weird with the SSL specifically and the hint.

jkotalik:
@HaoK can you rebase with dev to pick up some jenkins changes? There was an issue with the groovy file not uploading a file.

HaoK:
Rebased

jkotalik:
Appveyor failure #939.

Issue #873, opened 2018-05-25
Create a one off build of ANCM-inprocess.


jkotalik:
For partners to try. It will target 2.1.

PR #877, opened 2018-05-29
Allow for alpha builds of ANCM targeting 2.1.0-rc1

jkotalik:
For #873. We can add this to a CI configuration if needed.


natemcmaster:
This is why I was suggesting using a different branch: it is not possible to retarget the the 2.1 SDK in the dev branch. Also, VS won't allow you to set global properties before opening a solution, so the ANCMAlphaBuild flag will only work for command line builds.

jkotalik:
I think we are fine with this only working for command line builds. Though I would like to target the 2.1 SDK, I don't think it is required. 

From what I can tell, we want this branch to effectively be a mirror of dev that targets 2.1 If we create a separate branch, we will have to merge every commit back into the branch.

jkotalik:
cc @davidfowl @muratg

Directory.Build.props
```diff
@@ -5,6 +5,7 @@
 
   <Import Project="version.props" />
   <Import Project="build\dependencies.props" />
+  <Import Condition="'$(ANCMAlphaBuild)' != ''" Project="build\dependencies_alpha.props" />
```
Tratcher: ANCMAlphaBuild -> AncmOobBuild?

I assume you intend to get past alpha at some point.

Tratcher:
How hard would it be to do this as a rolling rebase feature branch? E.g. do the normal dev work targeting 2.2 but once per week rebase the feature branch targeting 2.1 to regression test and produce a partner build? These are not parts of the code with a lot of churn.

jkotalik:
That isn't usually the way we do branching where we take 2.1 and merge it into 2.2. I wouldn't be against that style of branching though.

build/dependencies.props
```diff
@@ -6,14 +6,14 @@
...
-    <MicrosoftAspNetCoreAuthenticationCorePackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreAuthenticationCorePackageVersion>
+    <MicrosoftAspNetCoreAuthenticationCorePackageVersion>2.1.0</MicrosoftAspNetCoreAuthenticationCorePackageVersion>
     <MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>
-    <MicrosoftAspNetCoreHostingAbstractionsPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHostingAbstractionsPackageVersion>
+    <MicrosoftAspNetCoreHostingAbstractionsPackageVersion>2.1.0</MicrosoftAspNetCoreHostingAbstractionsPackageVersion>
     <MicrosoftAspNetCoreHostingPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHostingPackageVersion>
     <MicrosoftAspNetCoreHttpExtensionsPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHttpExtensionsPackageVersion>
     <MicrosoftAspNetCoreHttpOverridesPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHttpOverridesPackageVersion>
     <MicrosoftAspNetCoreHttpPackageVersion>2.2.0-preview1-34326</MicrosoftAspNetCoreHttpPackageVersion>
```
natemcmaster: Seems like everything should point to 2.1.0, right?

jkotalik:
Updated PR to target release/2.1-oob. PRs will still continue to target dev, and we will periodically merge into release/2.1-oob.

build/dependencies.props
jkotalik: Hard question. The packages I updated are the only ones used by M.A.S.IIS. Tests are very likely to break if we target 2.1 as we rely on breaking changes to testing between 2.1 and 2.2. I'll try to make all dependencies 2.1 except internal ones.

jkotalik: The MicrosoftAspNetCoreServerIntegrationTestingPackageVersion has dependencies on logging, logging testing, etc: https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/Microsoft.AspNetCore.Server.IntegrationTesting.csproj#L20. We get package downgrade warnings for all of them if I keep the IntegrationTesting package version at 2.2-preview.

jkotalik:
@natemcmaster this good to go?

PR #881, opened 2018-05-30
Target c++17

jkotalik:
We want a few features in c++17 (like std::filesystem) for future PRs. Setting the version of c++ to 2017 in all vcxproj files we use.


src/AspNetCoreModuleV2/AspNetCore/AspNetCore.vcxproj
```diff
@@ -97,6 +97,7 @@
       <CompileAs>CompileAsCpp</CompileAs>
       <IntrinsicFunctions>true</IntrinsicFunctions>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
+      <LanguageStandard>stdcpp17</LanguageStandard>
```
pan-wang: is there any size change for the output assemblies? have you tested it on Win7 box?

jkotalik: Size is the same. I don't see why I would need to test this on a win7 box, but sure.

jkotalik: Looks fine on win7.

jkotalik:
@pan-wang this good to go?

PR #882, opened 2018-05-30
React to deployer changes

Tratcher:
Depends on https://github.com/aspnet/Hosting/pull/1436
- Pre-build the test projects so we can run tests against them without re-building / publishing.
- Move web.config elements to host.config so they can be mutable for un-published apps.
- Delete web.config to avoid conflicts with host.config. I wasn't able to do this for In-Proc due to the complex web.config. Those tests are all still set to publish.


test/WebSites/OutOfProcessWebSite/web.config
```diff
@@ -1,9 +0,0 @@
-<?xml version="1.0"?>
-<configuration>
```
jkotalik: Still on the fence about whether web.configs should be removed or not.

Tratcher: It's required to avoid conflicts.

Issue #875, opened 2018-05-25
Tests failing with framework not found


pakrym:
https://github.com/aspnet/IISIntegration/pull/872

```
   | [2018-05-25T22:55:52] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Information: Application ready at URL: http://localhost:54999/
   | [2018-05-25T22:55:52] Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.StartupTests Debug: Method: GET, RequestUri: 'http://localhost:54999/HelloWorld', Version: 2.0, Content: <null>, Headers:
   |                                                                                                      {
   |                                                                                                      }
   | [2018-05-25T22:55:52] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Information: iisexpress stdout: IncrementMessages called
   | [2018-05-25T22:55:52] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Information: iisexpress stdout: Request started: "GET" http://localhost:54999/HelloWorld
   | [2018-05-25T22:55:53] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Warning: iisexpress stderr: It was not possible to find any compatible framework version
   | [2018-05-25T22:55:53] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Warning: iisexpress stderr: The specified framework 'Microsoft.NETCore.App', version '2.2.0-preview1-26509-06' was not found.
   | [2018-05-25T22:55:53] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Warning: iisexpress stderr:   - Check application dependencies and target a framework version installed at:
   | [2018-05-25T22:55:53] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Warning: iisexpress stderr:       C:\Program Files\IIS Express\
   | [2018-05-25T22:55:53] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Warning: iisexpress stderr:   - Installing .NET Core prerequisites might help resolve this problem:
   | [2018-05-25T22:55:53] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Warning: iisexpress stderr:       http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
[...]

PR #884, opened 2018-06-01
Fix null ref on ExeLocation not being set

jkotalik:
For #875 (I think). I feel like this should have blown up way more than just 4 tests failing. 

Added a test and modified header includes to have less headers in precomp.hxx for the inprocess request handler project.


src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -103,31 +111,47 @@ __stdcall
...
     if (FAILED(hr))
     {
-        return hr;
+        goto Finished;
     }
 
-    pApplication = new IN_PROCESS_APPLICATION(pServer, pConfig, pwzExeLocation);
-    if (pApplication == NULL)
+    pApplication = new IN_PROCESS_APPLICATION(pServer, pConfig);
+
```
pan-wang: you should do NULL check for pApplication

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -103,31 +111,47 @@ __stdcall
...
+    hr = pApplication->Initialize(pwzExeLocation);
```
pan-wang: if Initialize failed, you will double delete pConfig which will caused undefined behavior. first is done by the destructor of in_process_application and then by the function. This is why we usually use reference counter

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
jkotalik: It can't be NULL as new will never return NULL.

Tratcher:
Any indication why it's only an issue on Win7/R2?

jkotalik:
> Any indication why it's only an issue on Win7/R2?

No, but when I ran this on a test machine, it was the first failure I found. I'll continue chunking through them if tests still fail on win7.

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
pan-wang: or you can simply set pConfig to NULL after new success indicating this config object belongs to application

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
pan-wang: just checked IISOOB and IIS, it seems we always link to nothrownew.obj to avoid exception

jkotalik: I disagree. It has been the standard to throw on new for the last 15 years. If we are out of memory, I would like to crash the process (or we can catch it).

pan-wang: we need at least add nothrownew for out process and shim

jkotalik: I'll make a separate issue for it.

build/repo.targets
```diff
@@ -119,7 +119,7 @@
   </Target>
 
   <Target Name="RunNativeTest" DependsOnTargets="GetToolsets">
-    <Exec Command="&quot;$(VisualStudioMSBuildx86Path)&quot; &quot;$(RepositoryRoot)test\CommonLibTests\CommonLibTests.vcxproj&quot; /t:Test $(BuildArgs) -p:Platform=%(Platforms.VCPlatform)"
+    <Exec Command="&quot;$(VisualStudioMSBuildx86Path)&quot; &quot;$(RepositoryRoot)test\CommonLibTests\CommonLibTests.vcxproj&quot; /t:Test $(BuildArgs) -p:Platform=x64"
```
pakrym: Why?

src/AspNetCoreModuleV2/CommonLib/stdafx.h
```diff
@@ -36,8 +36,12 @@
 #include "fx_ver.h"
 #include "hostfxr_utility.h"
 #include "hostfxroptions.h"
+<<<<<<< HEAD
```
pakrym: Cleanup/

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -1,8 +1,16 @@
 // dllmain.cpp : Defines the entry point for the DLL application.
+
```
pakrym: Move before comment.

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
pakrym: I would argue that we don't need nothrownew in shim

pakrym:
So what exactly was the issue?

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
pan-wang: Thank about there was a bug in our code which leads to null ref under stress. In old patten,  worker process will crash, while with try-catch, client occasionally receive 500 response. Will it be easy to debug the root cause for the latter?

pakrym: Sorry, I didn't really understand. I don't think there is such a thing as null reference exception in C++ so behavior in this case wouldn't change at all.

build/repo.targets
jkotalik: 1. Only wanted to run tests once.
2. I'll need to fix some linking issue with obj files.

jkotalik:
> So what exactly was the issue?

https://github.com/aspnet/IISIntegration/pull/884/files#diff-4653e8c4233ce04437f43552addb24f3L23

String is not set in a STRU, so it becomes corrupted after return.

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
jkotalik: Done.

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
jkotalik: Yeah, we would probably AV or have undefined behavior.

pan-wang: if new failed here, exception will be thrown and pConfig will be leaked.

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -103,31 +111,56 @@ __stdcall
...
+    HRESULT                 hr = S_OK;
+    IN_PROCESS_APPLICATION *pApplication = NULL;
+    REQUESTHANDLER_CONFIG  *pConfig = NULL;
+
     // Initialze some global variables here
     InitializeGlobalConfiguration(pServer);
 
-    hr = REQUESTHANDLER_CONFIG::CreateRequestHandlerConfig(pServer, pHttpContext->GetApplication(), &pConfig);
-    if (FAILED(hr))
+    try
```
pan-wang: if we want to user try-catch, we have to rescan all new operations like this to make sure they are safe.  I personally still prefer the NULL check.

jkotalik: I'll do another scan now.

jkotalik:
Test failures only seem to be flaky on jenkins builds, will make a follow up issue for it.

Issue #880, opened 2018-05-30
Make gtest and gmock a submodule


jkotalik:
See https://github.com/aspnet/SignalR/commit/508e66ece657938d3d3fdba990175c5ba1c38603 for how signalR did it.

PR #885, opened 2018-06-01
Make gtest a submodule

jkotalik:
For #880


jkotalik:
Think #884 will fix some of the tests failures for flakiness.

PR #893, opened 2018-06-06
Add VSTS ci pipeline

ryanbrandenburg:


jkotalik:
@ryanbrandenburg add: .\tools\update_schema.ps1 to the script section. The reason tests fail is that the aspnetcore schema isn't installed.

ryanbrandenburg:
Is `update_schema.ps1` only required on windows?

jkotalik:
Yeah.

jkotalik:
Says the operation was canceled?

ryanbrandenburg:
I suspect it's an issue with VSTS agent failing to capture the log, it happens on a lot of windows builds. I've started a thread with them.

jkotalik:
@ryanbrandenburg Can we merge this?

ryanbrandenburg:
Even though it seems to be a problem on the VSTS end, I don't think we should merge until we've seen a green build from VSTS.

ryanbrandenburg:
I believe this is blocked by the same thing as aspnet/Razor#2395. We'll need the VSTS team to help us get unblocked here.

ryanbrandenburg:
Merging this so we generate more data for the VSTS team.

natemcmaster:
Assigning to myself to drive this one.

jkotalik:
@natemcmaster are shells and consoles running in admin on VSTS? From what I remember, they are not.

natemcmaster:
Would update_schema.ps1 fail if it were not running as admin?

natemcmaster:
https://dotnet.visualstudio.com/public/_build/results?buildId=8471&view=ms.vss-test-web.test-result-details

It looks like most of the tests fail with the same problem. HTTPS tests fail with

```
System.Net.Http.HttpRequestException : The SSL connection could not be established, see inner exception.
---- System.IO.IOException : Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host..
-------- System.Net.Sockets.SocketException : An existing connection was forcibly closed by the remote host.
```

HTTP tests fail with 
```
Assert.Equal() Failure
↓ (pos 9)
Expected: HTTP/1.1 200 OK\r\n
Actual: HTTP/1.1 503 Serv
↑ (pos 9)
```

jkotalik:
Yeah it would fail, it's writing a file to a config directory.

natemcmaster:
It seems to be ok. I think admin or not permissions may vary based on Agent Pool. For now, we're using the Hosted VS2017 pool which apparently gives us admin rights.

image

jkotalik:
Think you need to re-add the Microsoft.NET.Test.Sdk
```
System.IO.FileNotFoundException: Unable to find tests for /Users/vsts/agent/2.136.1/work/1/s/test/Common.Tests/bin/Release/netcoreapp2.2/Common.Tests.dll. Make sure test project has a nuget reference of package "Microsoft.NET.Test.Sdk" and framework version settings are appropriate. Rerun with /diag option to diagnose further.
2018-07-27T21:32:31.5417570Z      at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Hosting.DotnetTestHostManager.GetTestHostProcessStartInfo(IEnumerable`1 sources, IDictionary`2 environmentVariables, TestRunnerConnectionInfo connectionInfo)
2018-07-27T21:32:31.5433980Z      at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager.SetupChannel(IEnumerable`1 sources, CancellationToken cancellationToken)
2018-07-27T21:32:31.5450240Z      at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyExecutionManager.StartTestRun(TestRunCriteria testRunCriteria, ITestRunEventsHandler eventHandler)
```

natemcmaster:
It's not really a test project though, right?  It doesn't have any test cases. Btw, you just put those in to a shproj. I'm not sure you really need to factor out a new test assembly for two classes.

jkotalik:
Yeah it isn't a test project. Would a shproj have the same capabilities as this? We followed what Kestrel did to test libuv and sockets: https://github.com/aspnet/KestrelHttpServer/tree/release/2.2/test/Kestrel.FunctionalTests https://github.com/aspnet/KestrelHttpServer/tree/release/2.2/test/Kestrel.Transport.Libuv.FunctionalTests

natemcmaster:
The difference is there isn't a Kestrel.FunctionalTests.csproj file. Instead, libuv/sockets test projects do `<Compile Include="..\Kestrel.FunctionalTests\**\*.cs" />`

shproj is just VS sugar for doing the same thing. Either way is fine with me

natemcmaster:
Adding Restart-Service w3svc didn't help. I don't know what's going wrong here.

natemcmaster:
I'm talking about https://github.com/aspnet/IISIntegration/blob/release/2.2/test/Common.Tests/Common.Tests.csproj, not the FunctionalTests folder.

jkotalik:
We should improve our naming 😢

natemcmaster:
Ok @jkotalik @pakrym , we have a few options to move this forward:

1. Merge with tests broken. Jenkins and AppVeyor still work. Open an issue to fix tests in VSTS
3. Skip all tests in VSTS and merge.
2. Leave this PR open

Which do you prefer?

pakrym:
Let's do option 1

pakrym:
Do we have access to event log on VSTS machines?

jkotalik:
Let me try one thing first to fix IIS tests.

natemcmaster:
> Do we have access to event log on VSTS machines?

No, they are temporary VMs which are disposed after use, but we can gather the logs after test runs. What is the logname we gather?

cref https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-eventlog?view=powershell-5.1

jkotalik:
Sweet only the HttpsTests are failing (as well as another flaky tests that is logged)

natemcmaster:
Just chatted with Justin. We'll use #1111  to follow up on the remaining failing tests.

Issue #859, opened 2018-05-24
ANCM Request Handler versioning spec


shirhatti:
ANCM v2 will be installed to `%ProgramFiles%\IIS\AspNetCoreModule`

```
> tree %ProgramFiles%\IIS\AspNetCoreModule
├── aspnetcorev2.dll
└── requestHandler
    ├── 1.0
    │   └── outofprocessrh.dll
    ├── 2.0
    │   └── outofprocessrh.dll
    ├── 2.1-preview1
    │   └── outofprocessrh.dll
    ├── 2.1
    │   └── outofprocessrh.dll
    └── 2.2-preview1
        └── outofprocessrh.dll
```

In the happy path, the user doesn't have to specify a version of the out-of-process requestHandler. By default the latest stable version wins. As an aside, if the `handlerVersion` is ignored when `hostingModel="inprocess"`

```xml
<configuration>
    <system.webServer>
        <handlers>
            <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified"/>
        </handlers>
        <aspNetCore processPath="dotnet"
                    arguments="myapp.dll"
                    hostingModel="outofprocess"/>
     </system.webServer>
</configuration>
```

If you don't want to be transparently upgraded you can either pin to a specific version or specify a version range or wildcard. For e.g.,

```xml
<configuration>
    <system.webServer>
        <handlers>
            <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified"/>
        </handlers>
        <aspNetCore processPath="dotnet"
                    arguments="myapp.dll"
                    handlerVersion=">=2.1"
                    hostingModel="outofprocess"/>
     </system.webServer>
</configuration>
```

This table denotes all the permissible values for specifying a **stable** version range. Previews don't work with range specifiers and only work with exact version. (I tried to borrow logic from NuGet/hostfxr but they are all predicated on using SemVer. Given that our requestHandler has to maintain back-compat we're effectively always performing a *patch* release and the additional complexity wasn't worth it.)

> NOTE: Applying version range rules  or exact version matches can prevent you from receiving security updates.

| Notation | Applied rule | Description |
|----------|--------------|-------------|
| 1.0 | x == 1.0 | Exact version match |
| >1.0 | x > 1.0 | Minimum version, exclusive |
| >=1.0 | x ≥ 1.0 | Minimum version, inclusive |
[...]

PR #895, opened 2018-06-06
Load ANCM out of process via global versioning

jkotalik:
For #859 . There is still some cleanup I need to do but I will be most likely offline the rest of the night. cc/ @muratg


jkotalik:
@pan-wang this is good for review now. I took into consideration your offline comments.

src/AspNetCoreModuleV2/CommonLib/config_utility.h
```diff
@@ -0,0 +1,76 @@
...
+        HRESULT hr = S_OK;
+        IAppHostElement                *pHandlerSettings = NULL;
+        IAppHostElementCollection      *pHandlerSettingsCollection = NULL;
+        ENUM_INDEX                      index;
+        IAppHostElement                *pHandlerVar = NULL;
+        STRU strHandlerName;
+        STRU strHandlerValue;
+
+        hr = GetElementChildByName(pElement,
+            CS_ASPNETCORE_HANDLER_SETTINGS,
```
pan-wang: please call release for pHandlerSettings , pHandlerSettingsCollection  and pHandlerVar, otherwise memory leak

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
```diff
@@ -0,0 +1,108 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "stdafx.h"
+#include <experimental/filesystem>
+
+// throws runtime error if no request handler versions are installed.
+std::wstring
+GlobalVersionUtility::GetGlobalRequestHandlerPath(PCWSTR pwzAspNetCoreFolderPath, PCWSTR pwzHandlerVersion, PCWSTR pwzHandlerName)
+{
+    std::wstring handlerVersion(pwzHandlerVersion);
```
pan-wang: why not use folderVersion here? in line 15 you can check if folderVersion is empty and reset it and thus you don't need line 21 to copy the string.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -358,60 +359,46 @@ APPLICATION_INFO::FindRequestHandlerAssembly(STRU& location)
 
 HRESULT
 APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(
-    PCWSTR libraryName,
-    STRU* struFilename)
+    PCWSTR pstrHandlerDllName,
```
pan-wang: we need a mechanism to only log the error once. I will do it in my PR

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -358,60 +359,46 @@ APPLICATION_INFO::FindRequestHandlerAssembly(STRU& location)
...
-    // the path (the targeted dll is the same folder of w3wp.exe/iisexpress)
-    // let's still load with full path to avoid security issue
-    if (FAILED(hr = struFilename->Resize(dwSize + 20)))
-    {
-        goto Finished;
-    }
+    std::wstring modulePath; // path to module
 
-    while (!fDone)
+    try
```
pan-wang: jumping between two different code pattern is hard :)

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -358,60 +359,46 @@ APPLICATION_INFO::FindRequestHandlerAssembly(STRU& location)
...
-    dwPosition = struFilename->LastIndexOf(L'\\', 0);
-    struFilename->QueryStr()[dwPosition] = L'\0';
-
-    if (FAILED(hr = struFilename->SyncWithBuffer()) ||
-        FAILED(hr = struFilename->Append(L"\\")) ||
-        FAILED(hr = struFilename->Append(libraryName)))
+    catch (std::exception& e)
     {
-        goto Finished;
+        STRU struEvent;
```
pakrym: If you want to catch copy the macros.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -358,60 +359,46 @@ APPLICATION_INFO::FindRequestHandlerAssembly(STRU& location)
...
+        if (SUCCEEDED(struEvent.CopyA(e.what())))
+        {
+            UTILITY::LogEvent(g_hEventLog,
```
pakrym: LogEventF, ASPNETCORE_EVENT_RECYCLE_APPOFFLINE?

jkotalik: copy paste error. good find.

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
```diff
@@ -0,0 +1,108 @@
...
+        .append(pwzHandlerName);
+    return aspNetCoreFolderPath;
+}
+
+// Throw filesystem_error if directory_iterator can't iterate over the directory
+std::vector<fx_ver_t>
+GlobalVersionUtility::GetRequestHandlerVersions(PCWSTR pwzAspNetCoreFolderPath)
+{
+    std::vector<fx_ver_t> versionsInDirectory;
+    for (auto& p : std::experimental::filesystem::directory_iterator(pwzAspNetCoreFolderPath))
```
pakrym: Alias std::experimental::filesystem to something sensible.

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.h
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma once
+#include <experimental/filesystem>
```
pakrym: filesystem is not experimental in C++17

src/AspNetCoreModuleV2/CommonLib/config_utility.h
pakrym: Or stick them into unique_ptr.

src/AspNetCoreModuleV2/CommonLib/config_utility.h
```diff
@@ -0,0 +1,76 @@
...
+            SUCCEEDED(hr);
+            hr = FindNextElement(pHandlerSettingsCollection, &index, &pHandlerVar))
+        {
+            if (hr == S_FALSE)
+            {
+                hr = S_OK;
+                break;
+            }
+
+            if (FAILED(hr = GetElementStringProperty(pHandlerVar,
```
pakrym: This is pretty hard to read.

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.h
jkotalik: There were issues with filesystem for c++17 on commandline. It kept saying std::filesystem was undefined even though all projects were targeting c++17.

test/IISIntegration.FunctionalTests/IISIntegration.FunctionalTests.csproj
```diff
@@ -14,7 +14,19 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Server.IIS\Microsoft.AspNetCore.Server.IIS.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Server.IISIntegration\Microsoft.AspNetCore.Server.IISIntegration.csproj" />
-    <ProjectReference Include="..\WebSites\**\*.csproj">
+    <ProjectReference Include="..\WebSites\InProcessWebSite\InProcessWebSite.csproj">
```
pakrym: Why expand?

version.props
```diff
@@ -11,5 +11,6 @@
     <AspNetCoreModuleVersionMajor>8</AspNetCoreModuleVersionMajor>
     <AspNetCoreModuleVersionMinor>1</AspNetCoreModuleVersionMinor>
     <AspNetCoreModuleVersionRevision>0</AspNetCoreModuleVersionRevision>
+    <ANCMOutOfProcessVersion>2.0.0</ANCMOutOfProcessVersion>
```
pakrym: AspNetCoreModuleOutOfProcessVersion

nuget/AspNetCoreV2.nuspec
```diff
@@ -21,8 +21,8 @@
   <files>
     <file src="src\AspNetCoreModuleV2\AspNetCore\bin\$Configuration$\Win32\aspnetcorev2.dll" target="contentFiles\any\any\x86\aspnetcorev2.dll" />
     <file src="src\AspNetCoreModuleV2\AspNetCore\bin\$Configuration$\x64\aspnetcorev2.dll" target="contentFiles\any\any\x64\aspnetcorev2.dll" />
-    <file src="src\AspNetCoreModuleV2\OutOfProcessRequestHandler\bin\$Configuration$\Win32\aspnetcorev2_outofprocess.dll" target="contentFiles\any\any\x86\aspnetcorev2_outofprocess.dll" />
-    <file src="src\AspNetCoreModuleV2\OutOfProcessRequestHandler\bin\$Configuration$\x64\aspnetcorev2_outofprocess.dll" target="contentFiles\any\any\x64\aspnetcorev2_outofprocess.dll" />
+    <file src="src\AspNetCoreModuleV2\OutOfProcessRequestHandler\bin\$Configuration$\Win32\aspnetcorev2_outofprocess.dll" target="contentFiles\any\any\x86\$(ANCMOutOfProcessVersion)\aspnetcorev2_outofprocess.dll" />
```
pakrym: Can we build dll with the correct name instead of renaming it later?

test/IISIntegration.FunctionalTests/IISIntegration.FunctionalTests.csproj
jkotalik: msbuild has been automatically expanding it.

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
```diff
@@ -0,0 +1,108 @@
...
+    return fileName.substr(0, fileName.find_last_of(L"\\/") + 1); // Keep the trailing slash.
+}
+
+HRESULT
+GlobalVersionUtility::GetModuleName(HMODULE hModuleName, std::wstring& returnString)
+{
+    HRESULT hr = S_OK;
+    DWORD dwSize = MAX_PATH;
+    BOOL  fDone = FALSE;
+    STRU struFilename;
```
pakrym: Don't mix STRU and wstring

pakrym:
Functional test would be nice for this.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
jkotalik: I'd say we centralize all logs to the event viewer.

src/AspNetCoreModuleV2/CommonLib/config_utility.h
jkotalik: As this is used by the aspnetcore_shim_config, I'm going to avoid std for now. I'd rather have the class all or none for exception handling.

nuget/AspNetCoreV2.nuspec
jkotalik: I assume you are referring to the folder path, correct?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pan-wang: I meant this function would be called for every request in case of failure, (e.g., requesthandler folder was deleted by user) and thus caused log flooding. We should ensure this kind event logged once per application until some change (appoffline or configuration) happened.

test/WebSites/OutOfProcessWebSite/Startup.cs
```diff
@@ -84,5 +86,25 @@ public Task RestrictedNTLM(HttpContext context)
 
         public Task UpgradeFeatureDetection(HttpContext context) =>
             context.Response.WriteAsync(context.Features.Get<IHttpUpgradeFeature>() != null? "Enabled": "Disabled");
+
+        public Task CheckRequestHandlerVersion(HttpContext context)
+        {
+            // We need to check if the aspnetcorev2_outofprocess dll is loaded by iisexpress.exe
+            // As they aren't in the same process, we will try to delete the file and expect a file
```
pakrym: You can use Process.Modules on the test side to see what modules are loaded into iiexpress.

test/WebSites/OutOfProcessWebSite/Startup.cs
```diff
@@ -84,5 +86,25 @@ public Task RestrictedNTLM(HttpContext context)
...
+            catch(UnauthorizedAccessException)
+            {
+                return context.Response.WriteAsync("Hello World");
+            }
+
+            return context.Response.WriteAsync(context.Request.Headers["ANCMRHPath"]);
+        }
+
+        [DllImport("kernel32.dll")]
+        private static extern IntPtr GetModuleHandle(string lpModuleName);
```
pakrym: Remove?

test/IISIntegration.FunctionalTests/OutOfProcess/GlobalVersionTests.cs
```diff
@@ -0,0 +1,160 @@
...
+        [InlineData("2.1.0-preview")]
+        public async Task GlobalVersion_MultipleRequestHandlers_UpgradeWorks(string version)
+        {
+            var deploymentParameters = GetGlobalVersionBaseDeploymentParameters();
+
+            var deploymentResult = await DeployAsync(deploymentParameters);
+            var originalANCMPath = Path.Combine(deploymentResult.DeploymentResult.ContentRoot,
+                deploymentParameters.RuntimeArchitecture.ToString(),
+                "2.0.0",
+                "aspnetcorev2_outofprocess.dll");
```
pakrym: const

test/CommonLibTests/GlobalVersionTests.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+        auto result = GlobalVersionUtility::GetGlobalRequestHandlerPath(tempPath.c_str(), L"2.0.0", L"aspnetcorev2_outofprocess.dll");
+
+        EXPECT_STREQ(result.c_str(), (tempPath + L"2.0.0\\aspnetcorev2_outofprocess.dll").c_str());
+    }
+
+    TEST(GetGlobalRequestHandlerPath, FindHighestVersionSpecificPreview)
+    {
+        std::wstring tempPath = Helpers::CreateRandomTempDirectory();
+        EXPECT_TRUE(std::experimental::filesystem::create_directories(tempPath + L"\\2.0.0"));
```
pakrym: alias std::experimental::filesystem

src/AspNetCoreModuleV2/CommonLib/utility.h
```diff
@@ -3,6 +3,9 @@
 
 #pragma once
 
+#include "stdafx.h"
+#include "fx_ver.h"
```
pakrym: Is it used in declaration?

src/AspNetCoreModuleV2/CommonLib/stdafx.h
```diff
@@ -31,14 +33,16 @@
 #include "application.h"
 #include "SRWLockWrapper.h"
 #include "environmentvariablehash.h"
+#include "fx_ver.h"
```
pakrym: You seem to include this header into h and cpp files themselves. Do you need it here?

pakrym: Same for GlobalVersionUtility

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+std::wstring 
+GlobalVersionUtility::GetModuleName(HMODULE hModuleName)
+{
+    DWORD dwSize = MAX_PATH;
+    BOOL  fDone = FALSE;
+
+    // Instead of creating a temporary buffer, use the std::wstring directly as the receive buffer.
+    std::wstring retVal;
+    retVal.resize(dwSize + 20);
```
pakrym: What is this 20?

pakrym:
LGTM, but wait for @pan-wang's review.

test/WebSites/OutOfProcessWebSite/Startup.cs
jkotalik: I don't see a good way to get the iisexpress process associated with the test. The Process is private: https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs#L31

jkotalik:
🆙 📅

src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
```diff
@@ -105,6 +106,14 @@ class ASPNETCORE_SHIM_CONFIG : IHttpStoredContext
         return m_hostingModel;
     }
 
+    STRU*
```
pan-wang: please change the return type as PCWSTR as the caller really needs is the string pointer and you don't want caller to change the string value

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -358,60 +358,47 @@ APPLICATION_INFO::FindRequestHandlerAssembly(STRU& location)
...
+            return hr;
         }
     }
-
-    if (FAILED(hr = struFilename->SyncWithBuffer()))
+    catch (std::exception& e)
     {
-        goto Finished;
+        STRU struEvent;
+        if (SUCCEEDED(struEvent.CopyA(e.what())))
```
pan-wang: you may want to give more info in this error log, e.g., application information and path tried to searched. It will help user to identify the root cause.

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+#include "stdafx.h"
+#include <experimental/filesystem>
+
+namespace fs = std::experimental::filesystem;
+
+// throws runtime error if no request handler versions are installed.
+std::wstring
+GlobalVersionUtility::GetGlobalRequestHandlerPath(PCWSTR pwzAspNetCoreFolderPath, PCWSTR pwzHandlerVersion, PCWSTR pwzHandlerName)
+{
```
pan-wang: please add debug assert for the input parameters which should not be NULL indicating the contact between caller and callee

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
```diff
@@ -0,0 +1,101 @@
...
+    aspNetCoreFolderPath = aspNetCoreFolderPath
+        .append(folderVersion)
+        .append(pwzHandlerName);
+    return aspNetCoreFolderPath;
+}
+
+// Throw filesystem_error if directory_iterator can't iterate over the directory
+std::vector<fx_ver_t>
+GlobalVersionUtility::GetRequestHandlerVersions(PCWSTR pwzAspNetCoreFolderPath)
+{
```
pan-wang: if passed in pointer is NULL, I think you will get AV. please make sure it is not NULL or at least do debug assert

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+    std::sort(versionsInDirectory.begin(), versionsInDirectory.end());
+
+    return versionsInDirectory.back().as_str();
+}
+
+// Throws std::out_of_range if there is an index out of range
+std::wstring
+GlobalVersionUtility::RemoveFileNameFromFolderPath(std::wstring fileName)
+{
```
pan-wang: do null assert. I don't think AV will be thrown as an exception

src/AspNetCoreModuleV2/CommonLib/config_utility.h
```diff
@@ -0,0 +1,98 @@
...
+            SUCCEEDED(hr);
+            hr = FindNextElement(pHandlerSettingsCollection, &index, &pHandlerVar))
+        {
+            if (hr == S_FALSE)
+            {
+                hr = S_OK;
+                break;
+            }
+
+            hr = hr = GetElementStringProperty(pHandlerVar,
```
pan-wang: double hr

src/AspNetCoreModuleV2/CommonLib/config_utility.h
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            if (strHandlerName.Equals(CS_ASPNETCORE_HANDLER_VERSION))
+            {
+                hr = strHandlerVersionValue->Copy(strHandlerValue);
+                goto Finished;
+            }
+
+            strHandlerName.Reset();
+            strHandlerValue.Reset();
```
pan-wang: please call pHandlerVar->Release here as you reuse the pointer in the loop

src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
jkotalik: All other calls in the shim_config return STRU*. I agree with you, but we should change the rest of them too if that is the case.

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.h
jkotalik: [The official microsoft docs](https://docs.microsoft.com/en-us/cpp/standard-library/filesystem) and [Community answers](https://developercommunity.visualstudio.com/content/problem/116454/c-filesystem-still-experimental-even-with-stdclate.html) say std file system is in the technical specification, but not part of the standard. I think we can still rely on it though. We use very trivial functionality of it and can adjust accordingly. cc @shirhatti

jkotalik:
@dotnet-bot test this

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -358,60 +358,48 @@ APPLICATION_INFO::FindRequestHandlerAssembly(STRU& location)
...
-    struFilename->QueryStr()[dwPosition] = L'\0';
-
-    if (FAILED(hr = struFilename->SyncWithBuffer()) ||
-        FAILED(hr = struFilename->Append(L"\\")) ||
-        FAILED(hr = struFilename->Append(libraryName)))
+    catch (...)
     {
-        goto Finished;
+        // as our error handling 
+        hr = HRESULT_FROM_WIN32(GetLastError());
```
pakrym: I think this is wrong as is the previous GetLastError call. There is no guarantee that LastError is set when exception is thrown it may return OK in these cases. Just set it to some predefined HR here.

src/AspNetCoreModuleV2/CommonLib/config_utility.h
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma once
+
+#include "stdafx.h"
+
+class ConfigUtility
```
pakrym: Not blocking this PR, but why isn't this part of shim config itself?

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
```diff
@@ -0,0 +1,130 @@
...
+    if (pwzAspNetCoreFolderPath == NULL)
+    {
+        throw new std::invalid_argument("pwzAspNetCoreFolderPath is NULL");
+    }
+
+    std::vector<fx_ver_t> versionsInDirectory = GetRequestHandlerVersions(pwzAspNetCoreFolderPath);
+    if (versionsInDirectory.empty())
+    {
+        throw new std::runtime_error("Cannot find request handler next to aspnetcorev2.dll. Verify a version of the request handler is installed in a version folder.");
```
pakrym: Resources?

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    return versionsInDirectory.back().as_str();
+}
+
+// Throws std::out_of_range if there is an index out of range
+// Throw invalid_argument if any argument is null
+std::wstring
+GlobalVersionUtility::RemoveFileNameFromFolderPath(std::wstring fileName)
+{
+    return fileName.substr(0, fileName.find_last_of(L"\\/") + 1); // Keep the trailing slash.
```
pakrym: fs::path can do this

src/AspNetCoreModuleV2/CommonLib/config_utility.h
jkotalik: Will be shared with request_handler_config eventually.

src/AspNetCoreModuleV2/CommonLib/GlobalVersionUtility.cpp
jkotalik: We need to discuss this. Exceptions only work with CHAR*, while resource and all other windows strings take WCHAR.

src/AspNetCoreModuleV2/CommonLib/config_utility.h
```diff
@@ -0,0 +1,105 @@
...
+            if (strHandlerName.Equals(CS_ASPNETCORE_HANDLER_VERSION))
+            {
+                hr = strHandlerVersionValue->Copy(strHandlerValue);
+                goto Finished;
+            }
+
+            strHandlerName.Reset();
+            strHandlerValue.Reset();
+
+            pHandlerVar->Release();
```
pan-wang: pHandlerVar = NULL;  you should AV under debugger due to double release at Finished

src/AspNetCoreModuleV2/CommonLib/config_utility.h
pakrym: Why?

jkotalik: The handler settings section needs is meant to be read and used by the request handlers. It's only used by the shim now.

src/AspNetCoreModuleV2/CommonLib/config_utility.h
pakrym: Right, but the pattern we use here (setting the field) is not sharable anyway. Additionally, they would probably care about different properties.

jkotalik: Good point. I originally implemented this to populate a std::map.

Issue #897, opened 2018-06-07
Flaky test: LoggingTests.CheckStdoutLogging


jkotalik:
This is questionably a product bug. Needs more investigation. Will disable for now as they fail frequently enough.

jkotalik:
Will be monitoring for any failures.

Issue #766, opened 2018-04-09
Flaky test:  SynchronousReadAndWriteTests


ryanbrandenburg:
# Likely Root Cause
https://github.com/aspnet/IISIntegration/commit/22a865b83237cd50bdb7b764a87c28a3ec046afa

# Test History
* Individual Test Case: http://aspnetci/project.html?projectId=AspNetEf&testNameId=-8781713674742420153&tab=testDetails
* All IISIntegration flaky tests: http://aspnetci/project.html?projectId=AspNetEf&tab=flakyTests&buildTypeId=Lite_IISIntegration
  * Many `SynchronousReadAndWriteTests` near the top when sorted by failure count

# Example
This test [fails](http://aspnetci/viewLog.html?buildId=443357&buildTypeId=XPlat_Windows_Win7_Universe) occasionally on the CI with the following error:

```
System.Net.Http.HttpRequestException : An error occurred while sending the request.
---- System.Net.Http.WinHttpException : The connection with the server was terminated abnormally
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.IntegrationTesting.LoggingHandler.<SendAsync>d__2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at System.Net.Http.HttpClient.<FinishSendAsyncBuffered>d__58.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #898, opened 2018-06-07
Disable tests until product issues are resolved

jkotalik:
Three flaky tests that need to be resolved. Both seem like product issues.

- Synchronous reading and writing fails due to not writing the entire response.
- Logging sometimes doesn't write to stdout of the file.


jkotalik:
For #897 and #766

PR #907, opened 2018-06-11
IIS scenario: stdout and stderr are not present

shhsu:


jkotalik:
dotnet test this

jkotalik:
@dotnet-bot test this please

jkotalik:
@pan-wang when running on IIS, stdout and stderr are set to -1, which causes _dup to return -1. Calling _dup2 with -1 causes an exception rather than returning -1, so we now explicitly check for it being -1.

pan-wang:
when WAS disable stdout and stderr when creating worker process. This is why I said dup2 does not work in the code comments and called another API.

jkotalik:
We don't use dup2 to redirect stdout/err to the file/pipe, we use it to store/restore the previous stdout that was set by IIS. Different scenario, but you are correct that it doesn't work as expected.

pan-wang:
with your new change, does the logging work?

jkotalik:
@shhsu can you verify that locally and see if stdout logs are put in the windows event log if there is an exception?

jkotalik:
Pending verification.

jkotalik:
I can take over this PR. There are few more issues left over with calling dup2 that I didn't realize. @pan-wang was right about calling dup2 I believe. Will revert appropriately.

jkotalik:
Going to revert to original behavior and disable tests for now.

shhsu:
Just to confirm: this is just a manual revert?

jkotalik:
Yeah, we still moved the location of output redirection to another file.

src/AspNetCoreModuleV2/CommonLib/FileOutputManager.cpp
```diff
@@ -147,38 +137,56 @@ FileOutputManager::Start()
...
     }
 
-    // Periodically flush the log content to file
-    m_Timer.InitializeTimer(STTIMER::TimerCallback, &m_struLogFilePath, FILE_FLUSH_TIMEOUT, FILE_FLUSH_TIMEOUT);
+    if (!SetStdHandle(STD_ERROR_HANDLE, m_hLogFileHandle))
+    {
+        hr = HRESULT_FROM_WIN32(GetLastError());
+        return hr;
+    }
+    //int nHandle = _open_osfhandle((intptr_t)m_hLogFileHandle, 0);
```
pan-wang: please clean  those lines  (171-178) if you don't need them

jkotalik:
@shhsu @pan-wang 🆙 📅

shhsu:
LGTM

Issue #899, opened 2018-06-07
Invoke solution build on all native components rather than running msbuild .vcxproj


jkotalik:
Right now, we build each invoke msbuild to build each c++ project, like:

```xml
    <Exec Command="&quot;$(VisualStudioMSBuildx86Path)&quot; &quot;$(RepositoryRoot)test\CommonLibTests\CommonLibTests.vcxproj&quot; /t:Test $(BuildArgs) -p:Platform=%(Platforms.VCPlatform)"
      Condition="'$(VisualStudioMSBuildx86Path)' != ''" />
```

Ideally, this shouldn't be necessary. We should just be able to build the solution, which will build the vcxproj files. I tried this locally, and it continued saying the $(VCTargetsPath) was empty on command line, while in VS it was defined. VCTargetsPath is defined via a reg key. Seems like a bug. 

cc/ @Tratcher 

@pan-wang @shirhatti when porting ANCM to github, was there any reason we invoke msbuilds this way rather than via sln?

shirhatti:
Unless I'm mistaken, it's because invoking sln build via the dotnet cli couldn't build native projects since it didn't have the required targets.

shirhatti:
Let me check, but I think we copied it off the SignalR cpp client.

shirhatti:
IIRC @pranavkm suggested it

muratg:
@shirhatti Could you take a look?

PR #916, opened 2018-06-14
Adds shared request handler lib project

jkotalik:
Part of some cleanup for #899 . It's confusing to know which libraries belong to which of the three dlls (shim, inproc, out of proc). Therefore, all components used the request handlers now live in RHLib, and shared components between the three live in CommonLib.

No code changes at all.


src/AspNetCoreModuleV2/RequestHandlerLib/stdafx.cpp
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
pan-wang: can we remove this stdafx.cpp file as it has no real code. We can use own precompiled header file instead VS default

jkotalik: What benefit does it give us? Doesn't seem like a big deal.

Issue #894, opened 2018-06-06
Package pdbs in Microsoft.AspNetCore.AspNetCoreModule packages


jkotalik:
For debugging purposes.

PR #929, opened 2018-06-15
Adds pdbs (and other files) to testing package

jkotalik:
For #894

### DIFF
--- a/nuget/AspNetCoreV1.nuspec
+++ b/nuget/AspNetCoreV1.nuspec
@@ -19,8 +19,10 @@
     </contentFiles>
   </metadata>
   <files>
-    <file src="src\AspNetCoreModuleV1\AspNetCore\bin\$Configuration$\Win32\**" target="contentFiles\any\any\x86\" />
-    <file src="src\AspNetCoreModuleV1\AspNetCore\bin\$Configuration$\x64\**" target="contentFiles\any\any\x64\" />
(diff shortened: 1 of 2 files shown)


pakrym:
Do we need all the other stuff?

jkotalik:
We don't, I just wanted the quick fix. It's only used for testing anyways.

pakrym:
Let's put just the things we need there.

nuget/AspNetCoreV1.nuspec
```diff
@@ -19,8 +19,8 @@
     </contentFiles>
   </metadata>
   <files>
-    <file src="src\AspNetCoreModuleV1\AspNetCore\bin\$Configuration$\Win32\aspnetcore.dll" target="contentFiles\any\any\x86\aspnetcore.dll" />
-    <file src="src\AspNetCoreModuleV1\AspNetCore\bin\$Configuration$\x64\aspnetcore.dll" target="contentFiles\any\any\x64\aspnetcore.dll" />
+    <file src="src\AspNetCoreModuleV1\AspNetCore\bin\$Configuration$\Win32\**" target="contentFiles\any\any\x86\" />
```
pakrym: here too

jkotalik:
If you are referring to this, then yes: https://github.com/aspnet/IISIntegration/blob/dev/nuget/Microsoft.AspNetCore.AspNetCoreModule.props

PR #930, opened 2018-06-16
Panwang/obcache

pan-wang:
cache the handler object to avoid potential heap fragmenting issue and fix some logging and AV issue


src/AspNetCoreModuleV2/InProcessRequestHandler/inprocesshandler.cpp
```diff
@@ -59,6 +61,80 @@ IN_PROCESS_HANDLER::OnExecuteRequestHandler()
...
+        }
+
+        hr = pAlloc->Initialize(sizeof(IN_PROCESS_HANDLER),
+            64); // nThreshold
+        if (FAILED(hr))
+        {
+            goto Failure;
+        }
+
+        AcquireSRWLockExclusive(&g_srwLockRH);
```
pakrym: Why the lock?

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocesshandler.cpp
```diff
@@ -59,6 +61,65 @@ IN_PROCESS_HANDLER::OnExecuteRequestHandler()
...
+    DBG_ASSERT(sm_pAlloc != NULL);
+    if (sm_pAlloc != NULL)
+    {
+        sm_pAlloc->Free(pMemory);
+    }
+}
+
+// static
+HRESULT
+IN_PROCESS_HANDLER::StaticInitialize( VOID )
```
jkotalik: I don't like the static initialize pattern; it's easy to forget calling it. Calling it in dll main also seems odd. Can we put this in the inprocess application?

src/AspNetCoreModuleV2/OutOfProcessRequestHandler/outofprocess/forwardinghandler.h
```diff
@@ -72,6 +72,10 @@ class FORWARDING_HANDLER : public REQUEST_HANDLER
         bool    fClientInitiated
     );
 
+    static void * operator new(size_t size);
```
jkotalik: override?

pakrym:
I have another idea to address startup failures, let's create another IAPPLICATION implementation that would return 502 and return it in startup error cases, this would allow us to move even more code to handlers. Also in this case restart just works, because when configuration would change application would get restarted without requiring additional logic.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -659,17 +659,20 @@ IN_PROCESS_APPLICATION::LogErrorsOnMainExit(
...
-            ASPNETCORE_EVENT_INPROCESS_THREAD_EXIT_STDOUT_MSG,
-            m_pConfig->QueryApplicationPath()->QueryStr(),
-            m_pConfig->QueryApplicationPhysicalPath()->QueryStr(),
-            hr,
-            struStdErrOutput.QueryStr());
+    STRA straStdErrOutput;
+    STRU struStdMsg;
+    if (m_pLoggerProvider->GetStdOutContent(&straStdErrOutput))
+    {
+        if (SUCCEEDED(struStdMsg.CopyA(straStdErrOutput.QueryStr()))) {
```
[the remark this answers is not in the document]
pakrym: Why is stdout content CHAR*?

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocesshandler.cpp
```diff
@@ -59,6 +61,65 @@ IN_PROCESS_HANDLER::OnExecuteRequestHandler()
     return m_pApplication->OnExecuteRequest(m_pW3Context, this);
 }
 
+// static
+void * IN_PROCESS_HANDLER::operator new(size_t)
+{
+    DBG_ASSERT(sm_pAlloc != NULL);
+    if (sm_pAlloc == NULL)
+    {
+        return NULL;
```
pakrym: Not a fan of returning null, I'd like to stop to check allocations for NULL every time. Can you throw bad_alloc here please?

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocesshandler.cpp
```diff
@@ -59,6 +61,65 @@ IN_PROCESS_HANDLER::OnExecuteRequestHandler()
...
+    HRESULT                         hr = S_OK;
+
+    sm_pAlloc = new ALLOC_CACHE_HANDLER;
+    if (sm_pAlloc == NULL)
+    {
+        hr = E_OUTOFMEMORY;
+        goto Finished;
+    }
+
+    hr = sm_pAlloc->Initialize(sizeof(IN_PROCESS_HANDLER),
```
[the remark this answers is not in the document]
pan-wang: yes

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocesshandler.cpp
pan-wang: the sm_pAlloc is a static variable shared cross applications in the same worker process. We have to call some static initialize function to init it.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocesshandler.cpp
pan-wang: yes, you can to try-catch on bad_alloc exception.

pakrym: Not, what I meant is can you throw bad_alloc if sm_pAlloc  is null, instead of returning null?

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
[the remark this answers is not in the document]
pan-wang: i don't want to touch the original design for pipeoutputmanage which passed a char buffer to receive the output. We can change the buffere to wchar.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -246,6 +252,20 @@ APPLICATION_INFO::EnsureApplicationCreated(
...
+    {
+        // Log the failure and update application info to not try again
+        UTILITY::LogEventF(g_hEventLog,
+            EVENTLOG_ERROR_TYPE,
+            ASPNETCORE_EVENT_ADD_APPLICATION_ERROR,
+            ASPNETCORE_EVENT_ADD_APPLICATION_ERROR_MSG,
+            pHttpContext->GetApplication()->GetApplicationId(),
+            hr);
+
+        goto Finished;
```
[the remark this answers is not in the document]
pan-wang: fixed

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
```diff
@@ -150,6 +156,8 @@ class APPLICATION_INFO
     STRU                    m_struInfoKey;
     BOOL                    m_fAppOfflineFound;
     BOOL                    m_fAllowStart; // Flag indicates whether there is (configuration) error blocking application from starting
+    BOOL                    m_fAppCreated; // Flag indicates if the application has been created successfully (runtime error).
```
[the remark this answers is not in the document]
pan-wang: yes, we can resue APPLICATION_STATUS

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocesshandler.cpp
pan-wang: for your comments about moving check createapplication failure to requesthandler. it is possible. It requires change for inprocess. Will do it with moving app_offline

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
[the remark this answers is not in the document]
jkotalik: That's not what I intended to ask. I just think if we are allowed to start, it means the app is created.

src/AspNetCoreModuleV2/AspNetCore/src/applicationmanager.cxx
```diff
@@ -254,25 +253,9 @@ APPLICATION_MANAGER::RecycleApplicationFromManager(
...
-            delete m_pApplicationInfoHash;
-
-            // We don't want to delete the table as m_pApplicationInfoHash = table
-            fKeepTable = TRUE;
-            m_pApplicationInfoHash = table;
-        }
-    }
-
-    if (m_pApplicationInfoHash->Count() == 0)
+    if (m_pApplicationInfoHash->Count() == 0 && m_hostingModel == HOSTING_OUT_PROCESS)
```
jkotalik: Why was this changed?

Issue #868, opened 2018-05-25
Refactor native logging infrastructure


pakrym:
We have 4 copies of `debugutil.h` and the only way to enable logging is to set a registry key. We need to get to a better place so we can add more logging and simplify CI failure debugging, having only HRESULT is not enough.

PR #932, opened 2018-06-18
Cleanup debug logging

pakrym:
https://github.com/aspnet/IISIntegration/issues/868

1. Remove unneeded pudebug.h
2. Unify initialization logic
3. Allow logging to be enabled by the environment variable
4. Set it everywhere


src/AspNetCoreModuleV2/CommonLib/debugutil.h
```diff
@@ -2,19 +2,75 @@
...
 
 #pragma once
-#define ASPNETCORE_DEBUG_FLAG_INFO          0x00000001
-#define ASPNETCORE_DEBUG_FLAG_WARNING       0x00000002
-#define ASPNETCORE_DEBUG_FLAG_ERROR         0x00000004
+#include "stdafx.h"
+#include "dbgutil.h"
 
-extern DWORD g_dwAspNetCoreDebugFlags;
+#define ASPNETCORE_DEBUG_FLAG_INFO          DEBUG_FLAG_INFO
```
jkotalik: Check if there are duplicate debugutil.h files anywhere.

pakrym: Nope, I was wrong on that one.

pakrym:
DBG_ASSERT was always coming from ntassert.h, additionally they are never enabled even in debug build but that's fix for another PR.

src/AspNetCoreModuleV2/CommonLib/debugutil.h
```diff
@@ -2,19 +2,75 @@
...
+#define ASPNETCORE_DEBUG_FLAG_WARNING       DEBUG_FLAG_WARN
+#define ASPNETCORE_DEBUG_FLAG_ERROR         DEBUG_FLAG_ERROR
+#define ASPNETCORE_DEBUG_FLAG_CONSOLE       0x00000008
```
pan-wang: w3wp does not have console

src/AspNetCoreModuleV2/CommonLib/debugutil.h
```diff
@@ -27,23 +83,24 @@ DebugPrint(
...
+            return;
         }
 
         OutputDebugStringA( strOutput.QueryStr() );
-    }
-
-Finished:
 
-    return;
+        if (IsEnabled(ASPNETCORE_DEBUG_FLAG_CONSOLE))
```
pan-wang: We did not do console log as w3wp does not have a console at all. For developer scenario, writing to debug context is sufficient enough. Instead of writing to console, we should consider writing to ETW log.

pakrym: I want to see logs from functional tests, the easiest way to get them is to log to console.

src/AspNetCoreModuleV2/CommonLib/debugutil.h
pakrym: Sure, enabling this flag there would not do anything.

src/AspNetCoreModuleV2/CommonLib/debugutil.h
pakrym: I would also like to be able to get logs into stdout log file so we can diagnose issues when running in deployed environments. But this is a task for the future.

src/AspNetCoreModuleV2/CommonLib/debugutil.h
jkotalik: Can you verify logs are debug logs are printed for both IIS and IISExpress to a file?

pakrym: Not really my intent at this point. We enable stdout capture for small periods of time, it won't capture all the logs. The intent here is to be able to see logs from functional tests first.

pakrym:
Added tests and a wide char log method

pakrym:
@pan-wang ping

src/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
```diff
@@ -1,241 +1,242 @@
...
-      <Outputs Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">%(Filename).rc;%(Filename).h;MSG0409.bin</Outputs>
-      <Command Condition="'$(Configuration)|$(Platform)'=='Release|x64'">mc %(FullPath)</Command>
-      <Message Condition="'$(Configuration)|$(Platform)'=='Release|x64'">Compiling Event Messages ...</Message>
-      <Outputs Condition="'$(Configuration)|$(Platform)'=='Release|x64'">%(Filename).rc;%(Filename).h;MSG0409.bin</Outputs>
-    </CustomBuild>
-  </ItemGroup>
-  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-  <ImportGroup Label="ExtensionTargets">
-  </ImportGroup>
+<?xml version="1.0" encoding="utf-8"?>
```
jkotalik: Remove the large diff here?

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+#include "debugutil.h"
+#include "dbgutil.h"
+
+inline HANDLE g_hStandardOutput;
+
+VOID
+DebugInitialize()
+{
+    g_hStandardOutput = GetStdHandle(STD_OUTPUT_HANDLE);
```
jkotalik: Check this behavior in full IIS. This may be an invalid handle on IIS. I'm concerned that this will act weirdly with setting stdout a file later.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        RegCloseKey(hKey);
+    }
+
+    // We expect single digit value and a null char
+    const size_t environmentVariableValueSize = 2;
+    std::wstring environmentVariableValue(environmentVariableValueSize, '\0');
+
+    if (GetEnvironmentVariable(L"ASPNETCORE_MODULE_DEBUG", environmentVariableValue.data(), environmentVariableValueSize) == environmentVariableValueSize - 1)
```
jkotalik: nit formatting.

jkotalik: Also, are .data() and &env[0] the same?

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    return ( dwFlag & DEBUG_FLAGS_VAR );
+}
+
+VOID
+DebugPrint(
+    DWORD   dwFlag,
+    const LPCSTR  szString
+    )
+{
```
jkotalik: Add WDebugPrint.

src/AspNetCoreModuleV2/CommonLib/debugutil.h
jkotalik: I'm just concerned it will cause exceptions/AVs in w3wp.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
pakrym: No point, if you are doing no formatting why are you logging wide string?

jkotalik: tru.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
pakrym: Yes, everyone was using `[0]` because `data` was const until recently

pakrym:
Verified, works fine in full IIS with env var.

Issue #556, opened 2018-01-25
New error page for inprocess application start failure


pan-wang:
For OutOfProcess scenario, if ANCM failed to start the backend process, it will sent a special 502.5 page to the client with some diagnose hints,  if disableStartUpErrorPage is configured as FALSE (default). We should consider a similar page for inprocess scenario to help user debug the failure.

muratg:
@shirhatti Could you check with @blowdart?

blowdart:
I'm going to need to know what is in the page, what circumstances send it, and if it's available to anyone on a crash, or just, say, to people accessing localhost

Tratcher:
![image](https://user-images.githubusercontent.com/1821173/35832181-1f274bda-0a81-11e8-9734-fffaf83fe392.png)

blowdart:
Oh just static? With nothing else? That's fine then.

muratg:
This is for cases where the app failed to load. Otherwise, if the app is working, Developer Exception Page could be used instead.

blowdart:
Oh, I missed "today".

OK, so what information, and under what circumstances? Would the environment affect the information displayed?

shirhatti:
We're still designing this. Thinking out loud we want to control it via [web.config](https://github.com/aspnet/IISIntegration/blob/dev/src/AspNetCore/aspnetcore_schema.xml#L27) or an env variable.

Tratcher:
What's changed from when we designed this a few years ago and decided a static page was the only possible safe content?

shirhatti:
Think developer exception page, but for startup failures instead.

Off by default, but makes it way easier to diagnose startup failures than having to look at event logs and redirecting stdout logs.

Tratcher:
So the equivalent of ASPNETCORE_CAPTURESTARTUPERRORS and ASPNETCORE_DETAILEDERRORS in Hosting? Better use the same settings.

PR #934, opened 2018-06-18
Return static 500.30 file on startup fails for in-process

jkotalik:
For #556 .
There was a duplicated check for validating the status of the application in the shim. I removed this as the application should be the one handling setting the http message and status.
This does create a drawback of us needing to allocate an INPROCESS_HANDLER before returning the error page.


src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -29,6 +29,7 @@ IN_PROCESS_APPLICATION::IN_PROCESS_APPLICATION(
 
     InitializeSRWLock(&m_srwLock);
     m_pConfig = pConfig;
+    html500Page = GetStaticHtml500Content();
```
mikeharder: Why store the error HTML in a member, rather than calling `GetStaticHtml500Content()` every time the error HTML is needed?  If it's for perf, this seems unnecessary since the HTML is only used in error cases.

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
```diff
@@ -152,16 +152,6 @@ ASPNET_CORE_PROXY_MODULE::OnExecuteRequestHandler(
 
     m_pApplicationInfo->ExtractApplication(&pApplication);
 
-    // make sure application is in running state
```
pakrym: Why was this removed?

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocesshandler.cpp
```diff
@@ -30,18 +30,28 @@ IN_PROCESS_HANDLER::OnExecuteRequestHandler()
 
     if (FAILED(hr))
     {
-        // TODO remove com_error?
-        /*_com_error err(hr);
-        if (ANCMEvents::ANCM_START_APPLICATION_FAIL::IsEnabled(m_pW3Context->GetTraceContext()))
+        if (!m_pApplication->QueryConfig()->QueryDisableStartUpErrorPage())
```
pakrym: Can we return a different IAPPLICATION instance if app failed to start?

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
jkotalik: We shouldn't be checking the status of the application in the shim. This caused an IIS 500 to be returned instead of ours.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocesshandler.cpp
jkotalik: Not without some refactoring. Here is the current order of operations.
- CreateApplication
- CreateHandler
- OnExecuteRequestHandler
- LoadApplication

I think we need to move LoadApplication before creating the handler. I'll take a quick stab at it and see how hard it is.

jkotalik: Will be fairly difficult. Let's not do this in this PR IMO.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
mikeharder: I still wouldn't bother caching this but I agree there's not a big difference either way.

jkotalik:
🆙 📅 . Instead of returning the exception in the inprocess application, we create a new application that constantly returns 500.30s.

pan-wang:
Can we fix the flaky CI  build /tests?   A lot of time,  I don't know the Whether  PR had issue or not and should I look at a PR which has CI failure.

jkotalik:
This is my failure. Investigating.

pakrym:
> Can we fix the flaky CI build /tests? A lot of time, I don't know the Whether PR had issue or not and should I look at a PR which has CI failure.

Tests don't seem to flaky, it's just PRs are breaking them.

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
pakrym: Should we bring this back since we return IAPPLCIATION instance now?

src/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include "StartupExceptionApplication.h"
+
+VOID StartupExceptionApplication::ShutDown()
+{
+    return VOID();
+}
+
+VOID StartupExceptionApplication::Recycle()
+{
+    return VOID();
```
pakrym: Are you sure this should be empty?

test/IISIntegration.FunctionalTests/Inprocess/StartupExceptionTests.cs
```diff
@@ -50,5 +50,20 @@ public class StartupExceptionTests : IISFunctionalTestBase
 
             Assert.Contains(TestSink.Writes, context => context.Message.Contains(new string('a', 4096)));
         }
+
+
+        [Fact]
+        public async Task Gets500_30_ErrorPage()
+        {
+            var deploymentParameters = Helpers.GetBaseDeploymentParameters("StartupExceptionWebsite");
+            deploymentParameters.PublishApplicationBeforeDeployment = true;
```
pakrym: why?

mikeharder:
@pakrym, @pan-wang: We have seen some flaky tests in IISIntegration: https://github.com/aspnet/iisintegration/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+flaky.  We do need to fix them eventually but they may not currently be the highest priority.

src/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.cpp
jkotalik: Good point. Let me think about it.

pan-wang: I am sure about the value of this change. We introduce multiple classes just for handling one error page. Why not just a simple UTILITY method?

jkotalik: If we want to update the error page in the future, we would need to modify the shim. If we want to use a utility function, we will effectively reuse the inprocess application loaded to just return 500s.

jkotalik:
@dotnet-bot test this please

src/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.cpp
```diff
@@ -0,0 +1,31 @@
...
+    {
+        // IIS scenario.
+        // notify IIS first so that new request will be routed to new worker process
+        m_pHttpServer->RecycleProcess(L"AspNetCore InProcess Recycle Process on Demand");
+    }
+    else
+    {
+        // IISExpress scenario
+        // Shutdown the managed application and call exit to terminate current process
+        exit(0);
```
pakrym: did you copy it from somewhere? We should avoid doing exit(0) and allow IIS to gracefully stop.

jkotalik: Yeah we shouldn't need it at all.

jkotalik:
Current code path works fine on IISExpress and IIS during recycle.

jkotalik:
I'll check appoffline scenarios too.

jkotalik:
There seems to be larger issues that should be handled in other PRs, including #950 and #949 . This handles all scenarios we expect it to.

src/AspNetCoreModuleV2/InProcessRequestHandler/InProcessApplicationBase.h
```diff
@@ -0,0 +1,24 @@
...
+{
+public:
+
+    InProcessApplicationBase(IHttpServer* pHttpServer);
+
+    ~InProcessApplicationBase() = default;
+
+    __override VOID Recycle(VOID);
+
+    BOOL m_fRecycleCalled;
```
pakrym: private

src/AspNetCoreModuleV2/InProcessRequestHandler/InProcessApplicationBase.h
```diff
@@ -0,0 +1,24 @@
...
+
+class InProcessApplicationBase : public APPLICATION
+{
+public:
+
+    InProcessApplicationBase(IHttpServer* pHttpServer);
+
+    ~InProcessApplicationBase() = default;
+
+    __override VOID Recycle(VOID);
```
pakrym: override

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -104,11 +107,17 @@ CreateApplication(
         {
             goto Finished;
         }
+        disableStartupPage = pConfig->QueryDisableStartUpErrorPage();
 
         pApplication = new IN_PROCESS_APPLICATION(pServer, pConfig);
-
         pConfig = NULL;
 
+        if (FAILED(pApplication->LoadManagedApplication()))
```
pakrym: This would cause a double free because of  IN_PROCESS_APPLICATION unique_ptr

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -90,46 +93,27 @@ CreateApplication(
...
-
-Finished:
-    if (FAILED(hr))
-    {
-        if (pApplication != NULL)
+        if (FAILED(pApplication->LoadManagedApplication()))
         {
-            delete pApplication;
-            pApplication = NULL;
+            *ppApplication = new StartupExceptionApplication(pServer, disableStartupPage);
```
pakrym:
```
pApplication  = make_unique<StartupExceptionApplication>();
```

and then `*ppApplication = pApplication.release();` after if

jkotalik: Can't do that exactly as pApplication is of type `std::unique_ptr<InProcessApplication>`. Would you prefer I don't call new?

jkotalik: Need to call `LoadManagedApplication()` on the InProcessApplication.

test/IISIntegration.FunctionalTests/Inprocess/StartupExceptionTests.cs
```diff
@@ -54,5 +53,19 @@ public class StartupExceptionTests : IISFunctionalTestBase
...
+
+        [Fact]
+        public async Task Gets500_30_ErrorPage()
+        {
+            var deploymentParameters = Helpers.GetBaseDeploymentParameters("StartupExceptionWebsite");
+
+            var deploymentResult = await DeployAsync(deploymentParameters);
+
+            var response = await deploymentResult.HttpClient.GetAsync("/");
+            Assert.False(response.IsSuccessStatusCode);
```
pakrym: Assert err code

PR #937, opened 2018-06-19
Reenable precompiled headers in AspNetCore project

pakrym:
We need precomp.c because it causes precompiled header generation.

Additionally, we need PrecompiledHeader property to be set to Create for it.

And `#include "precomp.h"` should be the first line of every cpp file, putting it into .h was my mistake.


pakrym:
CI Failure is unrelated https://github.com/aspnet/IISIntegration/issues/939

jkotalik:
We have a mixed model of precompiled headers throughout our projects. @pan-wang you mentioned in [this PR](https://github.com/aspnet/IISIntegration/pull/916) that you wanted us to use our own precompiled headers. By doing this change, we are going back to VS style of precompiled headers. 

I currently don't see and precompiled headers at all on a default build of our stuff, so I think we have to go with the VS style. @pakrym you probably should change all other projects to use VS precompiled headers.

pakrym:
Is there another way of doing precompiled headers? It's a compiler feature.

pakrym:
Ping @pan-wang  @jkotalik

src/AspNetCoreModuleV2/AspNetCore/src/precomp.cpp
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
jkotalik: I think you need to include precomp.h in this file.

src/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
```diff
@@ -100,6 +100,7 @@
       <AdditionalIncludeDirectories>..\iislib;</AdditionalIncludeDirectories>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <LanguageStandard>stdcpp17</LanguageStandard>
+      <ForcedIncludeFiles>stdafx.h</ForcedIncludeFiles>
```
jkotalik: Why are you force including these?

src/AspNetCoreModuleV2/CommonLib/SRWExclusiveLock.h
```diff
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <synchapi.h>
```
jkotalik: ?

jkotalik:
Do the RequestHandler projects use the same style of precompiled headers?

src/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
pakrym: So I don't have to include in every cpp file.

src/AspNetCoreModuleV2/AspNetCore/src/precomp.cpp
pakrym: It is included by `ForcedIncludeFiles`.

src/AspNetCoreModuleV2/CommonLib/SRWExclusiveLock.h
pakrym: `SRWLOCK``

pakrym:
> Do the RequestHandler projects use the same style of precompiled headers?

Not yet. I'll do it, just not in this PR.

jkotalik:
PR is breaking things in dev.

PR #940, opened 2018-06-19
Dispose server before checking test sink

jkotalik:
There has been some flakiness with checking the test sink's output, like here: https://ci3.dot.net/job/aspnet_IISIntegration/job/dev/job/windows-Configuration_Release_prtest/159/console. We may need to add a timeout here, but one mitigation we can add is disposing the iisexpress process, which should force stdout to be written.

### DIFF
--- a/test/IISIntegration.FunctionalTests/Inprocess/StartupExceptionTests.cs
+++ b/test/IISIntegration.FunctionalTests/Inprocess/StartupExceptionTests.cs
@@ -29,6 +29,8 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
 
             Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
 
+            Dispose();
+
             Assert.Contains(TestSink.Writes, context => context.Message.Contains($"Random number: {randomNumberString}"));
         }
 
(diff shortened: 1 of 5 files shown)


test/IISIntegration.FunctionalTests/Inprocess/StartupExceptionTests.cs
```diff
@@ -29,6 +29,8 @@ public class StartupExceptionTests : IISFunctionalTestBase
 
             Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
 
+            Dispose();
```
pakrym: Add `StopServer` method to a base class that would dispose deployer and set it to null. Not a fan of disposing `this` and it would prevent double dispose.

test/IISIntegration.FunctionalTests/Utilities/FunctionalTestsBase.cs
```diff
@@ -28,5 +28,10 @@ public override void Dispose()
         {
             _deployer?.Dispose();
         }
+
+        public void StopServer()
+        {
+            _deployer?.Dispose();
```
pakrym: set to null and call this method from dispose.

PR #944, opened 2018-06-20
Create a separate sln for V1

jkotalik:
Part of cleanup for #899. Searching through code with ANCMV1 and V2 in it is painful. This will make searching easier and keep V1 and V2 separate.


pakrym:
I though about this recently and think we need sln file without V2 (like MVC has MVCNoFun ) but the main one should have everything in it.

jkotalik:
How does Mvc make it so build.cmd only invokes build on the main solution?

pakrym:
Isn't that what korebuild is doing by default?

jkotalik:
@pakrym I updated this to create a .NoV1 sln.

pakrym:
What's our guidance on updating this? Make changes to the main one then remove a project and overwrite NoV1?

jkotalik:
Yeah I guess that's the best move.

pakrym:
Can we add a `dotnet sln` based script to regenerate this?

jkotalik:
@pranavkm does MVC have a script to update the NoFun sln? From what I can tell, there isn't a way.

pranavkm:
Nope, it's manually curated.

Issue #950, opened 2018-06-20
App_offline current does not work


jkotalik:
Tried dropping appoffline when site was up and the file change notification was not picked up. Reproed in both IIS and IISExpress.

muratg:
This is in V2, right?

jkotalik:
yes, I figured out the issue. Will make a PR shortly.

PR #953, opened 2018-06-21
Fix App_offline and start adding tests

jkotalik:
Will work on adding a few more test cases. For #950


pakrym:
Make sure this change doesn't cause process exit for inprocess TestServer

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        [ConditionalFact]
+        public async Task AppOfflineDropped_CanRemoveAppOfflineAfterAdding()
+        {
+            var deploymentResult = await StartSite();
+
+            AddAppOffline(deploymentResult.DeploymentResult.ContentRoot);
+
+            var response = await deploymentResult.RetryingHttpClient.GetAsync("HelloWorld");
+
+            while (response.IsSuccessStatusCode)
```
pakrym: Possible infinite loop./

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            var fs = File.Create(Path.Combine(appPath, "app_offline.htm"));
+            // Close the file to allow for 
+            fs.Close();
+        }
+
+        private void RemoveAppOffline(string appPath)
+        {
+
+            RetryHelper.RetryOperation(
```
pakrym: Do we lock appofline?

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            var response = await deploymentResult.RetryingHttpClient.GetAsync("HelloWorld");
+
+            var responseText = await response.Content.ReadAsStringAsync();
+            Assert.Equal("Hello World", responseText);
+            return deploymentResult;
+        }
+
+        private void AddAppOffline(string appPath)
+        {
+            var fs = File.Create(Path.Combine(appPath, "app_offline.htm"));
```
pakrym: File.WriteAllText?

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            AddAppOffline(deploymentResult.DeploymentResult.ContentRoot);
+            var response = await deploymentResult.RetryingHttpClient.GetAsync("HelloWorld");
+
+            while (response.IsSuccessStatusCode)
+            {
+                // Keep retrying until app_offline is present.
+                response = await deploymentResult.RetryingHttpClient.GetAsync("HelloWorld");
+            }
+
+            Assert.Equal(HttpStatusCode.ServiceUnavailable, response.StatusCode);
```
pakrym: Verify that  app_offline is served.

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            RemoveAppOffline(deploymentResult.DeploymentResult.ContentRoot);
+        }
+
+        private async Task<IISDeploymentResult> StartSite()
+        {
+            var deploymentParameters = Helpers.GetBaseDeploymentParameters();
+
+            var deploymentResult = await DeployAsync(deploymentParameters);
+
+            var response = await deploymentResult.RetryingHttpClient.GetAsync("HelloWorld");
```
pakrym: Too many hello worlds around. Let's add VerifyStarted helper and use it everywhere.

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
jkotalik: I think we do to read the contents of the file.

pakrym: Right, but we should set FILE_SHARE_READ | WRITE | DELETE while doing so.

jkotalik:
🆙 📅 . I would also say app_offline is not really a testable scenario for the hostable webcore skew as it requires dropping app_offline in a bin local folder.

pakrym:
I'm not worried about testing app_ofline on HWC, I'm worried that calling HWCShutdown which we do in test server would kill the process.

jkotalik:
This wouldn't hit shutdown for HWC because exit is only called from recycle, not shutdown.

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        private async Task AssertAppOffline(IISDeploymentResult deploymentResult, string expectedResponse = "")
+        {
+            var response = await deploymentResult.RetryingHttpClient.GetAsync("HelloWorld");
+
+            for (var i = 0; response.IsSuccessStatusCode && i < 5; i++)
+            {
+                // Keep retrying until app_offline is present.
+                response = await deploymentResult.RetryingHttpClient.GetAsync("HelloWorld");
```
pakrym: The client is already retrying, do you need to retry even more?
How long does it take?

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
```diff
@@ -0,0 +1,130 @@
...
+                var response = await deploymentResult.RetryingHttpClient.GetAsync("HelloWorld");
+            }
+            catch (HttpRequestException)
+            {
+                // dropping app_offline will kill the process
+            }
+
+            var hostShutdownToken = deploymentResult.DeploymentResult.HostShutdownToken;
+
+            hostShutdownToken.WaitHandle.WaitOne(millisecondsTimeout: 1000);
```
pakrym: Assert.True

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
```diff
@@ -0,0 +1,130 @@
...
+{
+    [SkipIfIISExpressSchemaMissingInProcess]
+    public class AppOfflineTests : IISFunctionalTestBase
+    {
+        [ConditionalFact]
+        public async Task AppOfflineDroppedWhileSiteIsDown_SiteReturns503()
+        {
+            var deploymentParameters = Helpers.GetBaseDeploymentParameters();
+
+            var deploymentResult = await DeployAsync(deploymentParameters);
```
pakrym: I wish there was a way to assert that site is no started yet.

pakrym:
Add a test where you delete app_offline and the start the app.

pakrym:
What about outofproc?

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
jkotalik: yeah no need for retrying.

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
jkotalik: Yeah it would be a bit tricky.

jkotalik:
I'll file an issue for out of proc once this is merged.

test/IISIntegration.FunctionalTests/Inprocess/AppOfflineTests.cs
```diff
@@ -67,6 +67,18 @@ public class AppOfflineTests : IISFunctionalTestBase
             RemoveAppOffline(deploymentResult.DeploymentResult.ContentRoot);
         }
 
+        [ConditionalFact]
+        public async Task AddRemoveAppOffline_AppStarts()
+        {
+            var deploymentResult = await AssertStarts();
+
+            AddAppOffline(deploymentResult.DeploymentResult.ContentRoot);
+
```
pakrym: Make sure app starts here and appofline is served

pakrym:
> This wouldn't hit shutdown for HWC because exit is only called from recycle, not shutdown.

But Recycle is called from ExecuteApplication after main exits.

PR #954, opened 2018-06-21
Add ability to print debug logs to a file

pakrym:
By setting ASPNETCORE_MODULE_DEBUG_FILE variable.


src/AspNetCoreModuleV2/CommonLib/Environment.cpp
```diff
@@ -0,0 +1,29 @@
+#include "Environment.h"
```
jkotalik: copyright.

src/AspNetCoreModuleV2/CommonLib/Environment.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
```
jkotalik: copyright.

src/AspNetCoreModuleV2/CommonLib/Environment.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include <string>
+
+class Environment
+{
+public:
+	Environment() = delete;
```
jkotalik: formatting.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -44,21 +46,54 @@ DebugInitialize()
...
+
+        if (!debugOutputFile.empty())
         {
-            // ignore
+            g_logFile = CreateFileW(debugOutputFile.c_str(),
+                FILE_GENERIC_WRITE,
+                FILE_SHARE_READ | FILE_SHARE_WRITE,
+                nullptr,
+                OPEN_ALWAYS,
+                FILE_ATTRIBUTE_NORMAL,
```
jkotalik: FILE_APPEND_DATA so you don't need to always call SetFilePointer?

pakrym: FILE_APPEND_DATA  is part of FILE_GENERIC_WRITE you still need to set it to append to existing file.

jkotalik: Then can you remove the SetFilePointer?

pakrym: It's required to append to an existing file, see https://msdn.microsoft.com/en-us/library/windows/desktop/aa363778(v=vs.85).aspx

jkotalik: I seeeeeeeeeeeeeeeeeeeeeeeeeeeeee.

src/AspNetCoreModuleV2/CommonLib/Environment.h
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+#pragma once
```
pan-wang: why not put this one into UTILITY. You only have some static member method

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
pan-wang: you allow shared write which could cause file corrupt. should we just use env  variable for debug file folder and then create a temp file for each process instead of mixing the log from different processes. This will also help user to easily parse the log

pakrym: We would have to create multiple files per process then, because inproc and shim initialize their own logging, allowing shared writes allows to reuse the same file.

pan-wang: you also need to make sure DebugInitialize will only initialize once. Otherwise, g_logFile will be reassigned and lead to handle leak.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -46,21 +49,54 @@ DebugInitialize()
...
+    catch (...)
+    {
+        // ignore
+    }
+}
+
+VOID
+DebugStop()
+{
+    if (IsEnabled(ASPNETCORE_DEBUG_FLAG_FILE))
```
pan-wang: check whether the handle is InvalidHAndle instead of the flag, as the CreateFile may fail.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
pakrym: I'll move it into DLL_ATTACH

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -93,11 +124,17 @@ DebugPrint(
...
         if (IsEnabled(ASPNETCORE_DEBUG_FLAG_CONSOLE))
         {
-            DWORD nBytesWritten = 0;
-            WriteFile(g_hStandardOutput, strOutput.QueryStr(), strOutput.QueryCB(), &nBytesWritten, NULL);
+            WriteFile(g_hStandardOutput, strOutput.QueryStr(), strOutput.QueryCB(), &nBytesWritten, nullptr);
+        }
+
+        if (g_logFile != INVALID_HANDLE_VALUE)
+        {
+            SetFilePointer(g_logFile, 0, nullptr, FILE_END);
```
pan-wang: what would happen if there are multiple threads calls DebugPrint simultaneously?

pan-wang:
how to consume this log file in AppService?

pakrym:
Set environment variable to `%HOME%\LogFiles\Application\file`  and use streaming logs pane.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
pakrym: Added a lock

pakrym:
Ping, anything else you want me to address?

jkotalik:
I think there are a few copyright headers you are missing. I'll take another glance over now.

src/AspNetCoreModuleV2/CommonLib/Environment.cpp
```diff
@@ -0,0 +1,29 @@
...
+std::wstring
+Environment::ExpandEnvironmentVariables(const std::wstring & str)
+{
+    DWORD requestedSize = ExpandEnvironmentStringsW(str.c_str(), nullptr, 0);
+    if (requestedSize == 0)
+    {
+        throw std::system_error(GetLastError(), std::system_category(), "ExpandEnvironmentVariables");
+    }
+
+    std::wstring expandedStr;
```
jkotalik: I've seen us use std::unique_ptr<wchar[]> and std::wstring interchangeably for calling windows functions. Let's pick one and stick with it?

src/AspNetCoreModuleV2/AspNetCore/src/dllmain.cpp
```diff
@@ -36,6 +36,8 @@ StaticCleanup()
         DeregisterEventSource(g_hEventLog);
         g_hEventLog = NULL;
     }
+
+    DebugStop();
```
jkotalik: Why not put this in DllDetatch?

jkotalik: Nvrm, this is only called from DllDetatch 😃

src/AspNetCoreModuleV2/CommonLib/Environment.cpp
pakrym: wchar[] is used only when it gets turned into STRU afterwards, otherwise wstring should be used.

pakrym:
Fixed.

PR #957, opened 2018-06-21
Fixing build issues

jkotalik:
build failed on CI.


pakrym:
How didn't we catch it?

jkotalik:
There must be some build difference between CI and appveyor/jenkins. I pulled it and it failed to build locally too.

PR #958, opened 2018-06-21
Cleanup where.exe invocation

pakrym:
And add more logs

### DIFF
--- a/src/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
+++ b/src/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
@@ -197,6 +197,7 @@
     <ClInclude Include="exceptions.h" />
     <ClInclude Include="GlobalVersionUtility.h" />
     <ClInclude Include="fx_ver.h" />
+    <ClInclude Include="HandleWrapper.h" />
     <ClInclude Include="hostfxroptions.h" />
     <ClInclude Include="hostfxr_utility.h" />
     <ClInclude Include="iapplication.h" />
--- a/src/AspNetCoreModuleV2/CommonLib/HandleWrapper.h
+++ b/src/AspNetCoreModuleV2/CommonLib/HandleWrapper.h
@@ -1,0 +1,52 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#pragma once
+
+#include <Windows.h>
+
+struct InvalidHandleTraits
+{
+    using HandleType = HANDLE;
+    static const HANDLE DefaultHandle;
+    static void Close(HANDLE handle) { CloseHandle(handle); }
+};
+
+// Workaround for VC++ bug https://developercommunity.visualstudio.com/content/problem/33928/constexpr-failing-on-nullptr-v141-compiler-regress.html
+const HANDLE InvalidHandleTraits::DefaultHandle = INVALID_HANDLE_VALUE;
+
+struct NullHandleTraits
+{
+    using HandleType = HANDLE;
+    static constexpr HANDLE DefaultHandle = NULL;
+    static void Close(HANDLE handle) { CloseHandle(handle); }
+};
+
+template<typename traits>
+class HandleWrapper
+{
+public:
+    using HandleType = typename traits::HandleType;
+
+    HandleWrapper(HandleType handle = traits::DefaultHandle) : m_handle(handle) { }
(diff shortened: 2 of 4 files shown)


src/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
```diff
@@ -197,6 +197,7 @@
     <ClInclude Include="exceptions.h" />
     <ClInclude Include="GlobalVersionUtility.h" />
     <ClInclude Include="fx_ver.h" />
+    <ClInclude Include="HandleWrapper.h" />
```
jkotalik: Delete the cpp file

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -440,9 +441,11 @@ HOSTFXR_UTILITY::InvokeWhereToFindDotnet()
     SECURITY_ATTRIBUTES securityAttributes;
 
     CHAR                pzFileContents[READ_BUFFER_SIZE];
-    HANDLE              hStdOutReadPipe = INVALID_HANDLE_VALUE;
-    HANDLE              hStdOutWritePipe = INVALID_HANDLE_VALUE;
-    LPWSTR              pwzDotnetName = NULL;
+    HandleWrapper<>     hStdOutReadPipe;
```
jkotalik: Weird that you have to have empty generics here.

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -584,53 +562,37 @@ HOSTFXR_UTILITY::InvokeWhereToFindDotnet()
...
-    {
-        SysFreeString(pwzDotnetName);
+        if (fIsCurrentProcess64Bit == (dwBinaryType == SCS_64BIT_BINARY))
+        {
+            // The bitness of dotnet matched with the current worker process bitness.
+            return std::make_optional(struDotnetSubstring.QueryStr());
+        }
     }
 
+    Finished:
```
jkotalik: I don't like that we still need the goto. Macro that returns a value if the results failed?

jkotalik:
build failures too.

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
pakrym: Nope, we need a THROW_IF* set of macros but that is a bit too much work. I want this PR out so the bug can be fixed first.

jkotalik: Sure.

src/AspNetCoreModuleV2/CommonLib/HandleWrapper.h
```diff
@@ -0,0 +1,52 @@
...
+
+struct InvalidHandleTraits
+{
+    using HandleType = HANDLE;
+    static const HANDLE DefaultHandle;
+    static void Close(HANDLE handle) { CloseHandle(handle); }
+};
+
+// Workaround for VC++ bug https://developercommunity.visualstudio.com/content/problem/33928/constexpr-failing-on-nullptr-v141-compiler-regress.html
+const HANDLE InvalidHandleTraits::DefaultHandle = INVALID_HANDLE_VALUE;
```
pan-wang: I understand your motivation for wrapping a handle. However, the performance cost is ignorable.

jkotalik: Clarify? Is the perf difference noticeable?

pan-wang: Sorry,  typo.  Updated the comments

pakrym:
Release failure: https://github.com/aspnet/IISIntegration/issues/959

src/AspNetCoreModuleV2/CommonLib/HandleWrapper.h
pakrym: Abstraction is completely free, it generates same code as without abstraction.
See https://godbolt.org/g/y8XzDe
Test.A and B methods.

pan-wang: I meant the constructor and destructor unless compiler put them inline

pakrym: They are both inlined.

PR #962, opened 2018-06-21
Reduce usage of static inprocess application

pakrym:


src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cxx
```diff
@@ -169,16 +177,15 @@ http_get_application_properties(
     _In_ IISConfigurationData* pIISCofigurationData
 )
 {
-    REQUESTHANDLER_CONFIG* pConfiguration = NULL;
-    IN_PROCESS_APPLICATION* pApplication = IN_PROCESS_APPLICATION::GetInstance();
-
-    if (pApplication == NULL)
+    auto pInProcessApplication = IN_PROCESS_APPLICATION::GetInstance();
```
jkotalik: Are we making it a rule that we don't throw with pInvokes from managed into native?

test/IISIntegration.FunctionalTests/Inprocess/ShutdownTests.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        [ConditionalFact]
+        public async Task ServerShutsDownWhenMainExits()
+        {
+            var parameters = Helpers.GetBaseDeploymentParameters();
+            var result = await DeployAsync(parameters);
+
+            var response = await result.RetryingHttpClient.GetAsync("/Shutdown");
+            Assert.True(result.DeploymentResult.HostShutdownToken.WaitHandle.WaitOne(Helpers.DefaultTimeout));
```
jkotalik: Can you add a test that sends "q" to the IISExpress process? Or file an issue to expand shutdown tests.

src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cxx
pakrym: Yeah, let's not leak exception in any direction of managed<->native boundary.

test/IISIntegration.FunctionalTests/Inprocess/ShutdownTests.cs
pakrym: Unfortunately this would require hosting changes: https://github.com/aspnet/Hosting/issues/1355

Issue #959, opened 2018-06-21
Flaky test GetRequestHandlerVersions.GetFolderManyVersions 


pakrym:
https://ci3.dot.net/job/aspnet_IISIntegration/job/dev/job/windows-Configuration_Release_prtest/256/console

```
GetRequestHandlerVersions.GetFolderManyVersions 
14:44:20 d:\j\workspace\windows-confi---30aaf9e2\test\commonlibtests\globalversiontests.cpp(54): error : Value of: fs::create_directories(tempPath + L"\\2.0.0") [D:\j\workspace\windows-Confi---30aaf9e2\test\CommonLibTests\CommonLibTests.vcxproj] [C:\Users\runner\.dotnet\buildtools\korebuild\2.2.0-preview1-17090\KoreBuild.proj]
14:44:20       Actual: false
14:44:20     Expected: true
```

PR #963, opened 2018-06-22
Use guid for temp directories and delete them after tests

pakrym:
May fix https://github.com/aspnet/IISIntegration/issues/959

I suspect we are hitting existing directories that we not cleaned up in previous runs.

### DIFF
--- a/test/CommonLibTests/Helpers.cpp
+++ b/test/CommonLibTests/Helpers.cpp
@@ -3,23 +3,6 @@
 
 #include "stdafx.h"
 
-std::wstring
-Helpers::CreateRandomValue()
-{
-    int randomValue = rand();
-    return std::to_wstring(randomValue);
-}
-
-std::wstring
-Helpers::CreateRandomTempDirectory()
-{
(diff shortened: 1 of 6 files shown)


jkotalik:
I think you are right. I realized now that get random number doesn't have a large range and we didn't delete the directories.

test/CommonLibTests/Helpers.cpp
```diff
@@ -37,12 +14,28 @@ Helpers::ReadFileContent(std::wstring file)
...
 }
+
+TempDirectory::TempDirectory()
+{
+    UUID uuid;
+    UuidCreate(&uuid);
+    RPC_CSTR szUuid = NULL;
+    if (UuidToStringA(&uuid, &szUuid) == RPC_S_OK)
+    {
+        m_path = std::experimental::filesystem::temp_directory_path() / szUuid;
```
jkotalik: make it != RPC_S_OK and flip the contents of the if with the throw.

pakrym: I'll do it in another PR, this was just too green.

PR #964, opened 2018-06-22
Add tracing to shim_config

pakrym:
So we can see "schema not found failures" in debug logs.

### DIFF
--- a/src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
+++ b/src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
@@ -24,66 +24,52 @@ class ASPNETCORE_SHIM_CONFIG
 {
 public:
     virtual
-    ~ASPNETCORE_SHIM_CONFIG();
+    ~ASPNETCORE_SHIM_CONFIG() = default;
 
     HRESULT
     Populate(
         IHttpServer         *pHttpServer,
-        IHttpApplication    *pHttpContext
+        IHttpApplication    *pHttpApplication
     );
 
     STRU*
-    QueryApplicationPhysicalPath(
-        VOID
-    )
+    QueryApplicationPhysicalPath()
     {
         return &m_struApplicationPhysicalPath;
     }
 
     STRU*
-    QueryApplicationPath(
-        VOID
-    )
+    QueryApplicationPath()
     {
         return &m_struApplication;
     }
 
     STRU*
-    QueryConfigPath(
-        VOID
-    )
+    QueryConfigPath()
     {
         return &m_struConfigPath;
     }
 
     STRU*
-    QueryProcessPath(
-        VOID
-    )
+    QueryProcessPath()
     {
         return &m_struProcessPath;
     }
 
     STRU*
-    QueryArguments(
-        VOID
-    )
+    QueryArguments()
     {
         return &m_struArguments;
     }
 
     APP_HOSTING_MODEL
-    QueryHostingModel(
-        VOID
-    )
+    QueryHostingModel()
     {
         return m_hostingModel;
     }
 
     STRU*
-    QueryHandlerVersion(
-        VOID
-    )
+    QueryHandlerVersion()
     {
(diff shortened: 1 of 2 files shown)


src/AspNetCoreModuleV2/AspNetCore/Inc/aspnetcore_shim_config.h
```diff
@@ -24,12 +24,12 @@ class ASPNETCORE_SHIM_CONFIG
...
 public:
     virtual
-    ~ASPNETCORE_SHIM_CONFIG();
+    ~ASPNETCORE_SHIM_CONFIG() = default;
 
     HRESULT
     Populate(
         IHttpServer         *pHttpServer,
-        IHttpApplication    *pHttpContext
+        IHttpApplication    *pHttpApplicationtext
```
jkotalik: huh?

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
```diff
@@ -7,63 +7,36 @@
...
-    if (FAILED(hr))
-    {
-        goto Finished;
-    }
+        &m_struProcessPath));
 
-    hr = GetElementStringProperty(pAspNetCoreElement,
+    // Swallow this error for backward compatability
+    // Use default behavior for empty string
+    GetElementStringProperty(pAspNetCoreElement,
```
jkotalik: Remove this comment. I believe this is incorrect as it will fail to obtain the aspnetcore element.

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
```diff
@@ -7,63 +7,36 @@
...
-
 HRESULT
 ASPNETCORE_SHIM_CONFIG::Populate(
     IHttpServer      *pHttpServer,
     IHttpApplication *pHttpApplication
 )
 {
-    STACK_STRU(strHostingModel, 300);
-    HRESULT                         hr = S_OK;
+    STACK_STRU(strHostingModel, 10);
```
jkotalik: wait why only a size of 10?

jkotalik: That isn't large enough for outofprocess.

pakrym: haha, being able to count wasn't in my resume

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
pakrym: It is correct, it's for hostingModel

pan-wang: this is about the scenario that user has very old ANCM schema (before hostingModel was introduced) and new ancm bits. This only can happen in our test scenario

jkotalik: No, I think behavior described is incorrect. If someone specified the hostingmodel and they didn't have the schema, I believe it would fail to get the aspnetcore element rather than the hosting model attribute.

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
jkotalik: Next offense will require a test.

src/AspNetCoreModuleV2/AspNetCore/src/aspnetcore_shim_config.cpp
pan-wang: At this line, we already has the aspnetcore element. the very old V1 schema does aspnetcore element but not hostingModel attribute. this code was inherited from V1.

PR #965, opened 2018-06-22
Switch to tracing macros in application info

pakrym:


src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -443,14 +406,10 @@ APPLICATION_INFO::FindNativeAssemblyFromHostfxr(
     {
         // Host fxr version is incorrect (need a higher version).
         // TODO log error
-        hr = E_FAIL;
-        goto Finished;
+        FINISHED(E_FAIL);
```
jkotalik: FINISHED doesn't imply it failed. FINISHED_FAIL?

pakrym: FINISHED is more like unconditional return/

PR #967, opened 2018-06-25
caching handlers to avoid heap fregamentation

pan-wang:
recreate the pull request with some small changes


jkotalik:
Why did you recreate it rather than pushing over your old one? Either force push this branch onto your other pull request or close your old pull request.

pan-wang:
I had to adopt the macroRebase of previous PR got totally messed. As there were lots of changes merged, I just to rethink about my change instead of fixing the rebase issue, e.g., I had to adopt the macro and changed it. You can treat it as a new PR.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -218,26 +219,40 @@ APPLICATION_INFO::EnsureApplicationCreated(
...
         {
-
+            m_appStatus = APPLICATION_STATUS::FAIL;
             // Move the request handler check inside of the lock
             // such that only one request finds and loads it.
             // FindRequestHandlerAssembly obtains a global lock, but after releasing the lock,
             // there is a period where we could call
 
-            RETURN_IF_FAILED(FindRequestHandlerAssembly(struExeLocation));
+            FINISHED_IF_FAILED(FindRequestHandlerAssembly(struExeLocation));
```
pakrym: Don't mix return and Finished in one method.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -218,26 +219,40 @@ APPLICATION_INFO::EnsureApplicationCreated(
...
+            m_appStatus = APPLICATION_STATUS::RUNNING;
         }
     }
 
-    return S_OK;
+    Finished:
+    if (FAILED(hr))
+    {
+        // Log the failure and update application info to not try again
+        UTILITY::LogEventF(g_hEventLog,
```
pakrym: Doesn't seem like you are updating apllication info

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -41,6 +42,12 @@ InitializeGlobalConfiguration(
         }
 
         g_pHttpServer = pServer;
+        if (FAILED(hr = ALLOC_CACHE_HANDLER::StaticInitialize()) ||
```
pakrym: FINISHED_IF_FAILED

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -94,13 +103,11 @@ CreateApplication(
...
 
-    // Initialze some global variables here
-    InitializeGlobalConfiguration(pServer);
-
     try
     {
-        hr = REQUESTHANDLER_CONFIG::CreateRequestHandlerConfig(pServer, pHttpApplication, &pConfig);
-        if (FAILED(hr))
+        // Initialze some global variables here
+        if (FAILED(hr = InitializeGlobalConfiguration(pServer)) ||
```
pakrym: FINISHED_IF_FAILED

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
pan-wang: I am not a big fun to use macro/script as it makes the code hard for read and debug (not just for us but also for CSS).  Updating it.

pakrym: But at least we would know what failed and where. Which makes debugging much easier

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
jkotalik: Isn't this log message repetitive?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pan-wang: fixed

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
jkotalik: Just remove the comment. above the call to LogEventF.

src/AspNetCoreModuleV2/InProcessRequestHandler/precomp.hxx
```diff
@@ -40,6 +40,10 @@
 #undef WINVER
 #undef _WIN32_WINNT
 
+#include <acache.h>
+#include <exceptions.h>
```
jkotalik: Add these to individual classes.

src/AspNetCoreModuleV2/OutOfProcessRequestHandler/dllmain.cxx
```diff
@@ -214,6 +214,12 @@ EnsureOutOfProcessInitializtion()
             goto Finished;
         }
 
+        hr = ALLOC_CACHE_HANDLER::StaticInitialize();
```
jkotalik: FINISHED_IF_FAILED()

src/AspNetCoreModuleV2/InProcessRequestHandler/precomp.hxx
pan-wang: those two will be added for almost every file. Not a good idea to duplicate the include everywhere

src/AspNetCoreModuleV2/OutOfProcessRequestHandler/dllmain.cxx
pan-wang: I can change  it now. We need rescan the files to switch all of them.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pakrym: There are still `return S_OK;` in this method

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -218,26 +219,40 @@ APPLICATION_INFO::EnsureApplicationCreated(
         //
         if (!m_fAppOfflineFound)
         {
-
+            m_appStatus = APPLICATION_STATUS::FAIL;
```
pakrym: Why don't you set it in `if (FAILED(hr)):` ?

pakrym: Also does it go from STARTING to FAIL to RUNNING?

pan-wang: the initial state of applicationinfo (during creation) is FAILED, we then do not need to set FAILED at finished which is not guarded by lock. The status goes from FAILED to STARTING or RUNNING.

pakrym: It's not what I see in code. 

OnExecuteRequestHandler calls GetOrCreateApplicationInfo that sets the status to STARTING then EnsureApplicationCreated gets called setting status to FAILED and then back to RUNNING. 

Did I miss something?

pan-wang: the first one is the status of applicationinfo (status is member element of applicationinfo) and the second one is status of application. I just reused the enum type.

pan-wang: my fault. I had thought I could use one enum to replace the two BOOLs in previous pr.  Revert this enum change.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -182,39 +184,73 @@ APPLICATION_INFO::EnsureApplicationCreated(
...
-    if (m_pApplication == NULL)
+    if (m_fDoneAppCreation)
     {
-        SRWExclusiveLock lock(m_srwLock);
+        // application is NULL and CreateApplication failed previously
+        FINISHED_IF_FAILED(E_APPLICATION_ACTIVATION_EXEC_FAILURE);
+    }
+    else
+    {
+        AcquireSRWLockExclusive(&m_srwLock);
```
jkotalik: Just create a SRWLockExclusive here so you can remove the fLocked bool

src/AspNetCoreModuleV2/AspNetCore/src/applicationmanager.cxx
```diff
@@ -251,25 +250,9 @@ APPLICATION_MANAGER::RecycleApplicationFromManager(
         // Removed the applications which are impacted by the configurtion change
         m_pApplicationInfoHash->DeleteIf(FindConfigChangedApplication, (PVOID)&context);
 
-        if (dwPreviousCounter != m_pApplicationInfoHash->Count())
```
jkotalik: I'm still a bit confused on why you removed this. Can you clarify?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pan-wang: I want make sure setting m_fDoneAppCreation= TRUE is inside the lock even for failure condition

src/AspNetCoreModuleV2/AspNetCore/src/applicationmanager.cxx
pan-wang: in the past, we did not create an applicationinfo instance and add it to the hash table in case of hostingmodel conflict and some other cases, i.e., we did not cache the failure state.  This leads to an issue that we tries to create applicationinfo again and again for such failure case.  If there is app recycle happened, we have to check whether the recycled app is in out hashtable or not.  This is why we did not counter check.
In the new code, we always create an applicationinfo instance and insert it into the hash table, no matter the application can be created or not so that we can cache the failure state.  We thus do not need to check the hashtable counter.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pakrym: then put SRWLockExclusive  one level higher and it would be active until method exits.
I removed all manual locking from shim, please don't add it back.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -182,39 +184,73 @@ APPLICATION_INFO::EnsureApplicationCreated(
         return S_OK;
     }
 
-    if (m_pApplication == NULL)
+    if (m_fDoneAppCreation)
     {
-        SRWExclusiveLock lock(m_srwLock);
+        // application is NULL and CreateApplication failed previously
+        FINISHED_IF_FAILED(E_APPLICATION_ACTIVATION_EXEC_FAILURE);
```
pakrym: Just FINISHED

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -182,39 +184,73 @@ APPLICATION_INFO::EnsureApplicationCreated(
...
+    else
+    {
+        AcquireSRWLockExclusive(&m_srwLock);
+        fLocked = TRUE;
+
         if (m_pApplication != NULL)
         {
-            return S_OK;
+            // another thread created the applicaiton
+            goto Finished;
```
pakrym: FINISHED(S_OK)

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -300,15 +336,15 @@ APPLICATION_INFO::FindRequestHandlerAssembly(STRU& location)
...
+                FINISHED_IF_FAILED(HRESULT_FROM_WIN32(GetLastError()));
             }
         }
 
         g_pfnAspNetCoreCreateApplication = (PFN_ASPNETCORE_CREATE_APPLICATION)
             GetProcAddress(g_hAspnetCoreRH, "CreateApplication");
         if (g_pfnAspNetCoreCreateApplication == NULL)
         {
-            FINISHED(HRESULT_FROM_WIN32(GetLastError()));
+            FINISHED_IF_FAILED(HRESULT_FROM_WIN32(GetLastError()));
```
pakrym: Why change this?

src/AspNetCoreModuleV2/AspNetCore/src/applicationmanager.cxx
```diff
@@ -116,6 +110,12 @@ APPLICATION_MANAGER::GetOrCreateApplicationInfo(
             }
         }
 
+        hr = m_pApplicationInfoHash->InsertRecord(pApplicationInfo);
```
pakrym: FINISHED_IF_FAILED

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
jkotalik: My bad on the merge conflicts. Revert this.

pan-wang:
It seems there are couple rebate issue.  Will fix them tonight

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pan-wang: This also makes the logging part inside the lock. We need some flexibility to do the minimum lock.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pan-wang: We have too many macros.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pakrym: In your code lock would never be returned in case of an exception. We log under lock in a lot of places.

Issue #943, opened 2018-06-20
Introduce project property for selecting ANCM version. (Will be used by websdk)


shirhatti:

shirhatti:
Required for https://github.com/aspnet/websdk/issues/367

Issue #928, opened 2018-06-15
Add in-proc capability to props file for tooling to detect 


shirhatti:
We'll need to carry a props file with Microsoft.AspNetCore.Server.IIS package to add a capability that tooling can use to detect whether in-proc is supported or not.

cc @BillHiebert

shirhatti:
Never mind. I realize, we've already added this https://github.com/aspnet/IISIntegration/blob/6f54ed1c68ad732802b8813dad4c6f0b872d8c0b/src/Microsoft.AspNetCore.Server.IIS/Microsoft.AspNetCore.Server.IIS.targets
@BillHiebert can you confirm if this will suffice?

shirhatti:
Whoops. We messed up https://github.com/aspnet/Universe/blob/dev/src/Microsoft.AspNetCore.App/build/netcoreapp2.2/Microsoft.AspNetCore.App.props

We'll need to introduce a new capability

PR #969, opened 2018-06-25
Adds project properties and project capabilities to IIS and IISIntegration

jkotalik:
For #943 and #928. @shirhatti I'll need some help verifying project capabilities because I didn't see it in the binlog.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IIS/Microsoft.AspNetCore.Server.IIS.targets
+++ b/src/Microsoft.AspNetCore.Server.IIS/Microsoft.AspNetCore.Server.IIS.targets
@@ -3,7 +3,11 @@
     Capability that enables Visual Studio support for hosting Asp.Net Core applications in the IIS process
   -->
   <ItemGroup>
-    <ProjectCapability Include="AspNetInProcessHosting" />
+    <ProjectCapability Include="AspNetCoreInProcessHosting" />
   </ItemGroup>
+
+  <PropertyGroup>
+    <AspNetCoreModuleVersion Condition=" '$(AspNetCoreModuleVersion)' == '' ">V2</AspNetCoreModuleVersion>
+  </PropertyGroup>
 
 </Project>
--- a/src/Microsoft.AspNetCore.Server.IISIntegration/Microsoft.AspNetCore.Server.IISIntegration.csproj
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration/Microsoft.AspNetCore.Server.IISIntegration.csproj
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.AspNetCore.HttpOverrides" Version="$(MicrosoftAspNetCoreHttpOverridesPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="$(MicrosoftExtensionsOptionsPackageVersion)" />
+    <Content Include="$(PackageId).targets" PackagePath="build/$(TargetFramework)/" />
   </ItemGroup>
 
 </Project>
--- a/src/Microsoft.AspNetCore.Server.IISIntegration/Microsoft.AspNetCore.Server.IISIntegration.targets
+++ b/src/Microsoft.AspNetCore.Server.IISIntegration/Microsoft.AspNetCore.Server.IISIntegration.targets
@@ -1,0 +1,7 @@
+<Project>
+
+  <PropertyGroup>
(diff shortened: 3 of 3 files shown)


src/Microsoft.AspNetCore.Server.IISIntegration/Microsoft.AspNetCore.Server.IISIntegration.csproj
```diff
@@ -24,6 +24,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="$(MicrosoftExtensionsOptionsPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.SecurityHelper.Sources" PrivateAssets="All" Version="$(MicrosoftExtensionsSecurityHelperSourcesPackageVersion)" />
+    <Content Include="$(PackageId).targets" PackagePath="build/$(TargetFramework)/" />
```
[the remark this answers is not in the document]
jkotalik: We need to start setting AspNetCoreModuleVersion from V1 to V2 by default. People can still use out of proc without calling .UseIIS().

jkotalik:
AppVeyor failure should be fixed by #970

jkotalik:
@shirhatti I re reviewed this and it looks good locally.

shirhatti:
Naive question, but targets file work transitively right? We don't need to update the shared framework and introduce this target, correct?

shirhatti:
FYI @BillHiebert

pakrym:
No, by default they don't work transitively.

shirhatti:
So we'll need to update https://github.com/aspnet/Universe/blob/dev/src/Microsoft.AspNetCore.App/build/netcoreapp2.2/Microsoft.AspNetCore.App.props as well?

jkotalik:
Not sure, is there a way to transitively flow .target files? If not I'll open another Pr in universe for .App and maybe .All

pakrym:
We did similar thing here https://github.com/aspnet/AzureIntegration/pull/190 but ask @natemcmaster  first.

src/Microsoft.AspNetCore.Server.IIS/Microsoft.AspNetCore.Server.IIS.targets
```diff
@@ -3,7 +3,11 @@
     Capability that enables Visual Studio support for hosting Asp.Net Core applications in the IIS process
   -->
   <ItemGroup>
-    <ProjectCapability Include="AspNetInProcessHosting" />
```
natemcmaster: So, does this mean we can drop the AspNetInProcessHosting capability? I think this was added to the 2.1 Microsoft.AspNetCore.App metapackage and wasn't removed when we yanked in-proc.

jkotalik: Yeah we forgot to drop it from 2.1 .App. We will have to rename it, hence AspNetCoreInProcessHosting.
@natemcmaster is there a way to make M.A.App transitively get the project capability from .IIS or .IISIntegration?

natemcmaster: It could come transitively from .IISIntegration, but not .IIS.

jkotalik: Why is that? Because .IIS is not part of .App yet? I think the plan is to add it to .App for 2.2

natemcmaster: .IIS can't be part of .App because it contains Windows-only binaries. It can be part of runtime.win.Microsoft.AspNetCore.App, but not the portable, "Microsoft.AspNetCore.App" version.

jkotalik: Then we will need to add it to .App as we can't require .IIS to include .IISIntegration

natemcmaster: Ok. Here's the cleanup: https://github.com/aspnet/Universe/pull/1237

jkotalik:
Is this good to go?

pakrym:
Let's figure out if `.IIS` goes into `.App` and in what way before doing this.

natemcmaster:
Context for when we designed and implemented this the first time…

https://github.com/aspnet/Universe/issues/762

https://github.com/aspnet/Universe/pull/989

These worked just fine, but were yanked because in-proc was dropped for 2.1. Based on our conversations, I’m still under the impression the original design works.

jkotalik:
This part of solving the .App issue can be tangential to this PR. Everyone okay with the current stuff here?

pakrym:
https://github.com/aspnet/IISIntegration/issues/1048

Issue #939, opened 2018-06-19
Flaky tests: It was not possible to find any compatible framework version 


pakrym:
I've seen this failure couple times mostly in AppVeyor runs:

```
 It was not possible to find any compatible framework version [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
 The specified framework 'Microsoft.NETCore.App', version '2.2.0-preview1-26606-01' was not found. [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
   - Check application dependencies and target a framework version installed at: [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
       C:\Program Files\dotnet\ [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
   - Installing .NET Core prerequisites might help resolve this problem: [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
       http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409 [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
   - The .NET Core framework and SDK can be installed from: [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
       https://aka.ms/dotnet-download [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
   - The following versions are installed: [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
       1.0.4 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App] [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
       1.0.5 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App] [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17081\KoreBuild.proj]
[...]

PR #970, opened 2018-06-25
Remove where.exe timeout

pakrym:
to 10 seconds

https://github.com/aspnet/IISIntegration/issues/939

### DIFF
--- a/src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.h
+++ b/src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.h
@@ -96,7 +96,5 @@ public:
 
     static
     std::wstring ExpandEnvironmentVariables(const std::wstring & str);
-
-    static const DWORD m_WhereExeProcessTimeoutMS = 10 * 1000;
 };
 
--- a/src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.h
+++ b/src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.h
@@ -96,5 +96,7 @@ public:
 
     static
     std::wstring ExpandEnvironmentVariables(const std::wstring & str);
+
+    static const DWORD m_WhereExeProcessTimeoutMS = 10 * 1000;
 };
 
--- a/src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
+++ b/src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
@@ -497,13 +497,8 @@ HOSTFXR_UTILITY::InvokeWhereToFindDotnet()
     hProcess = processInformation.hProcess;
     hThread = processInformation.hThread;
 
-    // Wait for where.exe to return, waiting 2 seconds.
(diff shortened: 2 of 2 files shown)


src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.h
```diff
@@ -96,5 +96,7 @@ class HOSTFXR_UTILITY
 
     static
     std::wstring ExpandEnvironmentVariables(const std::wstring & str);
+
+    static const DWORD m_WhereExeProcessTimeout = 10 * 1000;
```
Tratcher: Name or comment that the units are seconds.

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.h
```diff
@@ -96,5 +96,7 @@ class HOSTFXR_UTILITY
 
     static
     std::wstring ExpandEnvironmentVariables(const std::wstring & str);
+
+    static const DWORD m_WhereExeProcessTimeoutMS = 10 * 1000;
```
pan-wang: what's the motivation to increate the timeout to 10s? should we consider a environment variable for the value if we cannot determine the max timeout value

pakrym: Motivation is that server should not be failing to start on slow machines.

jkotalik: Calling where.exe was timing out on our CI (The main reason we saw test failures). We could make the timeout configurable.

pakrym: I don't think we should make it configurable. Maybe we should just increase it to a minute. 

The side effect of where.exe timing out is often app not starting.

pan-wang: I meant whether there would be any AV or crash.

pakrym:
After talking to @mikeharder I removed `where.exe` timeout all together. Not getting a result of `where.exe` and continuing just leads to bugs and startup flakiness. IIS is free to kill worker process if it doesn't start in time.

mikeharder:
Can we add a test for this behavior?  We could use a fake `where.exe` which never returns, and test what IIS does in this scenario.

pakrym:
We don't have full IIS testing infrastructure yet.

mikeharder:
@jkotalik: Fine with adding tests in separate PR, we can just create an issue for the test debt and prioritize it accordingly.  Might even want a separate project for these tests if they're very slow or different from other tests.

mikeharder:
@pakrym: Is there a big difference between IIS and IIS Express with regards to startup timeout?  When I said "IIS" I meant "IIS and/or IIS Express".

Issue to add tests for startup timeouts: https://github.com/aspnet/IISIntegration/issues/972

pakrym:
Yes, when we use IIS Express in tests we start `iisexpress.exe` process manually, full IIS has an orchestrator that coordinates worker processes and startup timeout is implemented in it.

It's not possible to test features related to the process management using IIS Express.

Issue #956, opened 2018-06-21
Add full IIS tests that can at least be run locally.


jkotalik:
- Revive IIS deployer?
- Get running on custom machines.

PR #973, opened 2018-06-25
Fix null handle issue on full IIS

jkotalik:
Another part of #897 for fixing logging tests. On full IIS, after we set stdout and stderr, we need to redirect it to "NUL:" to avoid Console.Write accessing an invalid handle.


src/AspNetCoreModuleV2/RequestHandlerLib/PipeOutputManager.cpp
```diff
@@ -32,6 +32,31 @@ PipeOutputManager::StopOutputRedirection()
...
     fflush(stderr);
 
+    // Restore the original stdout and stderr handles of the process,
+    // as the application has either finished startup or has exited.
+
+    // If stdout/stderr were not set, we need to set it to NUL:
+    // such that other calls to Console.WriteLine don't use an invalid handle
+    FILE *stream;
+
+    if (m_fdPreviousStdOut >= 0)
```
shhsu: could m_fdPreviousStdOut ever be 0?

jkotalik: Yeah, I believe stdin is by default set to 0. It's better to be defensive with this check.

jkotalik:
Working out how we can test this. I can try to fake it in native by setting stdout and stderr, but it would probably be better just to have #956

src/AspNetCoreModuleV2/RequestHandlerLib/PipeOutputManager.cpp
```diff
@@ -32,6 +32,31 @@ PipeOutputManager::StopOutputRedirection()
...
+    {
+        _dup2(m_fdPreviousStdOut, _fileno(stdout));
```
pakrym: Check all exit codes

pan-wang: please log a warning in case of failure

src/AspNetCoreModuleV2/RequestHandlerLib/PipeOutputManager.cpp
```diff
@@ -2,6 +2,11 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 #include "stdafx.h"
+#include "exceptions.h"
+
+#define LOG_IF_DUPFAIL(err) do { if (err == -1) { LOG_IF_FAILED(HRESULT_FROM_WIN32(_doserrno)); }} while (0, 0);
```
pan-wang: I am not a fun of macro as it breaks readability.
LOG_IF_FAILED should be DBG_LOG_IF_FAILED as it only log when debug flag enabled

src/AspNetCoreModuleV2/RequestHandlerLib/PipeOutputManager.cpp
```diff
@@ -32,6 +37,31 @@ PipeOutputManager::StopOutputRedirection()
...
+    // Restore the original stdout and stderr handles of the process,
+    // as the application has either finished startup or has exited.
+
+    // If stdout/stderr were not set, we need to set it to NUL:
+    // such that other calls to Console.WriteLine don't use an invalid handle
+    FILE *stream;
+
+    if (m_fdPreviousStdOut >= 0)
+    {
+        LOG_IF_DUPFAIL(_dup2(m_fdPreviousStdOut, _fileno(stdout)));
```
pan-wang: those are important failure which may lead to AV. I think we should log to windows event log so that user can immediately identifity the root cause. Otherwise user needs to set debug flag and restart the process which may not repro the failure.

pakrym: How does it differ from hundreds of other HRESULTs we don't log?

jkotalik: I think special casing this failure and logging to the windows event log is a bit much, however I do think it is important to capture these startup logs. Proposal here: https://github.com/aspnet/IISIntegration/issues/974

pan-wang: for application activation path, if there is a failure blocking activation, we always log it somewhere as it is critical.
for non-critical error, e.g., cannot create the log file given by user's configuration, we log a warning.  
In this case, we swallow the error which could be critical, e.g., reset the stdout failed will lead to AV (this is the motivation of the fix, it is worth to logging it in windows event log.

PR #979, opened 2018-06-27
Add full IIS tests

jkotalik:
A few tests with a full IIS deployer. Making PR to see if jenkins works and blatantly wrong behavior.


src/AspNetCoreModuleV2/RequestHandlerLib/FileOutputManager.cpp
```diff
@@ -7,7 +7,10 @@
 #include "exceptions.h"
 #include "debugutil.h"
 
-FileOutputManager::FileOutputManager()
+FileOutputManager::FileOutputManager() :
+    m_hLogFileHandle(INVALID_HANDLE_VALUE),
```
jkotalik: This is unrelated to this change. Will make it a separate PR.

pakrym: Handle wrapper? :)

jkotalik: Sure.

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,173 @@
...
+                _serverManager.CommitChanges();
+            }
+        }
+
+        private void ConfigureAppPool()
+        {
+            var pool = _serverManager.ApplicationPools.Add(AppPoolName);
+            pool.StartMode = StartMode.AlwaysRunning;
+            pool.ProcessModel.IdentityType = ProcessModelIdentityType.LocalSystem;
+            pool.ProcessModel.LoadUserProfile = true;
```
pakrym: Why?

test/IISIntegration.FunctionalTests/Utilities/SkipIfNotAllowedToRunIISAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+                var identity = WindowsIdentity.GetCurrent();
+                var principal = new WindowsPrincipal(identity);
+                var isInrole = principal.IsInRole(WindowsBuiltInRole.Administrator);
+
+                var w3wpExists = File.Exists(Path.Combine(Environment.SystemDirectory, "Windows", "System32", "inetsrv", "w3wp.exe"));
+                return w3wpExists && isInrole;
+            }
+        }
+
+        public string SkipReason { get; } = $"Hostable Web Core not availible, {TestServer.HostableWebCoreLocation} not found.";
```
pakrym: Change this to be no w3wp exist or not admin depending on why we are skipping

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                deploymentParameters.ApplicationType = applicationType;
+
+                using (var deployer = new IISDeployer(deploymentParameters, loggerFactory))
+                {
+                    var deploymentRes = await deployer.DeployAsync();
+                    var deploymentResult = new IISDeploymentResult(deploymentRes, logger);
+
+                    if (applicationType == ApplicationType.Standalone)
+                    {
+                        Helpers.ModifyAspNetCoreSectionInWebConfig(deploymentResult, "processPath", "InProcessWebSite.exe");
```
pakrym: Shouldn't this go into deployer?

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                        Helpers.ModifyAspNetCoreSectionInWebConfig(deploymentResult, "arguments", "");
+                    }
+
+                    // Work around for https://github.com/aspnet/IISIntegration/issues/976
+                    Helpers.ModifyAspNetCoreSectionInWebConfig(deploymentResult, "stdoutLogEnabled", "false");
```
pakrym: Just set it on web.config

pakrym: Or I may have done it already

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                    var response = await RetryHelper.RetryRequest(() =>
```
pakrym: Retrying client.

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        [SkipIfNotAllowedToRunIIS]
+        public Task HelloWorld_IISExpress_Clr_X64_Portable()
+        {
+            return HelloWorld(RuntimeFlavor.CoreClr, ApplicationType.Portable);
+        }
+
+        private async Task HelloWorld(RuntimeFlavor runtimeFlavor, ApplicationType applicationType)
+        {
+            var testName = $"HelloWorld_{runtimeFlavor}";
+            using (StartLog(out var loggerFactory, testName))
```
pakrym: Please don't bring the old patterns back.

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
jkotalik: Think you did.

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
jkotalik: Yeah. Lot of this will go away once we move this into hosting.

build/dependencies.props
```diff
@@ -38,6 +38,7 @@
     <MicrosoftNETCoreApp22PackageVersion>2.2.0-preview1-26618-02</MicrosoftNETCoreApp22PackageVersion>
     <MicrosoftNetHttpHeadersPackageVersion>2.2.0-preview1-34530</MicrosoftNetHttpHeadersPackageVersion>
     <MicrosoftNETTestSdkPackageVersion>15.6.1</MicrosoftNETTestSdkPackageVersion>
+    <MicrosoftWebAdministrationPackageVersion>11.1.0</MicrosoftWebAdministrationPackageVersion>
```
Tratcher: Is this in our external dependency list in Universe?

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
+{
+    public class IISTests : LoggedTest
+    {
+        public IISTests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        [SkipIfIISExpressSchemaMissingInProcess]
```
Tratcher: IIS has its own schema file. Do you need a seperate skip attribute?

test/IISIntegration.FunctionalTests/Utilities/SkipIfNotAllowedToRunIISAttribute.cs
jkotalik: copy pasta.

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
Tratcher: These Skips can be placed at the class level.

Tratcher: ConditionalFact

test/IISIntegration.FunctionalTests/Utilities/SkipIfNotAllowedToRunIISAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+    public sealed class SkipIfNotAllowedToRunIISAttribute : Attribute, ITestCondition
+    {
+        public bool IsMet {
+            get
+            {
+                // Two conditions need to be met:
+                // IIS needs to be installed
+                var identity = WindowsIdentity.GetCurrent();
+                var principal = new WindowsPrincipal(identity);
+                var isInrole = principal.IsInRole(WindowsBuiltInRole.Administrator);
```
Tratcher: There's a difference between being an administrator and running elevated. Does this detect elevation?

test/IISIntegration.FunctionalTests/Utilities/SkipIfNotAllowedToRunIISAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
+{
+    [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method)]
+    public sealed class SkipIfNotAllowedToRunIISAttribute : Attribute, ITestCondition
+    {
+        public bool IsMet {
+            get
+            {
+                // Two conditions need to be met:
```
Tratcher: And running as elevated admin

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
jkotalik: Yeah I'll remove it; I just wanted to get it working first off. Will cleanup now.

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
pakrym: Why not now?

pakrym: We may not move it into hosting.

test/IISIntegration.FunctionalTests/Utilities/IISDeployer.cs
```diff
@@ -0,0 +1,85 @@
...
+                var uri = TestIISUriHelper.BuildTestUri(ServerType.IIS, DeploymentParameters.ApplicationBaseUriHint);
+
+                lock (_syncObject)
+                {
+                    // To prevent modifying the IIS setup concurrently.
+                    _application.StartIIS(uri, contentRoot);
+                }
+
+                // Warm up time for IIS setup.
+                await Task.Delay(1000);
```
Tratcher: 👎 Find a way to check its status.

test/IISIntegration.FunctionalTests/Utilities/SkipIfNotAllowedToRunIISAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using System.IO;
+using System.Security.Principal;
+using Microsoft.AspNetCore.Testing.xunit;
+using Microsoft.Win32;
+
+namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
+{
+    [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method)]
+    public sealed class SkipIfNotAllowedToRunIISAttribute : Attribute, ITestCondition
```
Tratcher: SkipIfIISCannotRun?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,173 @@
...
+        public string AppPoolName { get; }
+
+        public void StartIIS(Uri uri, string contentRoot)
+        {
+            using (_logger.BeginScope("StartIIS"))
+            {
+                var port = uri.Port;
+                if (port == 0)
+                {
+                    port = (uri.Scheme == "https") ? TestPortHelper.GetNextSSLPort() : TestPortHelper.GetNextPort();
```
Tratcher: This SslPort logic won't work, those ports are specifically bound to IIS Express. You'd have to create an ACL using the dev cert. Throw NotSupported for now?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,173 @@
...
+        private void ConfigureAppHostConfig(string dllRoot)
+        {
+            var config = _serverManager.GetApplicationHostConfiguration();
+            var globalModulesSection = config.GetSection("system.webServer/globalModules");
+
+            var ancmFile = GetAncmLocation(dllRoot);
+            var found = false;
+            foreach (var element in globalModulesSection.GetCollection())
+            {
+                if ((string)element["name"] == @"AspNetCoreModuleV2")
```
Tratcher: What about v1? Or is this deployer only for in-proc?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+
+        private void RemoveAppHostConfig()
+        {
+            var config = _serverManager.GetApplicationHostConfiguration();
+            var globalModulesSection = config.GetSection("system.webServer/globalModules");
+            foreach (var element in globalModulesSection.GetCollection())
+            {
+                if ((string)element["name"] == @"AspNetCoreModuleV2")
```
Tratcher: What about the case where AspNetCoreModuleV2 had been added? Should it be removed again?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,173 @@
...
+                var addElement = globalModulesSection.GetCollection().CreateElement("add");
+                addElement["name"] = "AspNetCoreModuleV2";
+                addElement["image"] = ancmFile;
+                globalModulesSection.GetCollection().Add(addElement);
+            }
+        }
+
+        private string GetAncmLocation(string dllRoot)
+        {
+            var arch = _deploymentParameters.RuntimeArchitecture == RuntimeArchitecture.x64 ? @"x64\aspnetcorev2.dll" : @"x86\aspnetcorev2.dll";
```
Tratcher: We may have to find a way to prevent any IIS parallel test execution, even across repos.

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
pakrym: I think we need to carry our own apphostconfig backup old one and use ours for tests to avoid machine state affecting tests.

test/IISIntegration.FunctionalTests/Utilities/IISDeployer.cs
```diff
@@ -0,0 +1,85 @@
...
+                    contentRoot = DeploymentParameters.PublishedApplicationRootPath;
+                    //AddAccessPermissions(contentRoot, _application.AppPoolName);
+                }
+
+                var uri = TestIISUriHelper.BuildTestUri(ServerType.IIS, DeploymentParameters.ApplicationBaseUriHint);
+
+                lock (_syncObject)
+                {
+                    // To prevent modifying the IIS setup concurrently.
+                    _application.StartIIS(uri, contentRoot);
```
pakrym: Did you check if running other tests concurrently would kill already running apps?

test/IISIntegration.FunctionalTests/Utilities/IISDeployer.cs
pakrym: I thought management API had a way to check status.

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
Tratcher: Let's try keeping in here and see if it's easier to maintain. However you'd need to add a separate package if you want it to be consumed in other repos. That can wait though.

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
Tratcher: That's dramatic. Even more reason to work out a way to prevent parallel execution across repos.

test/IISIntegration.FunctionalTests/Inprocess/IISTests.cs
jkotalik: Alright, I'll err on the side of putting everything in here then.

build/dependencies.props
jkotalik: Believe we will need to add it.

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        private void RemoveAppPool()
+        {
+            var appPool = _serverManager.ApplicationPools.FirstOrDefault(pool => pool.Name == AppPoolName);
+            if (appPool != null)
+            {
+                appPool.Stop();
+                _serverManager.ApplicationPools.Remove(appPool);
+                _serverManager.CommitChanges();
```
mikeharder: When adding the site and app pool, the code calls `_serverManager.CommitChanges()` once at the end.  But when removing the site and app pool, the code calls `_serverManager.CommitChanges()` after each operation.  These should be consistent, and I would change the removal code to call once at the end if it's sufficient.  Minimize calls to `CommitChanges()`.

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+
+        private void ConfigureAppPool()
+        {
+            var pool = _serverManager.ApplicationPools.Add(AppPoolName);
+            pool.StartMode = StartMode.AlwaysRunning;
+            pool.ProcessModel.IdentityType = ProcessModelIdentityType.LocalSystem;
+            pool.ProcessModel.LoadUserProfile = true;
+        }
```
mikeharder: [Docs](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/?view=aspnetcore-2.1&tabs=aspnetcore2x#create-the-iis-site) recommend setting the CLR version to "No Managed Code".  I think we should do the same, by setting `pool.ManagedRuntimeVersion = string.Empty;`

jkotalik:
Still working out parallel test execution problems a few newer comments. If we want to keep the IISDeployer in this repo, we should move the IISExpress deployer here too. There are too many shared components between them.

test/IISIntegration.FunctionalTests/Utilities/SkipIfNotAllowedToRunIISAttribute.cs
jkotalik: It doesn't. Not sure if we care though?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
jkotalik: I'd handle this in another PR as this is the only repo doing it.

jkotalik:
It seems IIS in installed on appveyor. Seems to be timing out though; will look into why.

test/IISIntegration.FunctionalTests/Utilities/IISDeployer.cs
jkotalik: Tests are already running sequentially

jkotalik:
I think I hit most of the feedback. 🆙 📅

test/IISIntegration.FunctionalTests/Utilities/IISAncmValidator.cs
```diff
@@ -0,0 +1,71 @@
...
+                SkipReason = "Could not read ANCM schema configuration";
+                return;
+            }
+
+            IsMet = ancmConfig
+                .Root
+                .Descendants("attribute")
+                .Any(n => "hostingModel".Equals(n.Attribute("name")?.Value, StringComparison.Ordinal));
+
+            SkipReason = IsMet ? null : "IIS schema be upgraded to support ANCM.";
```
mikeharder: Typo in string

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,249 @@
...
+    {
+        private const int IISRetryLimit = 500;
+        private const int IISWaitMillisecondsTimeout = 10;
+        private readonly ServerManager _serverManager = new ServerManager();
+        private readonly DeploymentParameters _deploymentParameters;
+        private readonly ILogger _logger;
+        private readonly string _ancmVersion;
+        private readonly object _syncLock = new object();
+        private readonly string _apphostConfigPath = Path.Combine(Environment.SystemDirectory, "inetsrv", "config", "applicationhost.config");
+        private readonly string _apphostConfigBackupPath = Path.Combine(Environment.SystemDirectory, "inetsrv", "config", "applicationhost.config.backup");
```
mikeharder: Change to `private static readonly`

test/IISIntegration.FunctionalTests/Utilities/SkipIfIISCannotRunAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.AspNetCore.Server.IntegrationTesting;
+using Microsoft.AspNetCore.Testing.xunit;
+using Microsoft.Win32;
+
+namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
+{
+    [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method)]
+    public sealed class SkipIfIISCannotRunAttribute : Attribute, ITestCondition
+    {
+
```
mikeharder: Remove extra whitespace

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,249 @@
...
+
+namespace Microsoft.AspNetCore.Server.IntegrationTesting
+{
+    /// <summary>
+    /// Represents the IIS website registered in the global applicationHost.config
+    /// </summary>
+    internal class IISApplication
+    {
+        private const int IISRetryLimit = 500;
+        private const int IISWaitMillisecondsTimeout = 10;
```
mikeharder: I think it's a bit cleaner to instead use two `TimeSpan`:

```
private static readonly TimeSpan _timeout = TimeSpan.FromSeconds(5);
private static readonly TimeSpan _retryDelay = TimeSpan.FromMilliseconds(10);
...
var sw = Stopwatch.StartNew();
while (sw.Elapsed < _timeout) {
    // Attempt to check status, return if done
    await Task.Delay(_retryDelay);
}
```

pakrym:
Make appveyor pass.

test/IISIntegration.FunctionalTests/Utilities/IISAncmValidator.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                SkipReason += "The machine does not have IIS installed.";
+                return;
+            }
+
+            var ancmConfigPath = Path.Combine(Environment.SystemDirectory, "inetsrv", "config", "schema", "aspnetcore_schema_v2.xml");
+
+            if (!File.Exists(ancmConfigPath))
+            {
+                SkipReason = "IIS is not installed.";
```
pakrym: Schema

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,249 @@
...
+        private readonly ServerManager _serverManager = new ServerManager();
+        private readonly DeploymentParameters _deploymentParameters;
+        private readonly ILogger _logger;
+        private readonly string _ancmVersion;
+        private readonly object _syncLock = new object();
+        private readonly string _apphostConfigPath = Path.Combine(Environment.SystemDirectory, "inetsrv", "config", "applicationhost.config");
```
pakrym: Are we planning on doing x64/x86 testing?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,249 @@
...
+            RestoreAppHostConfig();
+
+            _serverManager.CommitChanges();
+
+            await WaitUntilSiteStopped();
+        }
+
+        private async Task WaitUntilSiteStopped()
+        {
+            var i = 0;
```
pakrym: move this into a loop and use IISRetryLimit for logs

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        private void AddTemporaryAppHostConfig()
+        {
+            File.Copy(_apphostConfigPath, _apphostConfigBackupPath);
+            _logger.LogInformation($"Backed up {_apphostConfigPath} to {_apphostConfigBackupPath}");
+        }
+
+        private void RestoreAppHostConfig()
+        {
+            File.Delete(_apphostConfigPath);
```
pakrym: Historically delete always needs a retry.

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
pakrym: Add a random part to it, if tests crashes subsequent backup would override original backup and file will be lost.

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,249 @@
...
+            {
+                _logger.LogInformation($"Could not find {_ancmVersion} section in global modules; creating section.");
+                var addElement = globalModulesSection.GetCollection().CreateElement("add");
+                addElement["name"] = _ancmVersion;
+                addElement["image"] = ancmFile;
+                globalModulesSection.GetCollection().Add(addElement);
+            }
+            else
+            {
+                _logger.LogInformation($"Replacing {_ancmVersion} section in global modules");
```
pakrym: with what?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
jkotalik: I think the site name needs to have more randomness. It is based off of current time up to the nearest second.

test/IISIntegration.FunctionalTests/Utilities/SkipIfNotAllowedToRunIISAttribute.cs
Tratcher: Won't the test fail without elevation?

Tratcher:
What's shared with the express deployer?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
jkotalik: Yes, but the config is shared across 32 and 64 bit for IIS.

jkotalik:
> What's shared with the express deployer?

Modifying apphost.config stuff, enabling inplace test running, ANCM finding, etc.

Tratcher:
Not enough to block a first pass. We can work out homes once we prove this one is working.

jkotalik:
Investigating appveyor failures. Also trying on a feature branch.

natemcmaster:
Can you hold off for an hour or so until I can finish 2.2 branching changes? Also, you may want to retarget this to release/2.2

test/IISIntegration.FunctionalTests/Utilities/IISAncmValidator.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                SkipReason = "IIS tests can only be run on Windows";
+                return;
+            }
+
+            var identity = WindowsIdentity.GetCurrent();
+            var principal = new WindowsPrincipal(identity);
+            if (!principal.IsInRole(WindowsBuiltInRole.Administrator))
```
Tratcher: I didn't see an answer to the elevation question. Do the tests pass without elevation?

test/IISIntegration.FunctionalTests/Utilities/SkipIfNotAllowedToRunIISAttribute.cs
Tratcher: You cannot enable these tests if they're going to fail when running non-elevated. Full Stop.

Is that why the AppVeyor tests are failing?

jkotalik: Resolved offline. WindowsBuiltInRole.Administrator returns true only if your current role is elevated. The skip condition works for skipping builds when running non-admin.

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -199,20 +199,22 @@ private ApplicationPool ConfigureAppPool(string contentRoot)
...
-            var addElement1 = envCollection.CreateElement("add");
-            addElement1["name"] = @"ASPNETCORE_MODULE_DEBUG";
-            addElement1["value"] = @"4";
-            envCollection.Add(addElement1);
-            var addElement2 = envCollection.CreateElement("add");
-            addElement2["name"] = @"ASPNETCORE_MODULE_DEBUG_FILE";
-            addElement2["value"] = Path.Combine(contentRoot, $"{WebSiteName}.txt");
-            envCollection.Add(addElement2);
+
+            AddEnvironmentVariableToAppPool(envCollection, @"ASPNETCORE_MODULE_DEBUG", @"4");
```
pakrym: What about other deployment options env variables?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -0,0 +1,303 @@
...
+            {
+                _logger.LogInformation($"Replacing {_ancmVersion} section in global modules with {ancmFile}");
+                globalConfigElement["image"] = ancmFile;
+            }
+        }
+
+        private void SetModulesSection(Configuration config)
+        {
+            var modulesSection = config.GetSection("system.webServer/modules");
+            var moduleConfigElement = modulesSection.GetCollection().Where(element => (string)element["name"] == _ancmVersion).FirstOrDefault();
```
pakrym: What if you found it but it was wrong?

test/IISIntegration.FunctionalTests/Utilities/IISApplication.cs
jkotalik: Across repos is a non-goal for right now.

Issue #983, opened 2018-06-27
Status code 400 response body is not returned to remote client


pakrym:
Repro: Use MVC controller with the following action:

```
   return new BadRequestObjectResult("Please pass a name on the query string or in the request body");
 
```

And request it from a remote host.

Issue #978, opened 2018-06-27
Different page is served depending if we tried to write reponse after setting error code


pakrym:
`.Response.Write("");` would result in IIS error page while not doing anything would produce an empty output.

PR #985, opened 2018-06-27
Do not override error responses

pakrym:
https://github.com/aspnet/IISIntegration/issues/983

https://github.com/aspnet/IISIntegration/issues/978

### DIFF
--- a/src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cxx
+++ b/src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cxx
@@ -94,7 +94,8 @@ http_set_response_status_code(
     _In_ PCSTR pszReason
 )
 {
-    return pInProcessHandler->QueryHttpContext()->GetResponse()->SetStatus(statusCode, pszReason);
+    return pInProcessHandler->QueryHttpContext()->GetResponse()->SetStatus(statusCode, pszReason, 0, 0, nullptr,
+        true); // fTrySkipCustomErrors
 }
 
 EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
--- a/test/IISIntegration.FunctionalTests/Inprocess/ResponseHeaderTests.cs
+++ b/test/IISIntegration.FunctionalTests/Inprocess/ResponseHeaderTests.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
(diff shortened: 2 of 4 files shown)


test/WebSites/InProcessWebSite/web.config
```diff
@@ -1,10 +1,10 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <system.webServer>
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified" />
     </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="true" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" hostingModel="inprocess">
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" hostingModel="inprocess">
```
jkotalik: Revert?

src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cxx
```diff
@@ -94,7 +94,8 @@ http_set_response_status_code(
     _In_ PCSTR pszReason
 )
 {
-    return pInProcessHandler->QueryHttpContext()->GetResponse()->SetStatus(statusCode, pszReason);
+    return pInProcessHandler->QueryHttpContext()->GetResponse()->SetStatus(statusCode, pszReason, 0, 0, nullptr,
```
jkotalik: I don't see this overload here: https://msdn.microsoft.com/en-us/library/ms690943(v=vs.90).aspx. Undocumented?

jkotalik: Also, do we care about returning the hresult?

test/WebSites/InProcessWebSite/web.config
pakrym: Nope, intentional setting stdoutLogEnabled prevents output going into stdout and test logger. Is it a bug?

jkotalik: Nah that is expected. We don't write the file contents back to stdout after the app is done.

src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cxx
pakrym: Seems to be undocumented, why wouldn't we care?

jkotalik: I don't think we care, it just sucks that it isn't discoverable.

pakrym: Why isn't it? We check it on managed side. We can also wrap them into LOG_IF tho.

jkotalik: No log info as it's on a request basis.

pakrym:
Filed: https://github.com/aspnet/IISIntegration/issues/987

src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cxx
Tratcher: Ping sourabh, we should get any APIs we use documented.

pan-wang: This is known issues. The API is in the public header file. The documentation for couple iis APIs need to be updated.

test/IISIntegration.FunctionalTests/Inprocess/ResponseHeaderTests.cs
```diff
@@ -51,15 +52,30 @@ public ResponseHeaders(IISTestSiteFixture fixture)
         }
 
         [ConditionalTheory]
-        [InlineData(200, "custom", "custom")]
-        [InlineData(500, "", "Internal Server Error")]
-        [InlineData(999, "", "")]
-        public async Task CustomErrorCodeWorks(int code, string reason, string expectedReason)
+        [InlineData(200, "custom", "custom", null)]
```
pan-wang: you really don't need to add a test variation to cover it. Otherwise, we may end up with testing for every IIS API.

PR #997, opened 2018-06-29
Add and use GetEnvironmentVariableValue

pakrym:


jkotalik:
Motivation for change is just adding another helper? Or was there an issue calling expand env variables for debug logging?

pakrym:
ExpandEnvVariable returns the original string if the variable is not found. making atoi throw for example.

jkotalik:
Gotcha, do we want to replace other instances of  Environment::ExpandEnvironmentVariables?

jkotalik:
Also StartupMessagesAreLoggedIntoDebugLogFile failed. Seems to still be flaky. We can disable the test if isn't resolved shortly.

pakrym:
> Gotcha, do we want to replace other instances of Environment::ExpandEnvironmentVariables?

Most of them actually need expanding variables, we these two I just cheated.

src/AspNetCoreModuleV2/CommonLib/Environment.cpp
```diff
@@ -30,3 +30,31 @@ Environment::ExpandEnvironmentVariables(const std::wstring & str)
...
+    std::wstring expandedStr;
+    do
+    {
+        expandedStr.resize(requestedSize);
+        requestedSize = GetEnvironmentVariableW(str.c_str(), expandedStr.data(), requestedSize);
+        if (requestedSize == 0)
+        {
+            throw std::system_error(GetLastError(), std::system_category(), "ExpandEnvironmentStringsW");
+        }
+    } while (expandedStr.size() != requestedSize + 1);
```
jkotalik: I think this while condition is right, but maybe `expandedStr.size() < requestedSize + 1` is a safer check.

Issue #993, opened 2018-06-28
dotnet.exe location optimization regressed


pakrym:
Looking into recent logs from App Services:
```
[aspnetcore.dll] Resolving hostfxr parameters for application: dotnet arguments: .\app2.2.dll path: D:\home\site\wwwroot\
[aspnetcore.dll] Process path D:\home\site\wwwroot\dotnet is dotnet, treating application as portable
[aspnetcore.dll] Resolving absolute path do dotnet.exe from D:\home\site\wwwroot\dotnet
>>>>> [aspnetcore.dll] Invoking where.exe to find dotnet.exe
[aspnetcore.dll] where.exe invocation returned: D:\home\SiteExtensions\AspNetCoreRuntime.2.2.x64\dotnet.exe
D:\Program Files (x86)\dotnet\dotnet.exe

[aspnetcore.dll] Current process bitness type detected as isX64=1
[aspnetcore.dll] Processing entry D:\home\SiteExtensions\AspNetCoreRuntime.2.2.x64\dotnet.exe
[aspnetcore.dll] Binary type 6
[aspnetcore.dll] Found dotnet.exe wia where.exe invocation at D:\home\SiteExtensions\AspNetCoreRuntime.2.2.x64\dotnet.exe
[aspnetcore.dll] Resolving absolute path do hostfxr.dll from D:\home\SiteExtensions\AspNetCoreRuntime.2.2.x64\dotnet.exe
[aspnetcore.dll] hostfxr.dll located at D:\home\SiteExtensions\AspNetCoreRuntime.2.2.x64\host\fxr\2.2.0-preview1-26618-02\hostfxr.dll
[aspnetcore.dll] Resolving hostfxr_main arguments application: D:\home\SiteExtensions\AspNetCoreRuntime.2.2.x64\dotnet.exe arguments: .\app2.2.dll path: D:\home\site\wwwroot\
[aspnetcore.dll] Argument[1] = D:\home\site\wwwroot\app2.2.dll
[aspnetcore.dll] Loading request handler: D:\home\site\wwwroot\runtimes\win-x64\nativeassets\netstandard2.0\aspnetcorev2_inprocess.dll
[aspnetcorev2_inprocess.dll] REQUESTHANDLER_CONFIG::GetConfig, set config to ModuleContext
[aspnetcorev2_inprocess.dll] Resolving hostfxr parameters for application: dotnet arguments: .\app2.2.dll path: D:\home\site\wwwroot\
[...]

PR #998, opened 2018-06-29
Pass parameters to createapplication method

pakrym:
This fixes https://github.com/aspnet/IISIntegration/issues/993 but is pretty hard to test.

### DIFF
--- a/src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
+++ b/src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
@@ -3,6 +3,7 @@
 
 #include "applicationinfo.h"
 
+#include <array>
 #include "proxymodule.h"
 #include "hostfxr_utility.h"
 #include "utility.h"
@@ -222,8 +223,16 @@ APPLICATION_INFO::EnsureApplicationCreated(
                 FINISHED(HRESULT_FROM_WIN32(ERROR_INVALID_FUNCTION));
             }
 
-            FINISHED_IF_FAILED(m_pfnAspNetCoreCreateApplication(m_pServer, pHttpContext->GetApplication(), &pApplication));
-            pApplication->SetParameter(L"InProcessExeLocation", struExeLocation.QueryStr());
+            std::array<APPLICATION_PARAMETER, 1> parameters {
+                {"InProcessExeLocation", struExeLocation.QueryStr()}
+            };
+
+            FINISHED_IF_FAILED(m_pfnAspNetCoreCreateApplication(
+                m_pServer,
+                pHttpContext->GetApplication(),
+                parameters.data(),
+                static_cast<DWORD>(parameters.size()),
+                &pApplication));
 
             m_pApplication = pApplication;
         }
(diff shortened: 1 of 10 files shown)


src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -17,21 +17,31 @@ IN_PROCESS_APPLICATION*  IN_PROCESS_APPLICATION::s_Application = NULL;
...
-    InProcessApplicationBase(pHttpServer),
+    m_fInitialized(FALSE),
     m_pConfig(std::move(pConfig))
 {
     // is it guaranteed that we have already checked app offline at this point?
     // If so, I don't think there is much to do here.
     DBG_ASSERT(pHttpServer != NULL);
     DBG_ASSERT(pConfig != NULL);
 
+    const auto exeLocationParameterName = "InProcessExeLocation";
```
jkotalik: define instead for this name?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -222,8 +223,16 @@ APPLICATION_INFO::EnsureApplicationCreated(
                 FINISHED(HRESULT_FROM_WIN32(ERROR_INVALID_FUNCTION));
             }
 
-            FINISHED_IF_FAILED(m_pfnAspNetCoreCreateApplication(m_pServer, pHttpContext->GetApplication(), &pApplication));
-            pApplication->SetParameter(L"InProcessExeLocation", struExeLocation.QueryStr());
+            std::array<APPLICATION_PARAMETER, 1> parameters {
+                {"InProcessExeLocation", struExeLocation.QueryStr()}
```
jkotalik: This won't copy the exeLocation. I don't like that it's up to the application to copy all pointers in the application_parameter array.

pakrym: I would prefer not to manage lifetime for this array, especially if app doesn't need it.

pakrym: And it was the same in SetParameter before.

jkotalik: IMO if we pass something from the shim to the request handler, ownership should transfer completely. If someone stores the application parameter array and expects the VOID* not to be freed, then we should copy the values.

pakrym: How do you know how to free void* correctly without knowing data type?

jkotalik: We don't, but it would avoid bugs where someone doesn't copy. Like here: https://github.com/aspnet/IISIntegration/pull/998/files#diff-4653e8c4233ce04437f43552addb24f3R42 ; instead of calling Copy, you just assign the string somewhere, it would AV.

pakrym: How do you free the memory after application would need to get recycled?

PR #977, opened 2018-06-27
Remove manual memory management from app_ofline

pakrym:
And add tracing

PR #1001, opened 2018-06-29
Panwang/filewatcher

pan-wang:
remove filewatcher from shim


src/AspNetCoreModuleV2/RequestHandlerLib/application.h
```diff
@@ -17,9 +18,18 @@ class APPLICATION : public IAPPLICATION
         return m_status;
     }
 
-    APPLICATION()
+    APPLICATION(REQUESTHANDLER_CONFIG* pConfig)
         : m_cRefs(1)
     {
+        m_pConfig = pConfig;
```
pakrym: Please don't do manual memory management. You can also just pass the application path in.

src/AspNetCoreModuleV2/OutOfProcessRequestHandler/dllmain.cxx
```diff
@@ -21,6 +21,8 @@ IHttpServer *       g_pHttpServer = NULL;
 HINSTANCE           g_hWinHttpModule;
 HINSTANCE           g_hAspNetCoreModule;
 HANDLE              g_hEventLog = NULL;
+FILE_WATCHER*       g_pFileWatcher = NULL;
```
pakrym: Why do you need it globally? Can't it be per `application` object?

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -99,12 +99,12 @@ CreateApplication(
...
         BOOL disableStartupPage = pConfig->QueryDisableStartUpErrorPage();
 
-        auto pApplication = std::make_unique<IN_PROCESS_APPLICATION>(pServer, std::move(config));
+        auto pApplication = std::make_unique<IN_PROCESS_APPLICATION>(pServer, config.release());
         
         if (FAILED(pApplication->LoadManagedApplication()))
         {
             // Set the currently running application to a fake application that returns startup exceptions.
-            *ppApplication = new StartupExceptionApplication(pServer, disableStartupPage);
+            *ppApplication = new StartupExceptionApplication(pServer, config.release(), disableStartupPage);
```
pakrym: You can't release twice, second time it might be null.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -54,42 +37,119 @@ APPLICATION_INFO::~APPLICATION_INFO()
...
+BOOL APPLICATION_INFO::LoadAppOffline(LPWSTR strFilePath)
+{
+    BOOL            fResult = TRUE;
+    LARGE_INTEGER   li = { 0 };
+    CHAR           *pszBuff = NULL;
+    HANDLE         handle = INVALID_HANDLE_VALUE;
+
+    DBG_ASSERT(strFilePath);
+
+    handle = CreateFile(strFilePath,
```
pakrym: Handle wrapped and logging in this method.

src/AspNetCoreModuleV2/OutOfProcessRequestHandler/outofprocess/outprocessapplication.cpp
```diff
@@ -47,6 +47,16 @@ OUT_OF_PROCESS_APPLICATION::Initialize(
         }
     }
 
+    if (m_pFileWatcherEntry == NULL)
+    {
+        m_pFileWatcherEntry = new  FILE_WATCHER_ENTRY(g_pFileWatcher);
```
jkotalik: Shouldn't the file watcher entry be created in the application?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pakrym: I did it in https://github.com/aspnet/IISIntegration/pull/977/files#diff-0ecd7731bba35e311fac903916b30e44R26

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cxx
```diff
@@ -99,12 +99,12 @@ CreateApplication(
 
         BOOL disableStartupPage = pConfig->QueryDisableStartUpErrorPage();
 
-        auto pApplication = std::make_unique<IN_PROCESS_APPLICATION>(pServer, std::move(config));
+        auto pApplication = std::make_unique<IN_PROCESS_APPLICATION>(pServer, config.release());
```
jkotalik: Why was this changed?

jkotalik:
App offline regressed: AppOfflineDroppedWhileSiteRunning_SiteShutsDown. Make sure the site shuts down when app offline is dropped?

src/AspNetCoreModuleV2/OutOfProcessRequestHandler/dllmain.cxx
pan-wang: You don't want to create so many file watcher threads. A global one, single watcher thread

src/AspNetCoreModuleV2/OutOfProcessRequestHandler/outofprocess/outprocessapplication.cpp
pan-wang: I just want to make the application.h simple and abstract enough. an application may not need do file watching for appoffline.

pakrym: Add `FILE_WATCHING_APPLICATION` then

src/AspNetCoreModuleV2/RequestHandlerLib/requesthandler_config.h
```diff
@@ -209,6 +207,25 @@ class REQUESTHANDLER_CONFIG
         return &m_struConfigPath;
     }
 
+    VOID
+    ReferenceConfig()
```
jkotalik: Why did you add manual Referencing/Dereferencing to the configuration?

pan-wang: in the case of inprocess application LoadManagedApplication failed, we then create a failederrorapplication which shares config object with the previous application. As the config object will be owned by the application, we cannot call delete or use smart point, otherwise double free.

jkotalik: failederrorapplication shouldn't share config with the previous application. Also if that's the case, can't you use a shared pointer?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -54,42 +38,103 @@ APPLICATION_INFO::~APPLICATION_INFO()
...
+APPLICATION_INFO::AppOfflineFound()
 {
-    if (m_pFileWatcherEntry != NULL)
+    ULONGLONG  ulCurrentTime = GetTickCount64();
+    //
+    // we only care about app offline presented. If not, it means the application has started
+    // and is monitoring  the app offline file
+    // we cache the file exist check for 1 minute
+    //
+    if (m_fAppOfflineFound && ulCurrentTime - m_ulLastCheckTime > 1000)
```
jkotalik: This would not check app_offline on the first request if ulCurrentTime - m_ulLastCheckTime < 1000. For example, if have app_offline already in the folder and then send a request, it will not check for app_offline even though it should.

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -54,42 +38,103 @@ APPLICATION_INFO::~APPLICATION_INFO()
...
-    UpdateAppOfflineFileHandle();
 
 Finished:
     return hr;
 }
 
-HRESULT
-APPLICATION_INFO::StartMonitoringAppOffline()
+BOOL
+APPLICATION_INFO::AppOfflineFound()
```
jkotalik: Rename to CheckIfAppOfflinePresent

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
```diff
@@ -145,16 +132,18 @@ class APPLICATION_INFO
...
     HRESULT FindNativeAssemblyFromHostfxr(HOSTFXR_OPTIONS* hostfxrOptions, PCWSTR libraryName, STRU* location);
+    BOOL    LoadAppOffline(LPWSTR strFilePath);
 
     static VOID DoRecycleApplication(LPVOID lpParam);
 
     mutable LONG            m_cRefs;
     STRU                    m_struInfoKey;
+    STRA                    m_strAppOfflineContent;
     BOOL                    m_fAppOfflineFound;
+    BOOL                    m_fAppOfflineLoaded;
```
jkotalik: Is m_fAppOfflineLoaded set to false after app_offline is removed?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -98,73 +143,45 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
...
-                m_pConfiguration->QueryApplicationPath()->QueryStr())))
-            {
-                UTILITY::LogEvent(g_hEventLog,
-                    EVENTLOG_INFORMATION_TYPE,
-                    ASPNETCORE_EVENT_RECYCLE_APPOFFLINE,
-                    strEventMsg.QueryStr());
-            }
-
-            RecycleApplication();
+            LoadAppOffline(strFilePath.QueryStr());
```
jkotalik: You never check the return value of LoadAppOffline(). Doesn't this determine m_fAppOfflineFound ?

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
```diff
@@ -147,15 +124,21 @@ ASPNET_CORE_PROXY_MODULE::OnExecuteRequestHandler(
...
-        if (pApplication != NULL &&
-            pApplication->QueryStatus() != APPLICATION_STATUS::RUNNING &&
-            pApplication->QueryStatus() != APPLICATION_STATUS::STARTING)
+        DBG_ASSERT(pHttpContext);
+        // make sure application is in running state
+        // cannot recreate the application as we cannot reload clr for inprocess
+        if (pApplication->QueryStatus() == APPLICATION_STATUS::OFFLINE)
+        {
+            pApplicationManager->RecycleApplicationFromManager(m_pApplicationInfo->QueryConfig()->QueryConfigPath()->QueryStr());
+            // todo: add appoffline
```
jkotalik: Is this Todo resolved? Doesn't look like it.

src/AspNetCoreModuleV2/RequestHandlerLib/filewatcher.h
```diff
@@ -1,14 +1,24 @@
...
-#include <windows.h>
+#pragma once
 
-#include "stringu.h"
+#include "stdafx.h"
+#include "iapplication.h"
 
 #define FILE_WATCHER_SHUTDOWN_KEY           (ULONG_PTR)(-1)
 #define FILE_WATCHER_ENTRY_BUFFER_SIZE      4096
+#ifndef CONTAINING_RECORD
```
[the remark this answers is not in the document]
pakrym: What about this?

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
pan-wang: yes. updated it

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pan-wang: we may fail to load the app_offline content due to file ACL or the file is exclusively opened by some other proces (e.g., antivirus) but the file does exist. this is why we set m_fAppOfflineFound to true. We will return 503 with empty content

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationinfo.h
pan-wang: once appoffline is detected, the instance of applicationinfo will be recycled which may take time. no need to reset m_fAppOfflineLoaded as we don't want bad thing happen in the middle

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pan-wang: m_ulLastCheckTime is 0, ulCurrentTime is current tick number since machine reboot which cannot be less than 1000 (i.e., 1s)

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -54,42 +38,103 @@ APPLICATION_INFO::~APPLICATION_INFO()
...
+
+    if (handle == INVALID_HANDLE_VALUE)
+    {
+        if (HRESULT_FROM_WIN32(GetLastError()) == ERROR_FILE_NOT_FOUND)
+        {
+            fResult = FALSE;
+        }
+
+        // If file is currenlty locked exclusively by other processes, we might get INVALID_HANDLE_VALUE even though the file exists. In that case, we should return TRUE here.
+        goto Finished;
```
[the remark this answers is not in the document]
pan-wang: OK. This function is the only one do "go finish". In general, I prefer single exit

src/AspNetCoreModuleV2/AspNetCore/Inc/applicationmanager.h
```diff
@@ -97,55 +93,30 @@ class APPLICATION_MANAGER
...
-            if(m_pFileWatcher == NULL)
+            try
             {
-                hr = HRESULT_FROM_WIN32( ERROR_NOT_ENOUGH_MEMORY );
-                goto Finished;
+                m_pApplicationInfoHash = new APPLICATION_INFO_HASH();
             }
-
-            m_pFileWatcher->Create();
+            CATCH_RETURN();
```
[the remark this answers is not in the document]
pan-wang: No. This initialize happened at RegisterModule which much earlier than ExecuteRequest

src/AspNetCoreModuleV2/RequestHandlerLib/filewatcher.h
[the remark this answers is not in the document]
pan-wang: A macro used for IIS hashtable

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pakrym: > we cache the file exist check for 1 minute

One second?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pakrym: Why does it have return value at all then?

src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
pan-wang: yes. one second. a typo

src/AspNetCoreModuleV2/RequestHandlerLib/filewatcher.h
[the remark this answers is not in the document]
pan-wang: OK. Remove it.

Issue #1003, opened 2018-06-29
HostFxr dotnet path should check if dotnet is in the argument before prepending application path.


jkotalik:
I don't believe this is a product issue, however the log is incorrect in showing we looked for dotnet in the application folder.
```
 Resolving hostfxr parameters for application: dotnet arguments: .\InProcessWebSite.dll path: C:\Users\appveyor\AppData\Local\Temp\1\d427e610-b31d-499f-9160-6c9ad3d2222d\
 Process path C:\Users\appveyor\AppData\Local\Temp\1\d427e610-b31d-499f-9160-6c9ad3d2222d\dotnet is dotnet, treating application as portable
 Resolving absolute path do dotnet.exe from C:\Users\appveyor\AppData\Local\Temp\1\d427e610-b31d-499f-9160-6c9ad3d2222d\dotnet
 Invoking where.exe to find dotnet.exe
  
```

PR #1004, opened 2018-07-02
Only run fallback logic for dotnet and dotnet.exe

pakrym:
https://github.com/aspnet/IISIntegration/issues/1003

### DIFF
--- a/src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
+++ b/src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
@@ -128,11 +128,6 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
     fs::path processPath = Environment::ExpandEnvironmentVariables(pcwzProcessPath);
     std::wstring arguments = Environment::ExpandEnvironmentVariables(pcwzArguments);
 
-    if (processPath.is_relative())
-    {
-        processPath = applicationPhysicalPath / processPath;
-    }
-
     // Check if the absolute path is to dotnet or not.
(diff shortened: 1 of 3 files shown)


src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -331,11 +331,18 @@ HOSTFXR_UTILITY::ParseHostfxrArguments(
 
 std::optional<fs::path>
 HOSTFXR_UTILITY::GetAbsolutePathToDotnet(
+     const fs::path & applicationPath,
```
jkotalik: It may be better to pass these by value than by ref as we set processPath=requestedPath immediately.

pakrym: prevent's extra copies

test/IISIntegration.FunctionalTests/Inprocess/StartupTests.cs
```diff
@@ -32,20 +32,18 @@ public StartupTests(ITestOutputHelper output) : base(output)
                 deploymentParameters => deploymentParameters.EnvironmentVariables["DotnetPath"] = _dotnetLocation);
         }
 
-        [ConditionalFact]
-        public async Task InvalidProcessPath_ExpectServerError()
+        [ConditionalTheory]
+        [InlineData("bogus")]
+        [InlineData("c:\\random files\\dotnet.exe")]
+        [InlineData(".\\dotnet.exe")]
```
jkotalik: Add a test for `\\dotnet.exe` (without the dot)

PR #1007, opened 2018-07-02
Only add environment variables if we are running on win8.1 or above.

jkotalik:
I may need to change it to check for win10+ or above; it'll be a bit more work to check it though as Environment.OSVersion always returns 6.2 starting in win8.


pakrym:
I would prefer it to fail or have a skip attribute, silently failing is hard to diagnose.

jkotalik:
We can try a skip attribute, but it would have to only be for tests that use environment variables.

Issue #991, opened 2018-06-28
Incorrect EventLog messages for v2 inproc and outproc


mikeharder:
The EventLog messages are incorrect for both v2 inproc and outproc.  Appears to impact both IIS and IIS Express.  Please include tests in the PR to fix the issue.

For example, here are the event messages and IDs after adding and removing `app_offline.htm`:

# v1 outproc
```
Message: App_offline file 'app_offline.htm' was detected.
Id: 1012

Message: Sent shutdown HTTP message to process '1096' and received http status '202'.
Id: 1006

Message: Application 'MACHINE/WEBROOT/APPHOST/DEFAULT WEB SITE' started process '2216' successfully and is listening on port '24524'.
Id: 1001
```

# v2 outproc
Issues
* Message for ID 1012 is missing application name (`Application ''`)
* Message for IDs 1030 and 1024 are missing

```
Message: Application '' was recycled after detecting the app_offline file.
Id: 1012

Message: Sent shutdown HTTP message to process '2152' and received http status '202'.
Id: 1006

Message: 
Id: 1030

Message: 
Id: 1024

Message: Application '/LM/W3SVC/1/ROOT' started process '372' successfully and process '372' is listening on port '17697'.
Id: 1001
```

# v2 inproc
Issues
* Message for ID 1012 is missing application name (`Application ''`)
* Message for ID 1024 is missing

```
Message: 
Id: 1024

Message: Application '' was recycled after detecting the app_offline file.
Id: 1012
```

# v1 outproc-express
```
Message              : Application 'MACHINE/WEBROOT/APPHOST/DEVELOPMENT WEB
                       SITE' started process '1404' successfully and is
                       listening on port '13749'.
Id                   : 1001
Version              :
Qualifiers           : 0
Level                : 4
Task                 : 0
Opcode               :
Keywords             : 36028797018963968
RecordId             : 131
ProviderName         : IIS Express AspNetCore Module
ProviderId           :
LogName              : Application
ProcessId            :
ThreadId             :
MachineName          : 1dbc4655b01b
UserId               :
TimeCreated          : 6/29/2018 6:00:55 PM
ActivityId           :
RelatedActivityId    :
ContainerLog         : Application
MatchedQueryIds      : {}
Bookmark             : System.Diagnostics.Eventing.Reader.EventBookmark
LevelDisplayName     : Information
OpcodeDisplayName    : Info
TaskDisplayName      :
KeywordsDisplayNames : {Classic}
[...]

PR #1009, opened 2018-07-02
Set correct event providers for ANCM V2

jkotalik:
For #991.


src/AspNetCoreModuleV2/AspNetCore/src/applicationinfo.cpp
```diff
@@ -153,7 +153,7 @@ APPLICATION_INFO::UpdateAppOfflineFileHandle()
             STACK_STRU(strEventMsg, 256);
             if (SUCCEEDED(strEventMsg.SafeSnwprintf(
                 ASPNETCORE_EVENT_RECYCLE_APPOFFLINE_MSG,
-                m_pConfiguration->QueryApplicationPath()->QueryStr())))
```
pakrym: Why change this?

jkotalik: `m_pConfiguration->QueryApplicationPath()->QueryStr()` is an empty string. It should have been QueryApplicationFullPath. I'm going to investigate what else is using QueryApplicationPath().

jkotalik: the filePath is the same as the application path except it has app_offline in the path.

pakrym: Just change it to QueryApplicationPhysicalPath then

Issue #1005, opened 2018-07-02
Get all iisexpress tests to run on full IIS.


jkotalik:
Now that we have hello world operation on CI, we can get IISExpress tests to run on IIS.

jkotalik:
Most of this is done for now. The rest of tests need to be handled on an individual case by case issues.

Issue #995, opened 2018-06-28
Flaky test: ShutdownTests.ServerShutsDownWhenMainExits


jkotalik:
```
System.Net.Http.HttpRequestException : No connection could be made because the target machine actively refused it.
---- System.Net.Sockets.SocketException : No connection could be made because the target machine actively refused it.
   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at System.Net.Http.HttpConnectionPool.CreateConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at System.Net.Http.HttpConnectionPool.WaitForCreatedConnectionAsync(ValueTask`1 creationTask)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Server.IntegrationTesting.LoggingHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken) in /_/test/IISIntegration.FunctionalTests/Utilities/LoggingHandler.cs:line 25
   at Microsoft.AspNetCore.Server.IntegrationTesting.RetryHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken) in /_/test/IISIntegration.FunctionalTests/Utilities/RetryHandler.cs:line 36
   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ShutdownTests.ServerShutsDownWhenMainExits() in /_/test/IISIntegration.FunctionalTests/Inprocess/ShutdownTests.cs:line 27
--- End of stack trace from previous location where exception was thrown ---
----- Inner Stack Trace -----
   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
------- Stdout: -------
[...]

PR #1010, opened 2018-07-02
Port IISExpress tests to run on both IIS and IISExpress

jkotalik:
Part of organizing for #1005. We want to run IISExpress and IIS tests without having to call UseIISDeployer or have multiple InlineData calls. No actual code changes here.


pakrym:
1. There are IISIntegration functional tests there so don't remove entire project
2. We would need 3 project structure eventually (common, IIS, IIS Express) so let's create it now.

jkotalik:
Sure I can create the 3 projects with this PR.

jkotalik:
There are a good set of tests that still haven't been ported to full IIS. I'll think it is better to get this PR in first, then fix the other issues (like log file creation, reusing published output, etc) in separate PRs.

jkotalik:
Summary of changes:
- Rename IISIntegration.FunctionalTests to Common.FunctionalTests
- Add two test projects IISExpress.FunctionalTests and IIS.FunctionalTests
- Single skip condition [SkipIISTestCondition] which will vary between the IIS and IISExpress projects
- namespace IISIntegration.FunctionalTests to Microsoft.AspNetCore.Server.IIS.FunctionalTests.Utilities;
- Various flakiness fixes with the IIS deployer/application

test/IISExpress.FunctionalTests/InProcess/ShutdownTests.cs
```diff
@@ -2,15 +2,14 @@
...
-using Microsoft.AspNetCore.Server.IntegrationTesting;
+using Microsoft.AspNetCore.Server.IIS.FunctionalTests.Utilities;
 using Microsoft.AspNetCore.Testing.xunit;
 using Xunit;
 using Xunit.Abstractions;
 
 namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
 {
-    [SkipIfIISExpressSchemaMissingInProcess]
+    [SkipIISTestCondition]
```
pakrym: This needs a better name.

test/AspNetCoreModuleTests/AspNetCoreModuleTests.vcxproj
```diff
@@ -1,178 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
pakrym: Why was this deleted?

test/IISExpress.FunctionalTests/InProcess/ShutdownTests.cs
jkotalik: SkipIfInstalltionInvalid?

test/AspNetCoreModuleTests/AspNetCoreModuleTests.vcxproj
jkotalik: Unused project that should have been deleted a while ago. I could put it in a separate PR if you really want it to.

pakrym: Nope, if it's intentional it's fine.

test/IISExpress.FunctionalTests/InProcess/ShutdownTests.cs
pakrym: RequiresIIS?

IISIntegration.sln
```diff
@@ -380,6 +382,18 @@ Global
 		{1533E271-F61B-441B-8B74-59FB61DF0552}.Release|x64.Build.0 = Release|x64
 		{1533E271-F61B-441B-8B74-59FB61DF0552}.Release|x86.ActiveCfg = Release|Win32
 		{1533E271-F61B-441B-8B74-59FB61DF0552}.Release|x86.Build.0 = Release|Win32
+		{1F0C8D9B-F47B-41F3-9FC9-6954B6DC7712}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
pakrym: Are P2P dependencies between test projects and native set correctly?

test/Common.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -177,17 +179,31 @@ private void ModifyAspNetCoreSectionInWebConfig(string key, string value)
...
+            if (File.Exists(_apphostConfigPath))
+            {
+                RetryHelper.RetryOperation(
+                    () => File.Delete(_apphostConfigPath),
+                    e => _logger.LogWarning($"Failed to delete file : {e.Message}"),
+                    retryCount: 3,
+                    retryDelayMilliseconds: 100);
+            }
 
             File.Move(_apphostConfigBackupPath, _apphostConfigPath);
```
pakrym: Why some things are in retries and other aren't?

test/Common.FunctionalTests/Utilities/FunctionalTestsBase.cs
```diff
@@ -25,14 +26,7 @@ public FunctionalTestsBase(ITestOutputHelper output = null) : base(output)
...
             // Currently hosting throws if the Servertype = IIS.
-            if (parameters.ServerType == ServerType.IIS)
-            {
-                _deployer = new IISDeployer(parameters, LoggerFactory);
-            }
-            else
-            {
-                _deployer = ApplicationDeployerFactory.Create(parameters, LoggerFactory);
-            }
+            _deployer = DeployerSelector.GetDeployer(parameters, LoggerFactory);
```
pakrym: Are we just silently ignoring parameters.Server?

jkotalik: Yeah, we have to if we wanted minimal code change between the two projects. I can remove setting .Server.

pakrym: Hm, what about exposing current server type as a static?

test/Common.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -205,7 +205,12 @@ private void RestoreAppHostConfig()
                     retryDelayMilliseconds: 100);
             }
 
-            File.Move(_apphostConfigBackupPath, _apphostConfigPath);
+            RetryHelper.RetryOperation(
+                () => File.Move(_apphostConfigBackupPath, _apphostConfigPath),
+                e => _logger.LogError($"Failed to backup apphost.config: {e.Message}"),
+                retryCount: 3,
```
pakrym: Add a helper, 3 and 100 are copied all over.

jkotalik: I wanted to, but it felt weird making a helper of a helper 😅

test/Common.FunctionalTests/Utilities/FunctionalTestsBase.cs
jkotalik: Sure, though I think we need to do something more. Is there a clean way to modify the test name to append the server type?

pakrym: Not that I know of.

IISIntegration.sln
jkotalik: They depend on .IIS, which depends on the native components.

jkotalik:
Previous failure in appveyor: https://github.com/aspnet/IISIntegration/issues/995

jkotalik:
https://ci3.dot.net/job/aspnet_IISIntegration/job/release_2.2/job/windows-Configuration_Release_prtest/53/console was due to .deps.json being used by another process. Seems like a test parallelism issue?

jkotalik:
@pakrym Added windows auth tests.

test/Common.FunctionalTests/AppHostConfig/IISExpress.config
```diff
@@ -177,7 +177,7 @@
 
   <system.webServer>
 
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" hostingModel="[HostingModel]" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" />
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" hostingModel="[HostingModel]" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" />
```
pakrym: Why is forwardWindowsAuthToken disabled for IISExpress?

test/Common.FunctionalTests/OutOfProcess/HelloWorldTest.cs
```diff
@@ -30,6 +31,9 @@ public static TestMatrix TestVariants
             {
                 ApplicationPath = Helpers.GetOutOfProcessTestSitesPath(),
             };
+            // The default in hosting sets windows auth to true.
+            // Set it to the IISExpress.config file
+            deploymentParameters.ServerConfigTemplateContent = File.ReadAllText("IISExpress.config");
```
pakrym: Why not set it in the same place where deployer is selected?

test/Common.FunctionalTests/Utilities/IISFunctionalTestBase.cs
```diff
@@ -18,7 +18,7 @@ public IISFunctionalTestBase(ITestOutputHelper output = null) : base(output)
 
         protected string GetServerConfig(Action<XElement> transform)
         {
-            var doc = XDocument.Load("AppHostConfig/Http.config");
+            var doc = XDocument.Load("IISExpress.config");
```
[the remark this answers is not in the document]
jkotalik: There aren't any tests in Common, should we care?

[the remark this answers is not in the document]
pakrym: But test using it won't work correctly on IIS because it'll pick up wrong base config file.

test/Common.FunctionalTests/OutOfProcess/HelloWorldTest.cs
```diff
@@ -26,9 +27,12 @@ public static TestMatrix TestVariants
         [MemberData(nameof(TestVariants))]
         public async Task HelloWorld(TestVariant variant)
         {
+            // The default in hosting sets windows auth to true.
+            // Set it to the IISExpress.config file
             var deploymentParameters = new DeploymentParameters(variant)
             {
                 ApplicationPath = Helpers.GetOutOfProcessTestSitesPath(),
+                ServerConfigTemplateContent = File.ReadAllText("IISExpress.config")
```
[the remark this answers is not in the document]
jkotalik: There is a Http.config that is used by default if the IISExpress.config isn't used. The only difference is the Http.config turns on windows auth by default. The reason I'm using the IISExpress config here is purely to have windows auth disabled. This also isn't used for IIS at all either. Cleanup needs to happen; I can make it so all tests use the IISExpress.config and only the tests that need windows auth use a modified config, or I can do it in a separate PR.

test/Common.FunctionalTests/Utilities/IISFunctionalTestBase.cs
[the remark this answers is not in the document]
jkotalik: You are right; the IIS tests currently ignore/ don't use the GetServerConfig method.

test/Common.FunctionalTests/OutOfProcess/HelloWorldTest.cs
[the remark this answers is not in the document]
pakrym: Do something like https://github.com/aspnet/IISIntegration/blob/e4b9fd0d340e8bb9c98b1c9afeb39c6ce632e74c/test/Common.FunctionalTests/Utilities/IISFunctionalTestBase.cs#L26 don't add another config just for one line change.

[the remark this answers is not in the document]
jkotalik: I'm trying to move away from using the Http.config file in Hosting. I will make it so it only uses the IISExpress.config and modifies it to enable windows auth.

Issue #1006, opened 2018-07-02
Flush is being called after RQ_NOTIFICATION_CONTINUE is returned


pakrym:
@pan-wang  observed it happening during his testing.

PR #1015, opened 2018-07-03
Sync native calls

pakrym:
https://github.com/aspnet/IISIntegration/issues/1006

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
+++ b/src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
@@ -205,6 +205,7 @@ namespace Microsoft.AspNetCore.Server.IIS.Core
                     return null;
                 }
 
+                // Synchronize access to native methods that might run in parallel with IO loops
                 lock (_contextLock)
                 {
                     return NativeMethods.HttpTryGetServerVariable(_pInProcessHandler, variableName, out var value) ? value : null;
--- a/src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
+++ b/src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
@@ -182,7 +182,7 @@ namespace Microsoft.AspNetCore.Server.IIS.Core
 
         bool IHttpResponseFeature.HasStarted => HasResponseStarted;
 
-        bool IHttpUpgradeFeature.IsUpgradableRequest => _server.IsWebSocketAvailable(_pInProcessHandler);
+        bool IHttpUpgradeFeature.IsUpgradableRequest => true;
 
         bool IFeatureCollection.IsReadOnly => false;
 
@@ -205,7 +205,10 @@ namespace Microsoft.AspNetCore.Server.IIS.Core
                     return null;
                 }
 
-                return NativeMethods.HttpTryGetServerVariable(_pInProcessHandler, variableName, out var value) ? value : null;
+                lock (_contextLock)
+                {
+                    return NativeMethods.HttpTryGetServerVariable(_pInProcessHandler, variableName, out var value) ? value : null;
+                }
             }
         }
 
@@ -265,7 +268,7 @@ namespace Microsoft.AspNetCore.Server.IIS.Core
             _hasRequestReadingStarted = false;
 
             // Upgrade async will cause the stream processing to go into duplex mode
-            AsyncIO = new WebSocketsAsyncIOEngine(_pInProcessHandler);
+            AsyncIO = new WebSocketsAsyncIOEngine(_contextLock, _pInProcessHandler);
 
             await InitializeResponse(flushHeaders: true);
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
```diff
@@ -182,7 +182,7 @@ CancellationToken IHttpRequestLifetimeFeature.RequestAborted
 
         bool IHttpResponseFeature.HasStarted => HasResponseStarted;
 
-        bool IHttpUpgradeFeature.IsUpgradableRequest => _server.IsWebSocketAvailable(_pInProcessHandler);
+        bool IHttpUpgradeFeature.IsUpgradableRequest => true;
```
jkotalik: Don't we still need this check?

src/Microsoft.AspNetCore.Server.IIS/Core/IO/AsyncIOEngine.cs
```diff
@@ -23,8 +24,9 @@ internal partial class AsyncIOEngine : IAsyncIOEngine
         private AsyncWriteOperation _cachedAsyncWriteOperation;
         private AsyncFlushOperation _cachedAsyncFlushOperation;
 
-        public AsyncIOEngine(IntPtr handler)
+        public AsyncIOEngine(object contextSync, IntPtr handler)
         {
+            this._contextSync = contextSync;
```
jkotalik: nit drop the this.

src/Microsoft.AspNetCore.Server.IIS/Core/IO/WebSocketsAsyncIOEngine.cs
```diff
@@ -100,7 +112,10 @@ private void ThrowIfNotInitialized()
 
         public void Dispose()
         {
-            NativeMethods.HttpTryCancelIO(_handler);
+            lock (_contextLock)
```
jkotalik: Why do you need to lock here?

jkotalik:
For #1006

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
pakrym: We reset feature to null if websockets are disabled (https://github.com/aspnet/IISIntegration/blob/3bdb33ccd4dbe5baa3162ca3d3778df1883db093/src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs#L151) so if user was able to get the feature request is upgradeable.

src/Microsoft.AspNetCore.Server.IIS/Core/IO/WebSocketsAsyncIOEngine.cs
pakrym: Locking all IO

davidfowl: We don't know that it's safe to call HttpTryCancelIO overlapped with other calls on the HttpContext

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
```diff
@@ -205,7 +205,10 @@ ClaimsPrincipal IHttpAuthenticationFeature.User
                     return null;
                 }
 
-                return NativeMethods.HttpTryGetServerVariable(_pInProcessHandler, variableName, out var value) ? value : null;
+                lock (_contextLock)
```
davidfowl: Add a comment explaining why we take the lock here.

pakrym:
> functional tests

I can add a test that tries to force the crash but it would be inherently racy and unreliable so I'm a bit sceptical about it.

jkotalik:
As I mentioned offline, at worst you can verify that the behavior that normally would be racy doesn't fail. Ex send a request to the server that obtains server variables and verify it doesn't crash.

pakrym:
:up::date:

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
```diff
@@ -241,7 +240,7 @@ private void EnsureIOInitialized()
             // If at this point request was not upgraded just start a normal IO engine
             if (AsyncIO == null)
             {
-                AsyncIO = new AsyncIOEngine(_pInProcessHandler);
+                AsyncIO = new AsyncIOEngine(_contextLock, _pInProcessHandler);
```
jkotalik: Seems bad to have a shared lock between the HttpContext and AsyncIOEngine. Worth it to move OnStarting and OnCompleted to the AsyncIOEngine?

pakrym: `OnStarting` and `OnCompleted` don't really belong to IOEngine having multiple lock objects might have been more correct but it's just more allocations.

Issue #1014, opened 2018-07-03
Change RegKey checks from "IIS AspNetCore Module" to "IIS AspNetCore Module V2"


jkotalik:
Also we need to check if there are IIS Express RegKeys that should be uniquely checked too.

muratg:
@shirhatti Thoughts on this one?

Please milestone or close it. Thanks.

shirhatti:
Why do we need a unique IIS Express reg key? I'm assuming if there's a vuln, we'd want one global kill switch.
@jkotalik Let's close this

PR #1017, opened 2018-07-05
Rename reg keys from IIS ANCM V1 to V2

jkotalik:
For #1014


pakrym:
Does it make sense to change reg key names? Who are they for?
@shirhatti

jkotalik:
@shirhatti ping.

PR #1023, opened 2018-07-09
Make facts and theories conditional

pakrym:


test/IISExpress.FunctionalTests/OutOfProcess/GlobalVersionTests.cs
```diff
@@ -60,7 +60,7 @@ public class GlobalVersionTests : IISFunctionalTestBase
             Assert.False(response.IsSuccessStatusCode);
         }
 
-        [Theory] // Tests need to publish to change folder locations
+        [ConditionalTheory] // Tests need to publish to change folder locations
```
Tratcher: These comments seem out of place. What line of code are they explaining? It's not the Conditional/Theory.

Issue #1012, opened 2018-07-03
In AppLifetime.ApplicationStarted event, stdout/stderr is not accessable


shhsu:
In E2E test, we registered a callback for

https://github.com/aspnet/AspNetCoreModule/blob/dev/test/AspNetCoreModule.TestSites.Standard/Program.cs#L159

This callback is invoked after `app.Run` is completed hence after `~PipeOutputManager()` is called and `Console.Write` would cause a invalid handle exception

shhsu:
@shirhatti This is a minor ci blocking issue. I can workaround it by commenting out this line in our test app

jkotalik:
Once I repro, I'll take a look at why. I think it's a race condition on closing stdout and calling console.WriteLine (though it could just be an invalid handle issue). It will become much easier to see these issues once full IIS tests are done.

Issue #976, opened 2018-06-26
Don't error out if we fail to set up stdout logging to a file.


jkotalik:
Today, we call CreateDirectory recursively inside of the published output to create the stdout log folder. If this fails, we should log a warning and continue processing requests. 
See: https://github.com/aspnet/IISIntegration/blob/dev/src/AspNetCoreModuleV2/RequestHandlerLib/FileOutputManager.cpp#L122

jkotalik:
> Event log? "Failed to create the stdout log directory, error code: {0}"

Or debug logs.

jkotalik:
Will be handled by #1029

PR #1029, opened 2018-07-09
Fix Console invalid handle issues.

jkotalik:
For #1012 . The reason invalid handle errors were occurring was due to calling Console.WriteLine before and after calling StopOutputRedirection. Managed doesn't pick up the new handle as it caches the handle on first write. 

- I can move the fLoggingEnabled from RegisterCallbacks to http_get_application_properties.
- Once I realized the repro, it failed both on IIS and IISExpress. We can only check stdout without the file with IISExpress as stdout/err is null in IIS.
- I modified the inprocess test site to read an environment variable whether to do an initial write.

Verified this doesn't cause startup failures too @shhsu


src/Microsoft.AspNetCore.Server.IIS/Core/IISNativeApplication.cs
```diff
@@ -31,7 +32,17 @@ public void StopCallsIntoManaged()
             IntPtr requestContext,
             IntPtr shutdownContext)
         {
-            NativeMethods.HttpRegisterCallbacks(_nativeApplication, requestHandler, shutdownHandler, onAsyncCompletion, requestContext, shutdownContext);
+            NativeMethods.HttpRegisterCallbacks(_nativeApplication, requestHandler, shutdownHandler, onAsyncCompletion, requestContext, shutdownContext, out var fLoggingEnabled);
+
+            if (!fLoggingEnabled)
```
shhsu: could the assumption fLoggingEnabled == !PipePutputRedirection ever change?

test/IISExpress.FunctionalTests/InProcess/LoggingIISExpressTests.cs
```diff
@@ -0,0 +1,38 @@
...
+    public class LoggingIISExpressTests : IISFunctionalTestBase
+    {
+        [ConditionalTheory]
+        [InlineData("CheckErrLogFile")]
+        [InlineData("CheckLogFile")]
+        public async Task CheckStdoutLoggingToPipeWithFirstWrite_GetOutput(string path)
+        {
+            var deploymentParameters = Helpers.GetBaseDeploymentParameters();
+            deploymentParameters.PublishApplicationBeforeDeployment = true;
+            deploymentParameters.PreservePublishedApplicationForDebugging = true; // workaround for keeping
```
pakrym: keeping what?

test/WebSites/InProcessWebSite/Program.cs
```diff
@@ -22,5 +25,21 @@ public static void Main(string[] args)
 
             host.Run();
         }
+
+        private static void CheckEnvironmentVariables()
```
pakrym: I don't think we need to make this conditional.

src/Microsoft.AspNetCore.Server.IIS/Core/IISNativeApplication.cs
```diff
@@ -31,7 +32,17 @@ public void StopCallsIntoManaged()
...
+            {
+                // By using the PipeOutputRedirection, after calling RegisterCallbacks,
+                // stdout and stderr will be redirected to NULL. However, if something wrote
+                // to stdout before redirecting, (like a Console.WriteLine during startup),
+                // we need to call Console.Set* to pick up the modified consoles outputs.
+                Console.SetOut(new StreamWriter(Console.OpenStandardOutput()));
```
pakrym: Use same arguments as https://github.com/dotnet/corefx/blob/master/src/System.Console/src/System/Console.cs#L120 as much as possible.

pakrym: Can OpenStandardOutput be anything but `NUL` at this point?

jkotalik: For IISExpress, it wouldn't be `NUL`.

pakrym: We won't redirect anywhere in the case where we have actual terminal right?

src/Microsoft.AspNetCore.Server.IIS/Core/IISNativeApplication.cs
jkotalik: It could, but we can always change the value we return in native to true/false appropriately.

src/Microsoft.AspNetCore.Server.IIS/Core/IISNativeApplication.cs
jkotalik: Yes. but I think some assumptions we have about tests logging to stdout will fail. I'll try and see what happens.

test/WebSites/InProcessWebSite/Program.cs
```diff
@@ -22,5 +25,15 @@ public static void Main(string[] args)
...
+        {
+            var envVariable = Environment.GetEnvironmentVariable("ASPNETCORE_INPROCESS_INITIAL_WRITE");
```
pakrym: Just write in the Main. Would it hurt other tests?

jkotalik: We need to selectively write/not write based on the env variable. Whether we write first or not will change the behavior of what handle is stored by Console.

pakrym: I see

src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cxx
```diff
@@ -16,7 +16,8 @@ register_callbacks(
     _In_ PFN_SHUTDOWN_HANDLER shutdown_handler,
     _In_ PFN_ASYNC_COMPLETION_HANDLER async_completion_handler,
     _In_ VOID* pvRequstHandlerContext,
-    _In_ VOID* pvShutdownHandlerContext
+    _In_ VOID* pvShutdownHandlerContext,
+    _Out_ BOOL* pfIsFileLoggingEnabled
```
pakrym: `pfResetStandardStreams`

test/Common.FunctionalTests/Inprocess/LoggingTests.cs
```diff
@@ -18,11 +17,11 @@ public class LoggingTests : IISFunctionalTestBase
         [ConditionalTheory]
         [InlineData("CheckErrLogFile")]
         [InlineData("CheckLogFile")]
-        public async Task CheckStdoutLogging(string path)
+        public async Task CheckStdoutLoggingToFile(string path)
         {
             var deploymentParameters = Helpers.GetBaseDeploymentParameters();
             deploymentParameters.PublishApplicationBeforeDeployment = true;
-            deploymentParameters.PreservePublishedApplicationForDebugging = true; // workaround for keeping
+            deploymentParameters.PreservePublishedApplicationForDebugging = true; // workaround for keeping published output for log files
```
pakrym: Would this waste disk space infinitely? Set stdout location to a temporary file outside publish location.

test/Common.FunctionalTests/Inprocess/LoggingTests.cs
```diff
@@ -91,5 +80,19 @@ public class LoggingTests : IISFunctionalTestBase
...
         }
 
+        private string GetFileContents(string file)
+        {
+            Assert.NotNull(file);
+
+            string contents = null;
+
+            // RetryOperation doesn't support async lambdas, call synchronous ReadAllText.
+            RetryHelper.RetryOperation(
```
pakrym: You don't need this retry. Server should be stopped and file accessible at the point where we try to read it.

test/IIS.FunctionalTests/InProcess/LoggingIISTests.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
+{
+    public class LoggingIISTests : IISFunctionalTestBase
+    {
+        [ConditionalTheory]
+        [InlineData("CheckErrLogFile", true)]
+        [InlineData("CheckLogFile", true)]
+        [InlineData("CheckErrLogFile", false)]
+        [InlineData("CheckLogFile", false)]
+        public async Task CheckStdoutLoggingToPipeWithFirstWrite(string path, bool firstWrite)
```
pakrym: Why can't we unify this test?

src/Microsoft.AspNetCore.Server.IIS/Core/IISNativeApplication.cs
```diff
@@ -31,7 +32,34 @@ public void StopCallsIntoManaged()
...
+                shutdownContext, 
+                out var consoleOutReset);
+
+            if (consoleOutReset)
+            {
+                // By using the PipeOutputRedirection, after calling RegisterCallbacks,
+                // stdout and stderr will be redirected to NULL. However, if something wrote
+                // to stdout before redirecting, (like a Console.WriteLine during startup),
+                // we need to call Console.Set* to pick up the modified consoles outputs.
+                Console.SetOut(
```
pakrym: Extract method.

src/AspNetCoreModuleV2/RequestHandlerLib/PipeOutputManager.cpp
```diff
@@ -54,7 +54,7 @@ PipeOutputManager::StopOutputRedirection()
 
     if (m_fdPreviousStdOut >= 0)
     {
-        LOG_IF_DUPFAIL(_dup2(m_fdPreviousStdOut, _fileno(stdout)));
+        LOG_LAST_ERROR_IF(SetStdHandle(STD_OUTPUT_HANDLE, (HANDLE)_get_osfhandle(m_fdPreviousStdOut)));
```
pakrym: What did this change? Do we have _dup2 in other places that need to be changed?

jkotalik:
I need #1033 before I merge this to verify IIS logs.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -272,10 +272,7 @@ IN_PROCESS_APPLICATION::LoadManagedApplication
                 goto Finished;
             }
 
-            if (FAILED(hr = m_pLoggerProvider->Start()))
-            {
-                goto Finished;
-            }
+            LOG_IF_FAILED(hr = m_pLoggerProvider->Start());
```
pakrym: Is this https://github.com/aspnet/IISIntegration/issues/976? Is there a test?

test/Common.FunctionalTests/Inprocess/LoggingTests.cs
```diff
@@ -91,6 +81,41 @@ public class LoggingTests : IISFunctionalTestBase
...
+                Helpers.ModifyEnvironmentVariableCollectionInWebConfig(deploymentResult, "ASPNETCORE_INPROCESS_INITIAL_WRITE", firstWriteString);
+            }
+
+            await Helpers.AssertStarts(deploymentResult, path);
+
+            StopServer();
+
+            if (deploymentParameters.ServerType == ServerType.IIS)
+            {
+                Assert.DoesNotContain(TestSink.Writes, context => context.Message.Contains("TEST MESSAGE"));
```
pakrym: What's the point of this verification? This test doesn't seem to assert anything useful on full IIS

test/Common.FunctionalTests/Inprocess/LoggingTests.cs
```diff
@@ -122,5 +147,12 @@ public class LoggingTests : IISFunctionalTestBase
                 File.Delete(secondTempFile);
             }
         }
+
+        private string GetFileContents(string file)
+        {
+            Assert.NotNull(file);
```
pakrym: `Directory.GetFiles(pathToLogs).Single();` would never be null

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
jkotalik: Thinking of how to write a test. Will be tricky.

pakrym: Just put a random non existent path into stdout location

jkotalik: We create the folder path if it doesn't exist. My current thought is to force invalid permissions. This may be tricky for full IIS as we run as local system/admin.

pakrym: just use disk name that doesn't exist or use invalid path characters.

PR #1035, opened 2018-07-11
Cleanup compilation warnings and re-enable asserts

pakrym:
Memory management cleanup here and there.

### DIFF
--- a/src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
+++ b/src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
@@ -48,7 +48,7 @@ Return value:
 }
 
 ASPNET_CORE_PROXY_MODULE::ASPNET_CORE_PROXY_MODULE(
-) : m_pApplicationInfo(NULL), m_pHandler(NULL)
+) : m_pApplicationInfo(nullptr), m_pApplication(nullptr), m_pHandler(nullptr)
 {
 }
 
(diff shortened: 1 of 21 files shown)


src/AspNetCoreModuleV2/IISLib/ntassert.h
```diff
@@ -11,6 +11,10 @@
     #undef ASSERT
 #endif
 
+#ifdef DEBUG
```
jkotalik: Can we have this in debugutil?

pakrym: Why?

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
```diff
@@ -126,13 +109,10 @@ ASPNET_CORE_PROXY_MODULE::OnExecuteRequestHandler(
...
+        IREQUEST_HANDLER* pHandler;
         // Create RequestHandler and process the request
-        hr = pApplication->CreateHandler(pHttpContext, &m_pHandler);
-
-        if (FAILED(hr))
-        {
-            goto Finished;
-        }
+        FINISHED_IF_FAILED(pApplication->CreateHandler(pHttpContext, &pHandler));
+        m_pHandler.reset(pHandler);
```
jkotalik: Why use reset here?

src/AspNetCoreModuleV2/IISLib/ntassert.h
jkotalik: Nvrm I don't think it would work defining it elsewhere. I just wanted our debug stuff to be centralized.

src/AspNetCoreModuleV2/AspNetCore/src/proxymodule.cxx
pakrym: You can't just assign unique_ptr it's either calling ctor or calling `.reset` and reset is much shorter.

PR #1045, opened 2018-07-12
Fix CI test failures

jkotalik:
I choked and didn't realize that environment variables were being used to set state on these tests. Fortunately we can set the environment variables on the web.config and the tests will still work. Also, a skip condition I created was inversed.


test/Common.FunctionalTests/Utilities/IISApplication.cs
```diff
@@ -248,8 +248,38 @@ private void AddEnvironmentVariables(string contentRoot, ApplicationPool pool)
             }
             catch (COMException comException)
             {
-                _logger.LogInformation($"Could not add environment variables to worker process: {comException.Message}");
+                // If we cannot set environment variables on the app pool, set them on the web.config
```
pakrym: I think we shouldn't mix them and automatically fall back. Global environment variables have different effect on some features and tests should be explicit in what they are trying to set.

test/Common.FunctionalTests/Utilities/SkipIISAttribute.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
     [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method)]
     public sealed class SkipIISAttribute : Attribute, ITestCondition
```
pakrym: We don't need this attribute. We need capability aware one. Like in:

`[RequireIISCapability(IISCapability.PoolEnvironmentVariables | IISCapability.WebSockets_]`

jkotalik: Yeah this is significantly better.

test/IIS.FunctionalTests/RequiresIISAttribute.cs
```diff
@@ -68,6 +73,56 @@ static RequiresIISAttribute()
                 .Any(n => "hostingModel".Equals(n.Attribute("name")?.Value, StringComparison.Ordinal));
 
             _skipReason = _isMet ? null : "IIS schema needs to be upgraded to support ANCM.";
+
+            _websocketsAvailable = File.Exists(Path.Combine(Environment.SystemDirectory, "inetsrv", "config", "iiswsock.dll"));
```
pakrym: 👍

test/IIS.FunctionalTests/RequiresIISAttribute.cs
```diff
@@ -8,14 +8,19 @@
...
 
 namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
 {
     [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method)]
     public sealed class RequiresIISAttribute : Attribute, ITestCondition
     {
-        private static readonly bool _isMet;
-        public static readonly string _skipReason;
+        private static bool _isMet;
+        private static readonly string _skipReason;
```
pakrym: This shouldn't be static.

test/IIS.FunctionalTests/RequiresIISAttribute.cs
```diff
@@ -8,14 +8,19 @@
...
+using Microsoft.Win32;
 
 namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
 {
     [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method)]
     public sealed class RequiresIISAttribute : Attribute, ITestCondition
     {
-        private static readonly bool _isMet;
-        public static readonly string _skipReason;
+        private static bool _isMet;
```
pakrym: Neither should this.

jkotalik: It has to be to be set in a static constructor.

pakrym: And then you assign it in non-static ctor?

jkotalik: I modify them in the non-static ctor. What you recommend assigning in the static constructor then? Something needs to be static.

pakrym: Put feature checks (_poolEnvironmentVariablesAvailable  etc.) into static ctor and skip reason and _isMet calculation into non-static.

Issue #982, opened 2018-06-27
Inproc should write to event log at module startup


mikeharder:
Inproc should write to event log at module startup, to help users know the module has started successfully.  Outproc already does this:

```
Log Name:      Application
Source:        IIS Express AspNetCore Module
Date:          6/18/2018 11:24:33 AM
Event ID:      1001
Task Category: None
Level:         Information
Keywords:      Classic
User:          N/A
Computer:      [redacted]
Description:
Application 'MACHINE/WEBROOT/APPHOST/HTTPTESTSITE' started process '11140' successfully and is listening on port '31020'.
Event Xml:
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
  <System>
    <Provider Name="IIS Express AspNetCore Module" />
    <EventID Qualifiers="0">1001</EventID>
    <Level>4</Level>
    <Task>0</Task>
    <Keywords>0x80000000000000</Keywords>
    <TimeCreated SystemTime="2018-06-18T18:24:33.746872100Z" />
    <EventRecordID>31461</EventRecordID>
    <Channel>Application</Channel>
    <Computer>[redacted]</Computer>
    <Security />
  </System>
  <EventData>
    <Data>Application 'MACHINE/WEBROOT/APPHOST/HTTPTESTSITE' started process '11140' successfully and is listening on port '31020'.</Data>
  </EventData>
</Event>
```

mikeharder:
Verified in containers

PR #1047, opened 2018-07-12
Add startup event log for ANCM inproc

jkotalik:
For #982 . Can't effectively add tests until we have container infra setup.


pakrym:
Add a test helper method VerifyEventLogEvent() that uses debug log for now. We will switch to something better as soon as we can.

test/Common.FunctionalTests/Utilities/Helpers.cs
```diff
@@ -108,5 +110,13 @@ public static DeploymentParameters GetBaseDeploymentParameters(string site = nul
         private static string GetWebConfigFile(IISDeploymentResult deploymentResult)
             => Path.Combine(deploymentResult.DeploymentResult.ContentRoot, "web.config");
 
-    }
+        public static async Task AssertStarts(IISDeploymentResult deploymentResult, string path)
```
pakrym: set path to "/HelloWorld" by default

mikeharder:
@jkotalik: I can manually test in a container after this is merged and a hosting bundle is built.

jkotalik:
IMO it's low priority to manually verify this. But feel free to if you think it's worth it.

jkotalik:
>   Total tests: Unknown. Passed: 118. Failed: 0. Skipped: 1.

Appveyor failed to run all iisexpress tests. Maybe due to calling exit(0)?

jkotalik:
https://github.com/aspnet/IISIntegration/issues/1049

PR #1050, opened 2018-07-12
[2.1]: pin baseline package versions to 2.1.2

natemcmaster:
Part of https://github.com/aspnet/Home/issues/3316

This pins package versions to the 2.1.2 baseline. Universe will not override variables in the 'Pinned' section. This helps ensure that this repo does not upgrade its dependency versions for all future patches of 2.1.

### DIFF
--- a/version.props
+++ b/version.props
@@ -1,6 +1,6 @@
 ﻿<Project>
   <PropertyGroup>
-    <VersionPrefix>2.1.2</VersionPrefix>
+    <VersionPrefix>2.1.1</VersionPrefix>
     <VersionSuffix>rtm</VersionSuffix>
     <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' == 'rtm' ">$(VersionPrefix)</PackageVersion>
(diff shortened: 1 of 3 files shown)


jkotalik:
@pakrym this is how we should pin versions in our feature/2.1-oob branch. I'll probably cherry-pick/ remerge this commit into that branch.

version.props
```diff
@@ -1,6 +1,6 @@
 ﻿<Project>
   <PropertyGroup>
-    <VersionPrefix>2.1.2</VersionPrefix>
+    <VersionPrefix>2.1.1</VersionPrefix>
```
jkotalik: Why is this 2.1.1?

pakrym:
Yep, exactly

version.props
natemcmaster: Per policy change in https://github.com/aspnet/Home/issues/3316, IISIntegration will not need to ship new packages in 2.1.3...unless I'm mistaken. Can you confirm you're not expecting IISIntegration patches to land in 2.1.3?

jkotalik: No patches in 2.1.3. And glad we are making this change, though the jagged edges just move to a new place 😢 .

PR #1063, opened 2018-07-16
Move config operations from helpers to functions on the deployment parameters

jkotalik:
This change modifies the way we apply configuration changes on the web.config and applicationhost.config files. This helps with writing tests that need to modify configuration. Also, all web.config modification should be added before deployment, rather than after, which intuitively makes more sense.


src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployer.cs
```diff
@@ -22,8 +22,17 @@ public partial class IISDeployer : ApplicationDeployer
...
+            IISDeploymentParameters = (IISDeploymentParameters)deploymentParameters;
         }
 
+        public IISDeployer(IISDeploymentParameters deploymentParameters, ILoggerFactory loggerFactory)
+            : base(deploymentParameters, loggerFactory)
+        {
+            IISDeploymentParameters = deploymentParameters;
+        }
+
+        public IISDeploymentParameters IISDeploymentParameters { get; }
```
jkotalik: I would like to have IISDeployer and IISExpressDeployer have a shared base class, however I would prefer to have it in a separate PR.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeploymentParameters.cs
```diff
@@ -0,0 +1,173 @@
...
+                        .Single()
+                        .SetAttributeValue("protocol", "https");
+
+                    element.Descendants("access")
+                        .Single()
+                        .SetAttributeValue("sslFlags", "Ssl, SslNegotiateCert");
+                });
+        }
+
+        public void AddWindowsAuthToServerConfig()
```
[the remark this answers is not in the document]
jkotalik: Extensions to DeploymentParameters?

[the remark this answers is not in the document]
pakrym: `IISDeploymentParameters`

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/ApplicationDeployerFactory.cs
```diff
@@ -18,7 +18,7 @@ public class IISApplicationDeployerFactory
         /// <param name="deploymentParameters"></param>
         /// <param name="loggerFactory"></param>
         /// <returns></returns>
-        public static ApplicationDeployer Create(DeploymentParameters deploymentParameters, ILoggerFactory loggerFactory)
+        public static ApplicationDeployer Create(IISDeploymentParameters deploymentParameters, ILoggerFactory loggerFactory)
```
Tratcher: This is a non-starter, it breaks every other repo.

jkotalik: Agreed. Thoughts on how we are going to have ServerTests use the IISApplicationDeployerFactory?

pakrym: It already does. Just don't force argument to be `IISDeploymentParameters `

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployer.cs
```diff
@@ -53,28 +56,29 @@ public override void Dispose()
...
 
-                WebConfigHelpers.AddDebugLogToWebConfig(contentRoot, Path.Combine(contentRoot, $"{_application.WebSiteName}.txt"));
-
                 var uri = TestUriHelper.BuildTestUri(ServerType.IIS, DeploymentParameters.ApplicationBaseUriHint);
                 // To prevent modifying the IIS setup concurrently.
                 await _application.StartIIS(uri, contentRoot);
 
                 // Warm up time for IIS setup.
-                Logger.LogInformation("Successfully finished IIS application directory setup.");
+                Logger.LogInformation("Successfully finished IIS application directory seWebConfigBasedEnvironmentVariablestup.");
```
pakrym: nit

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployerBase.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace Microsoft.AspNetCore.Server.IntegrationTesting.IIS
+{
+    public class IISDeployerBase : ApplicationDeployer
+    {
+        public IISDeploymentParameters IISDeploymentParameters { get; }
+
+        public IISDeployerBase(DeploymentParameters deploymentParameters, ILoggerFactory loggerFactory)
+            : base(deploymentParameters, loggerFactory)
+        {
+            if (deploymentParameters is IISDeploymentParameters)
```
pakrym: What if it's not?

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployerBase.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            var element = XDocument.Parse(serverConfig);
+            foreach (var action in IISDeploymentParameters.ServerConfigActionList)
+            {
+                action.Invoke(element.Element("configuration"));
+            }
+            return element.ToString();
+        }
+
+        public override Task<DeploymentResult> DeployAsync()
```
pakrym: Make class abstract?

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeploymentParameterExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public static void ModifyAttributeInWebConfig(this IISDeploymentParameters parameters, string key, string value, string section)
+        {
+            parameters.WebConfigActionList.Add((element) =>
+            {
+                element.SetAttributeValue(key, value);
+            });
+        }
+
+        public static void ModifyHandlerSectionInWebConfig(this IISDeploymentParameters parameters, string handlerVersionValue)
```
pakrym: Why does anyone need this?

pakrym: Considering we have HandlerSettings?

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeploymentParameters.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        public IISDeploymentParameters(
+           string applicationPath,
+           ServerType serverType,
+           RuntimeFlavor runtimeFlavor,
+           RuntimeArchitecture runtimeArchitecture)
+            : base(applicationPath, serverType, runtimeFlavor, runtimeArchitecture)
+        {
+            WebConfigActionList = new List<Action<XElement>>() { AddWebConfigEnvironmentVariables(), AddHandlerSettings() };
```
pakrym: Extract method.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISExpressDeployer.cs
```diff
@@ -277,13 +282,15 @@ private void PrepareConfig(string contentRoot, string dllRoot, int port)
                 ModifyHandlerSectionInWebConfig(key: "modules", value: DeploymentParameters.AncmVersion.ToString());
```
pakrym: There is still a mixup of old and new style web config modification.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployerBase.cs
jkotalik: We can't call RunWebConfigActions or RunServerConfigActions. Right now I check for null in those methods.

pakrym: ugh

pakrym: Is this what we are going with long-term?

jkotalik: It's ugly for sure. I think it's probably worth it now to add the copy constructor.

jkotalik: In this repo yes but in other repos no.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployer.cs
```diff
@@ -48,24 +51,25 @@ public override void Dispose()
                 StartTimer();
 
                 var contentRoot = string.Empty;
-                if (string.IsNullOrEmpty(DeploymentParameters.ServerConfigTemplateContent))
+                if (string.IsNullOrEmpty(IISDeploymentParameters.ServerConfigTemplateContent))
```
Tratcher: All of these uses of IISDeploymentParameters will null ref if you use the wrong constructor.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeploymentParameters.cs
```diff
@@ -0,0 +1,116 @@
...
+            : base(applicationPath, serverType, runtimeFlavor, runtimeArchitecture)
+        {
+            WebConfigActionList = CreateDefaultWebConfigActionList();
+        }
+
+        public IISDeploymentParameters(DeploymentParameters parameters)
+        {
+            foreach (var propertyInfo in typeof(IISDeploymentParameters).GetProperties())
+            {
+                propertyInfo.SetValue(this, propertyInfo.GetValue(parameters));
```
Tratcher: Not all of them are settable, which tells me you've never run this code path.
https://github.com/aspnet/Hosting/blob/8836de411408257dd75f5db5f7d6724bfb0fab9e/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs#L157-L162

Tratcher: It would be better to add a copy constructor to DeploymentParameters that explicitly set all of the properties and then call base from here.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISExpressDeployer.cs
```diff
@@ -103,35 +108,35 @@ public IISExpressDeployer(DeploymentParameters deploymentParameters, ILoggerFact
...
+                    IISDeploymentParameters.PublishApplicationBeforeDeployment = true;
                 }
-                else if (DeploymentParameters.RuntimeFlavor == RuntimeFlavor.Clr
-                    && DeploymentParameters.RuntimeArchitecture == RuntimeArchitecture.x86)
+                else if (IISDeploymentParameters.RuntimeFlavor == RuntimeFlavor.Clr
+                    && IISDeploymentParameters.RuntimeArchitecture == RuntimeArchitecture.x86)
                 {
                     // x64 is the default. Publish to rebuild for the right bitness
-                    DeploymentParameters.PublishApplicationBeforeDeployment = true;
+                    IISDeploymentParameters.PublishApplicationBeforeDeployment = true;
```
Tratcher: Why bother replacing most of these fields?

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployer.cs
jkotalik: Yeah, the IISDeployerBase was incorrect by not using the copy constructor. Good catch.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISExpressDeployer.cs
```diff
@@ -272,21 +277,23 @@ private void PrepareConfig(string contentRoot, string dllRoot, int port)
             if (DeploymentParameters.PublishApplicationBeforeDeployment)
             {
                 // For published apps, prefer the content in the web.config, but update it.
-                ModifyAspNetCoreSectionInWebConfig(key: "hostingModel",
-                    value: DeploymentParameters.HostingModel == HostingModel.InProcess ? "inprocess" : "");
-                ModifyHandlerSectionInWebConfig(key: "modules", value: DeploymentParameters.AncmVersion.ToString());
+                IISDeploymentParameters.ModifyAspNetCoreSectionInWebConfig(key: "hostingModel",
```
Tratcher: More reverts

jkotalik: That one is required 😄

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeploymentParameters.cs
```diff
@@ -0,0 +1,123 @@
...
+                }
+
+                foreach (var handlerSetting in HandlerSettings)
+                {
+                    CreateOrSetElement(element, handlerSetting.Key, handlerSetting.Value, "handlerSetting");
+                }
+            };
+        }
+
+        public static void CreateOrSetElement(XElement rootElement, string name, string value, string elementName)
```
pakrym: private?

PR #1064, opened 2018-07-16
Use less global variables in shim

pakrym:
1. Remove all of them from precompiled headers
2. Move some to instance fields
3. Move others to statics


pakrym:
@dotnet-bot test this please

Issue #1068, opened 2018-07-17
IISIntegration tests are running on mac/ubuntu agents


pakrym:
/cc @jkotalik

PR #1070, opened 2018-07-17
Set IsMet and SkipReason with empty constructor to skip correctly.

jkotalik:
For #1068. Empty constructor didn't set the _isMet or _skipReason flags.


pakrym:
Is Jenkins failing because of this too?

jkotalik:
@pakrym we shall see.

jkotalik:
Yep.

Issue #1026, opened 2018-07-09
Extend app_offline tests to run on IIS


pakrym:

jkotalik:
@pakrym this is done?

pakrym:
Done in https://github.com/aspnet/IISIntegration/pull/1212

PR #1071, opened 2018-07-17
Fix another shutdown race and appofline race

pakrym:
Shutdown race was between ExtractApplication and EnsureApplicationCreated, the application might have been freed in between the calls.


test/IISExpress.FunctionalTests/InProcess/AppOfflineTests.cs
```diff
@@ -29,19 +31,96 @@ public class AppOfflineTests : IISFunctionalTestBase
...
+            File.WriteAllText(Path.Combine(deploymentResult.DeploymentResult.ContentRoot, "Microsoft.AspNetCore.Server.IIS.dll"), "");
+
+            var result = await deploymentResult.HttpClient.GetAsync("/");
+            Assert.Equal(500, (int)result.StatusCode);
+            Assert.Contains("500.30", await result.Content.ReadAsStringAsync());
+
+            AddAppOffline(deploymentResult.DeploymentResult.ContentRoot);
+            await AssertStopsProcess(deploymentResult);
+        }
+
```
jkotalik: nit new line.

test/IISExpress.FunctionalTests/InProcess/AppOfflineTests.cs
```diff
@@ -5,11 +5,13 @@
 using System.IO;
 using System.Net;
 using System.Net.Http;
+using System.Threading;
```
jkotalik: I don't believe these tests will be sufficient. I think we need something for app_offline with full IIS for this PR. We may also want to consider adding IIS specific shutdown tests as that can be triggered via configuration changes or global recycle.

src/AspNetCoreModuleV2/AspNetCore/ServerErrorApplication.h
```diff
@@ -0,0 +1,48 @@
...
+        PollingAppOfflineApplication(pApplication, PollingAppOfflineApplicationMode::StopWhenAdded)
+    {
+        m_status = APPLICATION_STATUS::RUNNING;
+    }
+
+    ~ServerErrorApplication() = default;
+
+    HRESULT CreateHandler(IHttpContext * pHttpContext, IREQUEST_HANDLER ** pRequestHandler) override
+    {
+        *pRequestHandler = new ServerErrorHandler(pHttpContext, m_HR);
```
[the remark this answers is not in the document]
pakrym: Can't do `make_unique` here, it's shim<->handler interrop.

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -84,7 +66,6 @@ APPLICATION_INFO::EnsureApplicationCreated(
             LOG_INFO("Application went offline");
             // Application that went offline
             // are supposed to recycle themselves
```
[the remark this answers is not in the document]
pakrym: Why?

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -84,7 +66,6 @@ APPLICATION_INFO::EnsureApplicationCreated(
...
-            m_pApplication->DereferenceApplication();
```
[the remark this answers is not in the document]
pakrym: Yes.

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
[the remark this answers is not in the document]
jkotalik: oops meant to delete this comment <_<

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -143,8 +116,15 @@ APPLICATION_INFO::EnsureApplicationCreated(
             EVENTLOG_ERROR_TYPE,
             ASPNETCORE_EVENT_ADD_APPLICATION_ERROR,
             ASPNETCORE_EVENT_ADD_APPLICATION_ERROR_MSG,
-            pHttpContext->GetApplication()->GetApplicationId(),
+            httpApplication.GetApplicationId(),
             hr);
+
+        m_pApplication.reset(new ServerErrorApplication(httpApplication, hr));
```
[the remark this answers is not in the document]
pakrym: more like `std::make_unique<IAPPLICATION, IAPPLICATION_DELETER>(httpApplication, hr));`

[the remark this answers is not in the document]
pakrym: No, because we need custom deleter.

src/AspNetCoreModuleV2/AspNetCore/ServerErrorApplication.h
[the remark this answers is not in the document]
jkotalik: We should be committing to not calling `new`. We can do `make_unique` and then `release()`?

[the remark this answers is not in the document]
pakrym: we immediately return after this call and caller takes the ownership, what's the point?

test/IISExpress.FunctionalTests/InProcess/AppOfflineTests.cs
pakrym: Moved and enabled ones the work. Filed https://github.com/aspnet/IISIntegration/issues/1074 in addition to existing https://github.com/aspnet/IISIntegration/issues/1026

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
```diff
@@ -0,0 +1,84 @@
...
+    pResponse->SetStatus(503, "Service Unavailable", 0, S_OK, nullptr, TRUE);
+    pResponse->SetHeader("Content-Type",
+        "text/html",
+        static_cast<USHORT>(strlen("text/html")),
+        FALSE
+    );
+
+    DataChunk.DataChunkType = HttpDataChunkFromMemory;
+    DataChunk.FromMemory.pBuffer = m_strAppOfflineContent.data();
+    DataChunk.FromMemory.BufferLength = static_cast<ULONG>(m_strAppOfflineContent.size());
```
mikeharder: Assuming `m_strAppOfflineContent` can be updated at any time, there is a potential race between `m_strAppOfflineContent.data()` and `m_strAppOfflineContent.size()`.

mikeharder: @jkotalik had a good idea to add a functional test which deletes and updates `app_offline.htm` in a tight loop for several seconds while simultaneously executing requests, to trigger possible race conditions.  Can we include this test in the PR?

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.h
```diff
@@ -0,0 +1,44 @@
...
+
+    REQUEST_NOTIFICATION_STATUS OnExecuteRequestHandler() override;
+
+private:
+    IHttpContext* m_pContext;
+    std::string m_strAppOfflineContent;
+};
+
+
+class AppOfflineApplication: public PollingAppOfflineApplication
```
mikeharder: This naming seems confusing to me.  I would expect `PollingAppOfflineApplication` to extend `AppOfflineAppliction` based on the naming (prefix indicates subclass), rather than the current implementation which is the opposite.

src/AspNetCoreModuleV2/CommonLib/PollingAppOfflineApplication.cpp
```diff
@@ -7,25 +7,16 @@
...
-
-    return S_OK;
-}
-
 APPLICATION_STATUS PollingAppOfflineApplication::QueryStatus()
 {
     if (AppOfflineExists())
     {
-        return APPLICATION_STATUS::RUNNING;
+        return m_mode == StopWhenRemoved ? APPLICATION_STATUS::RUNNING : APPLICATION_STATUS::RECYCLED;
```
mikeharder: I think a cleaner design would be to remove `m_mode` from `PollingAppOfflineApplication`, and instead have the subclasses `AppOfflineApplication` and `ServerErrorApplication` implement this logic.  It took me a while to understand how the `m_mode` was being used by the different subclasses.

src/AspNetCoreModuleV2/RequestHandlerLib/filewatcher.cpp
```diff
@@ -354,22 +354,29 @@ FILE_WATCHER_ENTRY::Monitor(VOID)
...
 
-    if (!ReadDirectoryChangesW(_hDirectory,
-        _buffDirectoryChanges.QueryPtr(),
-        _buffDirectoryChanges.QuerySize(),
-        FALSE,        // Watching sub dirs. Set to False now as only monitoring app_offline
-        FILE_NOTIFY_VALID_MASK & ~FILE_NOTIFY_CHANGE_LAST_ACCESS,
-        &cbRead,
-        &_overlapped,
-        NULL))
+    // Check if file exist before polling for changes
```
mikeharder: Why is this check necessary?

test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -29,32 +30,114 @@ public class AppOfflineTests : IISFunctionalTestBase
...
+            // Set file content to empty so it fails at runtime
+            File.WriteAllText(Path.Combine(deploymentResult.DeploymentResult.ContentRoot, "Microsoft.AspNetCore.Server.IIS.dll"), "");
 
-            await AssertAppOffline(deploymentResult, expectedResponse);
+            var result = await deploymentResult.HttpClient.GetAsync("/");
+            Assert.Equal(500, (int)result.StatusCode);
+            Assert.Contains("500.30", await result.Content.ReadAsStringAsync());
+
+            AddAppOffline(deploymentResult.DeploymentResult.ContentRoot);
+            AssertStopsProcess(deploymentResult);
```
mikeharder: When `app_offline` is added, why is the expected behavior different if the site failed to start in the shim or in the request handler?  Shouldn't the user-visible behavior be the same?

test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -29,32 +30,114 @@ public class AppOfflineTests : IISFunctionalTestBase
...
+        }
+
+        
+        [ConditionalFact]
+        [RequiresIIS(IISCapability.ShutdownToken)]
+        public async Task AppOfflineDroppedWhileSiteFailedToStart_SiteStops_OutOfProcess()
```
mikeharder: First, I think the test name should contain `_AppOfflineServed_` instead of `_SiteStops`, based on the actual implementation of the test.

Second, should this test be combined with `AppOfflineDroppedWhileSiteFailedToStartInShim_AppOfflineServed_InProcess` as a conditional theory with `HostingModel` parameters?  The tests are very similar.

test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -29,32 +30,114 @@ public class AppOfflineTests : IISFunctionalTestBase
...
+            Assert.Equal(502, (int)result.StatusCode);
+
+            AddAppOffline(deploymentResult.DeploymentResult.ContentRoot);
+            await AssertAppOffline(deploymentResult);
+            AssertFilesNotLocked(deploymentResult);
         }
 
         [ConditionalFact]
+        [RequiresIIS(IISCapability.ShutdownToken)]
+        public async Task AppOfflineDroppedWhileSiteStarting_SiteShutsDown_InProcess()
```
mikeharder: I like the idea of this test, but from reading the implementation I'm afraid it might be too unreliable.  Does the test depend on hitting a certain race condition within 10 tries (the `for` loop), and if it doesn't then the test fails?  Are we confident this won't become a flaky test?

test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -67,6 +150,10 @@ public class AppOfflineTests : IISFunctionalTestBase
                 RemoveAppOffline(deploymentResult.DeploymentResult.ContentRoot);
                 await AssertRunning(deploymentResult);
             }
+
+            AddAppOffline(deploymentResult.DeploymentResult.ContentRoot);
+            await AssertAppOffline(deploymentResult);
+            AssertFilesNotLocked(deploymentResult);
```
mikeharder: Since `AddAppOffline` and `AssertAppOffline` are already run 5 times in the loop above, I assume this new code is really about `AssertFilesNotLocked`.  If so, why not just add this inside the existing loop, rather than adding another iteration with this one additional check?

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
```diff
@@ -0,0 +1,84 @@
...
+    return S_OK;
+}
+
+HRESULT AppOfflineApplication::OnAppOfflineFound()
+{
+    LARGE_INTEGER   li = {};
+
+    HandleWrapper<InvalidHandleTraits> handle = CreateFile(m_appOfflineLocation.c_str(),
+                        GENERIC_READ,
+                        FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
```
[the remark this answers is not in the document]
mikeharder: It appears the .NET Framework 4 app_offline code only sets `FileShare.Read`:

https://referencesource.microsoft.com/#System.Web/HttpRuntime.cs,1254

I believe this will lock the file while it's being read, which would prevent a publish from modifying or deleting the file.  However, this might be fine since publish should retry failed operations, and the file read should be very quick (a few ms).  I would match the behavior of .NET Framework 4, unless we have a good reason to use different behavior with ANCM.

test/Common.FunctionalTests/AppOfflineTests.cs
pakrym: Because `AssertFilesNotLocked` deletes published output so there would be no restarting the site.

mikeharder: OK, I recommend renaming this method to make it clear it deletes the output.  `Assert*` methods are typically assumed to have no side effects.

test/Common.FunctionalTests/AppOfflineTests.cs
pakrym: > Does the test depend on hitting a certain race condition within 10 tries 

Yes. Unfortunately, this is as close as I can get to reproing the issue. Additionally considering how long it takes the site to start it should be hitting the race pretty frequently.

>  Are we confident this won't become a flaky test

No, but I'd like to test this scenario. If test becomes flaky I'll try to harden it even more.

test/Common.FunctionalTests/AppOfflineTests.cs
pakrym: Will do.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISApplication.cs
```diff
@@ -148,7 +148,7 @@ private void ModifyAspNetCoreSectionInWebConfig(string key, string value)
             {
                 return;
             }
-
+
```
pakrym: nit

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
```diff
@@ -0,0 +1,84 @@
...
+                        nullptr,
+                        OPEN_EXISTING,
+                        FILE_ATTRIBUTE_NORMAL,
+                        nullptr);
+
+    RETURN_LAST_ERROR_IF(handle == INVALID_HANDLE_VALUE);
+
+    RETURN_LAST_ERROR_IF (!GetFileSizeEx(handle, &li));
+
+    if (li.HighPart != 0)
```
mikeharder: .NET Framework 4 limited the filesize to 1MB:

https://referencesource.microsoft.com/#System.Web/HttpRuntime.cs,264

I would have recommended matching the .NET Framework limit, however do we also need to now think about backwards compat with ANCM v1?  Since it allowed larger sizes, it would be a breaking change to reduce the size now.  I doubt many users would have an app_offline file larger than 1MB, but at the same time I don't know a good reason to limit the size to 1MB either.

src/AspNetCoreModuleV2/RequestHandlerLib/filewatcher.cpp
pakrym: If app_offline exists at the time we start file watcher it would not get reported an application would not be stopped. This is a fix for a race that `AppOfflineDroppedWhileSiteStarting_SiteShutsDown_InProcess` is trying to force.

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cpp
```diff
@@ -93,7 +95,14 @@ CreateApplication(
     try
     {
         RETURN_IF_FAILED(InitializeGlobalConfiguration(pServer, pHttpApplication));
-        
+
+        // In process application was already created
+        if (g_fInProcessApplicationCreated)
+        {
+            *ppApplication = new ShuttingDownApplication(*pServer, *pHttpApplication);
```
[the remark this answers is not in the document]
pakrym: >  Why does a second call to CreateApplication() indicate shutdown?

Because only one InProcess application can be created per w3wp process (CLR is not restartable). If `CreateApplication` being called the second time means that original InProcess application was recycled and server shutdown is happening. So we return the application that would serve "server shut down" messages.

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
mikeharder: The .NET Framework code is a bit strange, in that it seems to treat an app_offline file larger than 1MB as if it doesn't exist.  I could understand ignoring the file content if over 1MB (or the portion over 1MB), but leaving the app online if app_offline is too large seems like a bug rather than intentional behavior, so we wouldn't want to match it in ANCM.

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
pakrym: `m_strAppOfflineContent` is being copied when the handler is created so it won't be updated while the response is being written.

>  Can we include this test in the PR

Will add

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
```diff
@@ -0,0 +1,84 @@
...
+    {
+        // > 4gb file size not supported
+        // todo: log a warning at event log
+        return TRUE;
```
mikeharder: Any reason to return `TRUE` instead of `S_OK` here?  Is the goal to ignore the app_offline content if the file is > 4 GB, but still take the app offline?  If so, why limit at 4GB as opposed to another arbitrary size?

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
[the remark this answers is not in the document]
pakrym: It would be better for us not to read the content than to lock the file.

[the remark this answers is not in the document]
pakrym: Adding a test where `app_offline` file is locked would be nice.

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -413,107 +393,47 @@ APPLICATION_INFO::FindNativeAssemblyFromHostfxr(
...
-        if (m_pApplication != NULL)
-        {
-            pApplication = m_pApplication;
-
-            // Set m_pApplication to NULL first to prevent anyone from using it
-            m_pApplication = NULL;
-            pApplication->ShutDown();
-            pApplication->DereferenceApplication();
-        }
+        m_pApplication ->Stop(true); // fServerInitiated
```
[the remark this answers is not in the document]
pakrym: Yes

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -60,22 +48,16 @@ APPLICATION_INFO::Initialize(
...
+    
+    SRWExclusiveLock lock(m_applicationLock);
 
-    if (m_pApplication != nullptr && m_pApplication->QueryStatus() != RECYCLED)
-    {
-        return S_OK;
-    }
-
-    SRWExclusiveLock lock(m_srwLock);
+    auto& httpApplication = *pHttpContext->GetApplication();
```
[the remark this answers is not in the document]
pakrym: We use `httpApplication` past `Finished:` label, and `FINISHED()` would `goto` to it. We have to have `httpApplication` initialized before first `goto`

src/AspNetCoreModuleV2/AspNetCore/ServerErrorApplication.h
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+#pragma once
+#include "PollingAppOfflineApplication.h"
+#include "requesthandler.h"
+
+class ServerErrorHandler : public REQUEST_HANDLER
```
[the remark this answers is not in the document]
pakrym: Will do

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
pakrym: It should be S_OK. 

> Is the goal to ignore the app_offline content if the file is > 4 GB, but still take the app offline?  

Yes

>  If so, why limit at 4GB as opposed to another arbitrary size

This is the code we inherited from ANCM V1. I think it's just an arbitrary number they picked at that time.

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cpp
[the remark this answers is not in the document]
jkotalik: Let's add a comment here to avoid this confusion in the future as it is fairly confusing.

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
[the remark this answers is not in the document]
pakrym: >  However, this might be fine since publish should retry failed operations

Considering the number of problems we have even with having publish retry failed operations I would prefer to err on the side of caution.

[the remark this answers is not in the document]
mikeharder: Do we know if the .NET Framework technique of locking the file during read has caused real-world problems for customers?  If not, I hesitate to make the ANCM code more complicated to avoid locking the file.  If we do avoid locking, one invariant I think we must hold is that we either use the entire content of some version of the app_offline file, or no content.  We never use partial content.

For example, say app_offline v1 is 100 bytes, then it's overwritten by a larger file of 200 bytes between the calls to `CreateFile()` and `ReadFile()`.  I believe we might end up serving the first 100 bytes of app_offline v2, which IMO is unacceptable.  I will test this specific scenario in a standalone repro.

src/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cpp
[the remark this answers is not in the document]
mikeharder: OK, please add this in a comment.

src/AspNetCoreModuleV2/RequestHandlerLib/filewatcher.cpp
mikeharder: This may reduce the size of the race window, but isn't there still a race?  What if app_offline is added between the calls to `GetFileAttributes` and `ReadDirectoryChangesW`?

test/Common.FunctionalTests/AppOfflineTests.cs
pakrym: If request handler was not loaded it's fine to try again (this is shim failure scenario).
But if inprocess request handler was loaded but was not able to start an application we have to recycle the process because CLR is not restartable.

src/AspNetCoreModuleV2/CommonLib/AppOfflineApplication.cpp
mikeharder: I understand now, but I was confused in part by having two classes in the same file :).  Moving `AppOfflineHandler` and `AppOfflineApplication` into separate `.h` and `.cpp` files would make the classes easier to understand.

src/AspNetCoreModuleV2/RequestHandlerLib/filewatcher.cpp
pakrym: There is an inherent race when using file watcher, the file might have appeared between we called the method and OS initialized data structures required for while watching. This at least takes race window from hundreds of milliseconds to tens of milliseconds. 

Let me think if moving exists check after ReadDirectoryChangesW call would help.

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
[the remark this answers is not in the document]
mikeharder: OK, I am not used to reviewing code with `GOTO` :).

src/AspNetCoreModuleV2/RequestHandlerLib/filewatcher.cpp
mikeharder: I would expect there to be some way to check for the existence of a file without any races.  Worst case you could always fallback to polling (say once per second) in case everything else fails.  But it might also be sufficient to check for existence exactly once after the file watcher is created.

test/Common.FunctionalTests/AppOfflineTests.cs
mikeharder: I understand the difference now.  But would it maybe be better to recycle the process when shim fails as well?  Could this prevent customers from manually needing to restart their app, if the shim failed due to some transient error?

src/AspNetCoreModuleV2/CommonLib/PollingAppOfflineApplication.cpp
```diff
@@ -45,89 +36,16 @@ PollingAppOfflineApplication::AppOfflineExists()
             m_fAppOfflineFound = is_regular_file(m_appOfflineLocation);
```
mikeharder: In the code above, the comment says `1 second` but `c_appOfflineRefreshIntervalMS` is `200ms`.

test/Common.FunctionalTests/AppOfflineTests.cs
pakrym: >  Could this prevent customers from manually needing to restart their app, if the shim failed due to some transient error

I don't really understand how. We would retry when app_offline file would be cycled or configuration changed.

src/AspNetCoreModuleV2/RequestHandlerLib/filewatcher.cpp
```diff
@@ -358,27 +358,25 @@ FILE_WATCHER_ENTRY::Monitor(VOID)
...
-            &_overlapped,
-            NULL))
-        {
-            hr = HRESULT_FROM_WIN32(GetLastError());
-            DereferenceFileWatcherEntry();
-        }
+        PostQueuedCompletionStatus(_pFileMonitor->QueryCompletionPort(), 0, 0, &_overlapped);
     }
-
+
```
jkotalik: nit space.

test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -173,6 +158,44 @@ public class AppOfflineTests : IISFunctionalTestBase
...
+            });
+
+            for (int i = 0; i < 100; i++)
+            {
+                // AddAppOffline might fail if app_offline is being read by ANCM and deleted at the same time
+                RetryHelper.RetryOperation(
+                    () => AddAppOffline(deploymentResult.DeploymentResult.ContentRoot),
+                    e => Logger.LogError($"Failed to create app_offline : {e.Message}"),
+                    retryCount: 3,
+                    retryDelayMilliseconds: 100);
```
jkotalik: We may want to add variants that send a request in between adding and removing app offline.

pakrym: Wouldn't it be AppOfflineDroppedWhileSiteRunning_SiteShutsDown_InProcess and AppOfflineDroppedWhileSiteRunning_SiteShutsDown_OutOfProcess?

src/AspNetCoreModuleV2/CommonLib/PollingAppOfflineApplication.h
```diff
@@ -4,54 +4,40 @@
 #pragma once
 #include <experimental/filesystem>
 #include "application.h"
-#include "requesthandler.h"
+
+enum PollingAppOfflineApplicationMode
```
jkotalik: Consider just having this be a bool.

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -118,22 +117,17 @@ IN_PROCESS_APPLICATION::ShutDown(
...
-
-        //
-        // Managed layer may block the shutdown and lead to shutdown timeout
-        // Assumption: only one inprocess application is hosted.
-        // Call process exit to force shutdown
-        //
-        exit(hr);
     }
-    else
+	else
```
jkotalik: nit

src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cpp
```diff
@@ -469,6 +472,8 @@ EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
 VOID
 set_main_handler(_In_ hostfxr_main_fn main)
 {
+    // Allow inprocess application to be recreated as we reuse the same CLR
+    g_fInProcessApplicationCreated = false;
```
jkotalik: We will need to be careful here with the test server. Technically this isn't the right place to add it, as the next request will create a shutdown application

pakrym: > as the next request will create a shutdown application?

Why? We don't shutdown test server in the middle of the test.

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -413,107 +392,47 @@ APPLICATION_INFO::FindNativeAssemblyFromHostfxr(
...
+
+    return 0;
 }
 
 
 VOID
 APPLICATION_INFO::ShutDownApplication()
 {
-    IAPPLICATION* pApplication = NULL;
+    SRWExclusiveLock lock(m_applicationLock);
```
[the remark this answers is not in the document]
pakrym: Why would it delay shutdown? When the first thread running `ShutDownApplication` unlocks everyone else would just fall through.

src/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cpp
jkotalik: I see now. This makes sense.

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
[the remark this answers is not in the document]
jkotalik: You are correct, it wouldn't delay shutdown.

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -413,107 +392,47 @@ APPLICATION_INFO::FindNativeAssemblyFromHostfxr(
 VOID
 APPLICATION_INFO::RecycleApplication()
 {
-    IAPPLICATION* pApplication;
-    HANDLE       hThread = INVALID_HANDLE_VALUE;
+    SRWExclusiveLock lock(m_applicationLock);
```
davidfowl: Nice to have this logic cleaned up

pakrym:
@dotnet-bot test this please

pakrym:
@dotnet-bot test this please

Issue #795, opened 2018-04-24
V2: Inprocess: Diagnostic: Wrong event when an invalid path is used


jhkimnew:
STEPS TO REPRO:
1. Use V2 inprocess mode
2. Set a bogus process path to verify that we shows the right error event log
3.  Send a request and check the new ANCM event log generated after that

EXPECTED:

The event log should say that the process path was wrong or something similar information when we got in the out-of-process mode.

NOTE:

Below event is shown, which is wrong information and makes users confused.

<><><>
Could not find the aspnetcorerh.dll for in-process application. Please confirm the Microsoft.AspNetCore.Server.IIS package is referenced in your application. 
<><><>

Issue #552, opened 2018-01-24
ANCM: There should be better error handling for hostingModel value


jhkimnew:
Currently the hostingModel value is a string value. So, users can do set a wrong value by mistake.
If the wrong value is configured, we show 500 error. 
I checked Freb log but there was no hint on the root cause.

Here is the example scenario.

1. Set "bogusvalue" instead of "inprocess" for the hostingModel
2. Send a request to the application

EXPECTED:

Users should get some specific error in either Freb tracing log or Windows event error.
Or, we should use enumeration value for the hostingModel value to avoid this situation.

ACTUAL:

500 error with general configuration failure like this.

```
AspNetCoreModule
EXECUTE_REQUEST_HANDLER
500
Internal Server Error
0
The data is invalid.  (0x8007000d)
```

PR #1081, opened 2018-07-20
Add event logs for some startup failure cases

jkotalik:
For #795 and #552


src/AspNetCoreModuleV2/AspNetCore/aspnetcore_shim_config.cpp
```diff
@@ -13,7 +14,7 @@ ASPNETCORE_SHIM_CONFIG::Populate(
     IHttpApplication *pHttpApplication
 )
 {
-    STACK_STRU(strHostingModel, 12);
+    STRU                            strHostingModel;
```
jkotalik: If someone puts an arbitrary value for the hosting model, we don't want to worry about buffer overflows.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISExpressDeployer.cs
```diff
@@ -280,8 +280,9 @@ private void PrepareConfig(string contentRoot, string dllRoot, int port)
             if (DeploymentParameters.PublishApplicationBeforeDeployment)
             {
                 // For published apps, prefer the content in the web.config, but update it.
-                IISDeploymentParameters.ModifyAspNetCoreSectionInWebConfig(key: "hostingModel",
+                IISDeploymentParameters.PrependAspNetCoreSectionInWebConfig(key: "hostingModel",
```
jkotalik: Need to prepend this as tests may overwrite the hostingModel section (like I do below).

src/AspNetCoreModuleV2/AspNetCore/aspnetcore_shim_config.cpp
pakrym: STACK_STRU would reallocate AFAIK

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -368,6 +371,8 @@ HOSTFXR_UTILITY::GetAbsolutePathToDotnet(
     // Only do it if no path is specified
     if (requestedPath.has_parent_path())
     {
+        WLOG_INFOF(L"Absolute path to dotnet.exe was not found at %s", requestedPath.c_str());
```
pakrym: It's not "not found", it just wasn't an absolute path.

pakrym: So what would we have in event log in this case?

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -141,14 +141,17 @@ HOSTFXR_UTILITY::GetHostFxrParameters(
...
+            EVENTLOG(hEventLog, INVALID_PROCESS_PATH, processPath.c_str(), hr);
+            return hr;
         }
 
         processPath = fullProcessPath.value();
 
         auto hostFxrPath = GetAbsolutePathToHostFxr(processPath, hEventLog);
         if (!hostFxrPath.has_value())
         {
+            hr = HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND);
```
pakrym: You are setting hr and returning E_FAIL

test/IISExpress.FunctionalTests/InProcess/StartupTests.cs
```diff
@@ -144,6 +146,21 @@ public static TestMatrix TestVariants
...
+            var deploymentParameters = GetBaseDeploymentParameters();
+            deploymentParameters.ModifyAspNetCoreSectionInWebConfig("hostingModel", "bogus");
+
+            var deploymentResult = await DeployAsync(deploymentParameters);
+
+            var response = await deploymentResult.RetryingHttpClient.GetAsync("HelloWorld");
+
+            Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
+
+            EventLogHelpers.VerifyEventLogEvent(TestSink, "Unknown hosting model bogus. Please specify either hostingModel=\"inprocess\" or hostingModel=\"outofprocess\" in the web.config file.");
```
pakrym: add quotes aroung bogus

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
jkotalik: L"Invalid or unknown processPath provided in web.config: processPath = '%s', ErrorCode = '0x%x'."

Issue #1069, opened 2018-07-17
Resolve win7 test failures with 'api-ms-win-service-core'


pakrym:

jkotalik:
Nah some are still failing: http://aspnetci/viewLog.html?buildId=510928&tab=buildResultsDiv&buildTypeId=XPlat_Windows_Win2008r2_Universe
`System.DllNotFoundException : Unable to load DLL 'api-ms-win-service-core-l1-1-1.dll' or one of its dependencies: The specified module could not be found. (Exception from HRESULT: 0x8007007E)`

jkotalik:
Will keep open for monitoring purposes.

pakrym:
Fixed in #1197.

PR #1096, opened 2018-07-25
Skip win7 tests

jkotalik:
Issue: https://github.com/aspnet/IISIntegration/issues/1069
This will cover a majority of the failures seen on CI.


pakrym:
Ugh, I don't know about this. I don't want us to skip them forever

jkotalik:
I don't either, but we have 3 issues filed against us all revolving around this issue. I'd like to at least stop the fire before we try to figure out why this is happening.

PR #1109, opened 2018-07-27
Fix win7 test failures.

jkotalik:
For #1069. Apparently by retrying after receiving the `api.dll` error, it works 😅 .

The applicationInitialization doesn't exist on some win7 machines either, so removed that from the default config.


pakrym:
Don't retry everything. Do DllNotFoundException and ComException with HRESULT 0x800710D8

PR #1115, opened 2018-07-29
[WIP] Skip VSTS attribute

pakrym:


.vsts-pipelines/templates/build-steps.yml
```diff
@@ -3,6 +3,8 @@ phases:
   parameters:
     agentOs: Windows
     beforeBuild:
+      - powershell: "git submodule update --init"
```
pakrym: @natemcmaster is there a better way?

natemcmaster: Change the checkout settings to include submodules.

natemcmaster:
Do we know why these tests fail in VSTS? Do we have a follow-up item to ensure these get fixed? We plan to drop everything except VSTS for CI.

pakrym:
@natemcmaster https://github.com/aspnet/IISIntegration/issues/1111

.vsts-pipelines/templates/build-steps.yml
pakrym: Where?

natemcmaster: In the VSTS build definition. Unfortunately this is one you can't set via YAML yet. I've updated it for you

https://dotnet.visualstudio.com/public/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=27

image

Issue #1108, opened 2018-07-27
Flaky test: AppOfflineAddedAndRemovedStress


jkotalik:
@pakrym #1107 should resolve this? https://ci.appveyor.com/project/aspnetci/iisintegration/build/1.0.1691

pakrym:
Yes

jkotalik:
AppOfflineDroppedWhileSiteFailedToStartInShim_AppOfflineServed also failed. https://dotnet.visualstudio.com/public/_build/results?buildId=9059&view=logs

ryanbrandenburg:
AppOfflineAddedAndRemovedStress [failed again](http://aspnetci/viewLog.html?buildId=519145&buildTypeId=Releases_22xPublic_Win2012).

This comment was made automatically. If there is a problem contact @ryanbrandenburg.

ryanbrandenburg:
AppOfflineAddedAndRemovedStress [failed with about the same error](http://aspnetci/viewLog.html?buildId=518725&buildTypeId=XPlat_Windows_Win8_Universe).

This comment was made automatically. If there is a problem contact @ryanbrandenburg.

ryanbrandenburg:
AppOfflineAddedAndRemovedStress [failed with about the same error](http://aspnetci/viewLog.html?buildId=521299&buildTypeId=XPlat_Windows_Win2008r2_Universe).

This comment was made automatically. If there is a problem contact @ryanbrandenburg.

mikeharder:
Another repro in VSTS: https://dotnet.visualstudio.com/public/_build/results?buildId=10523&view=logs

mikeharder:
@pakrym, @jkotalik: Appears test is still flaky after #1107.

pakrym:
> Another repro in VSTS: https://dotnet.visualstudio.com/public/_build/results?buildId=10523&view=logs

Fix for it is in https://github.com/aspnet/IISIntegration/pull/1148

mikeharder:
Another repro: https://ci3.dot.net/job/aspnet_IISIntegration/job/release_2.2/job/windows-Configuration_Release_prtest/363/console

PR #1120, opened 2018-07-30
Fix ServerShutsDownWhenMainExits test

pakrym:
https://github.com/aspnet/IISIntegration/issues/995

We are not tracking connections on the managed side and are not waiting for them to complete before shutting down manged `IServer`. 

What happens when this test fails is that we stop CLR entirely before the request has a chance to finish triggering HttpClient retries that are not able to connect to anything.

I changed `RetryingHttpClient` to just `HttpClient` so we don't retry anymore. ~What might also happen is an exception being thrown from `HttpClient` when the server cuts off connection but I'd like to see it first before adding code to ignore.~

I did add exception supression.


pakrym:
VSTS failure: https://github.com/aspnet/IISIntegration/issues/1108

PR #1135, opened 2018-08-01
Reduce msbuild invocations while building native

pakrym:
Build solution in parallel instead of building projects one by one.


pakrym:
Takes >1m off VSTS build.

Issue #1124, opened 2018-07-31
Port collisions when multiple apps in separate app pools are started at the same time


mikeharder:
When multiple apps in separate app pools are started at the same time, it's possible for some apps to fail startup due to port collisions with other apps.  I have confirmed the root cause is:

https://github.com/aspnet/IISIntegration/blob/e1f1705d11d97560c574dbe68a0b560441470f58/src/AspNetCoreModuleV1/AspNetCore/src/serverprocess.cxx#L1915

The resolution of `GetTickCount()` is 10-16ms, so if two apps (in two app pools) are started within this time window, it's possible that `GetTickCount()` returns the same value for both apps, which causes the port collision.

I have verified both v1 and v2 out-of-process are impacted.

# Repro Steps
1. Configure IIS with 10 apps, each in their own app pool.
2. Make requests to all 10 apps in parallel, as quickly as possible.
3. Some apps may fail to start, since they try to start on the same port (as shown in event logs)

# v1 out-of-process
```
Application 'MACHINE/WEBROOT/APPHOST/DEFAULT WEB SITE/2.1/OUTPROC-MULTI-APPPOOL/4'
started process '2360' successfully and is listening on port '6074'.

Application 'MACHINE/WEBROOT/APPHOST/DEFAULT WEB SITE/2.1/OUTPROC-MULTI-APPPOOL/2'
with physical root 'C:\apps\2.1\outproc-multi-apppool\2\'
created process with commandline 'dotnet .\app.dll' but failed to listen on the given port '6074'
```

# v2 out-of-process
```
Application '/LM/W3SVC/1/ROOT/2.2/outproc-multi-apppool/10'
started process '376' successfully and process '376' is listening on port '30362'.

Application '/LM/W3SVC/1/ROOT/2.2/outproc-multi-apppool/5'
with physical root 'C:\share\2.2\outproc-multi-apppool\5\' 
created process with commandline 'dotnet .\app.dll' but failed to listen on the given port '30362'

Application '/LM/W3SVC/1/ROOT/2.2/outproc-multi-apppool/5'
with physical root 'C:\share\2.2\outproc-multi-apppool\5\' 
failed to start process with commandline 'dotnet .\app.dll'
at stage 'PostStartCheck', ErrorCode = '0x8007065f', assigned port 30362, retryCounter '0'.

Application '/LM/W3SVC/1/ROOT/2.2/outproc-multi-apppool/5'
[...]

PR #1136, opened 2018-08-01
Reduce probability of startup port collisions

mikeharder:
- `GetTickCount()` is limited to the resolution of the system timer, which is typically 10-16ms.  If two apps in separate app pools are started within this time window, it's possible `GetTickCount()` will return the same value, which causes the apps to try the same random port(s).
- Addresses #1124


pakrym:
Did you have a repro? Were you able to verify that conflict happens less often?

davidfowl:
Don't we want to bind(0)? And pass that port to kestrel?

mikeharder:
@pakrym: I can repro in both v1 and v2 outproc.  See https://github.com/aspnet/IISIntegration/issues/1124 for repro steps and log messages.  You can also repro in containers:

* https://github.com/mikeharder/ancm-docker/tree/master/app/2.1/outproc-multi-apppool
* https://github.com/mikeharder/ancm-docker/tree/master/app/2.2/outproc-multi-apppool.

In v1 outproc, if two apps select the same port, it appears the second app fails to start and never retries.

In v2 outproc, if two apps initially select the same port, it appears the second app usually retries and then does start.  However, one time I was able to repro with v2 outproc where one of the apps failed even after retry.

After this change, I was never able to repro a port collision with 10 apps in separate app pools.

jkotalik:
I hesitate not having some sort of automated test. 

> Don't we want to bind(0)? And pass that port to kestrel?

I'd don't see how binding to 0, unbinding the port, and then passing it to kestrel is better than this? Wouldn't there be issues with kestrel thinking the port is in use? 
If we want anyone to call bind(0), its kestrel. 

@mikeharder I think there are still issues we need to address with retrying to start kestrel. There is a new loop in ANCM V2 that retries starting the process: https://github.com/aspnet/IISIntegration/blob/release/2.2/src/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp#L768. I think we need something equivalent in V1.

davidfowl:
> I'd don't see how binding to 0, unbinding the port, and then passing it to kestrel is better than this? Wouldn't there be issues with kestrel thinking the port is in use?
If we want anyone to call bind(0), its kestrel.

The OS selects the "random" port in the relevant port range instead of us having to rely on the hard coded random port range specified here.

mikeharder:
@davidfowl, @jkotalik: In my estimation, "bind to 0 and pass to kestrel" and "random" have approximately the same probability of collision (say 0.001%).  Very low, but high enough to require retry logic in either case.  So I don't think the added complexity of "bind to 0 and pass to kestrel" is worth it.  I would prefer to keep our changes to outproc (especially v1) as small as possible.

mikeharder:
@jkotalik: In theory, we could add an automated test which does the same thing as my manual repro:

* Add 10 apps in separate app pools
* Make a request to each app in parallel (so they start as close in time as possible)
* Verify all responses are HTTP 200

With the previous code, this test would usually fail on my dev machine (at least for ANCM v1).  It should never fail with the new code.

Do you think such a test would be reliable enough and worth adding?

pakrym:
@mikeharder I think it would be beneficial to have it both for this fix and as a startup stress scenario in general.

jkotalik:
> Do you think such a test would be reliable enough and worth adding?
Yeah, I think so.

mikeharder:
> > I'd don't see how binding to 0, unbinding the port, and then passing it to kestrel is better than this? Wouldn't there be issues with kestrel thinking the port is in use?
If we want anyone to call bind(0), its kestrel.

> The OS selects the "random" port in the relevant port range instead of us having to rely on the hard coded random port range specified here.

This would also be somewhat of a breaking change for ANCM v1 (different range used to select random ports).

mikeharder:
> I think there are still issues we need to address with retrying to start kestrel. There is a new loop in ANCM V2 that retries starting the process: https://github.com/aspnet/IISIntegration/blob/release/2.2/src/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp#L768. I think we need something equivalent in V1.

I think this should be treated as a separate issue: https://github.com/aspnet/IISIntegration/issues/1137

jkotalik:
As long as it is addressed 😄

mikeharder:
@jkotalik, @pakrym: I will work on adding that test.  If we want to test this scenario thoroughly (including retries), we would need a way to use a fixed random seed for testing to force port collisions (rather than `random_device`).  In managed code this would typically be done via dependency injection (constructor or otherwise).  Is there a similar pattern for native code?  Have we done anything like this in existing tests?

jkotalik:
>  dependency injection 

Nope 😢 

I don't think it's required to test the negative scenario of forced port collisions.

pakrym:
@mikeharder I would not be too worried about retries. Let's just add a test that starts 10 sites all at once and makes sure they all start. If this scenario failed before it would be a good place to start.

mikeharder:
> I don't think it's required to test the negative scenario of forced port collisions.

I agree it's not required for this PR.  But I do think we'll need a way to force collisions if we want to verify retries work correctly.

test/Common.FunctionalTests/Utilities/Helpers.cs
```diff
@@ -25,7 +25,23 @@ public static string GetTestWebSitePath(string name)
         public static string GetInProcessTestSitesPath() => GetTestWebSitePath("InProcessWebSite");
 
         public static string GetOutOfProcessTestSitesPath() => GetTestWebSitePath("OutOfProcessWebSite");
-  
+
+        public static IISDeploymentParameters GetBaseDeploymentParameters(string site = null, HostingModel hostingModel = HostingModel.InProcess, bool publish = false)
```
pakrym: Why is this back?

test/IISExpress.FunctionalTests/OutOfProcess/MutlipleAppTests.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.AspNetCore.Testing.xunit;
+using Microsoft.Extensions.Logging.Abstractions;
+using Xunit;
+
+namespace IISExpress.FunctionalTests.OutOfProcess
+{
+    public class MutlipleAppTests
+    {
+        [ConditionalFact]
+        public async Task Test()
```
pakrym: Is there a reason this test is IISExpress specific?

test/IISExpress.FunctionalTests/OutOfProcess/MutlipleAppTests.cs
```diff
@@ -0,0 +1,50 @@
...
+                var deployer = new IISExpressDeployer(deploymentParameters, new NullLoggerFactory());
+                deployers.Add((deployer));
+                deploymentResults[i] = await deployer.DeployAsync();
+            }
+
+            var taskList = new Task<HttpResponseMessage>[numSites];
+
+            for (var i = 0; i < numSites; i++)
+            {
+                taskList[i] = deploymentResults[i].HttpClient.GetAsync("/HelloWorld");
```
pakrym: Can we use the stress test helper in some way?

test/IISExpress.FunctionalTests/OutOfProcess/MutlipleAppTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            const int numSites = 10;
+            var deployers = new List<IISExpressDeployer>();
+            var deploymentResults = new DeploymentResult[numSites];
+            for (var i = 0; i < numSites; i++)
+            {
+                var deploymentParameters = Helpers.GetBaseDeploymentParameters(site: "OutOfProcessWebsite");
+                deploymentParameters.AncmVersion = AncmVersion.AspNetCoreModule;
+
+                var deployer = new IISExpressDeployer(deploymentParameters, new NullLoggerFactory());
```
pakrym: Why null logger factory?

test/IISExpress.FunctionalTests/OutOfProcess/MutlipleAppTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        [ConditionalFact]
+        public async Task Test()
+        {
+            const int numSites = 10;
+            var deployers = new List<IISExpressDeployer>();
+            var deploymentResults = new DeploymentResult[numSites];
+            for (var i = 0; i < numSites; i++)
+            {
+                var deploymentParameters = Helpers.GetBaseDeploymentParameters(site: "OutOfProcessWebsite");
+                deploymentParameters.AncmVersion = AncmVersion.AspNetCoreModule;
```
pakrym: We need both V1 and V2

pakrym:
Would be nice to have a test that starts 10 web applications inside a single iisexpress/w3wp instance.

mikeharder:
@pakrym: The test is still a WIP.  I will be adding coverage for both v1 and v2, IIS and IIS Express.  It's a good idea to also test 10 applications in a single app pool for completeness, though this configuration should not be impacted by the random seed issue, since apps within an app pool are started sequentially rather than in parallel.

jkotalik:
Sorry, I should have commented on this thread that @mikeharder and I worked on a general outline for what tests will look like.

mikeharder:
@jkotalik, @pakrym: Ready for next round of reviews

test/Common.Tests/Utilities/DisposableEnumerable.cs
```diff
@@ -0,0 +1,25 @@
...
+            _enumerable = enumerable;
+        }
+
+        public void Dispose()
+        {
+            foreach (var item in _enumerable)
+            {
+                if (item != null)
+                {
+                    item.Dispose();
```
davidfowl: nit: item?.Dispose()

test/Common.Tests/Utilities/DisposableEnumerable.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Server.IntegrationTesting
+{
+    public class DisposableEnumerable : IDisposable
```
pakrym: I would've done disposable list but it doesn't matter in context of this PR.

test/Common.Tests/Utilities/DisposableEnumerable.cs
```diff
@@ -0,0 +1,25 @@
+using System;
```
pakrym: (C)

test/IISExpress.FunctionalTests/OutOfProcess/MultipleAppTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.IIS.FunctionalTests.Utilities;
+using Microsoft.AspNetCore.Server.IntegrationTesting;
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
+{
+    [Collection(PublishedSitesCollection.Name)]
+    public class MultipleAppTests : IISFunctionalTestBase
```
jkotalik: If this test does fail on CI, #1137 will probably be the fix.

PR #1129, opened 2018-07-31
Ship shim with integration testing 

pakrym:
And cleanup test projects.

Issue #1049, opened 2018-07-12
IISExpress tests sometimes fail to run all tests (process exit).


jkotalik:
```
Starting test execution, please wait...
  [xUnit.net 00:02:07.46]     Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.HttpsTest.HttpsHelloWorld_ClientCert [SKIP]
  Skipped  Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.HttpsTest.HttpsHelloWorld_ClientCert
  The active test run was aborted. Reason: 
  Results File: C:\projects\iisintegration\artifacts\logs\UnitTests-netcoreapp2.2-t000.trx
  
  Total tests: Unknown. Passed: 118. Failed: 0. Skipped: 1.
  Test Run Aborted.
  Test execution time: 2.3104 Minutes
C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17099\modules\vstest\module.targets(108,5): error : Test group UnitTests/netcoreapp2.2 failed with exit code '1'. [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.2.0-preview1-17099\KoreBuild.proj]
```

pakrym:
I started seeing this on VSTS lately.

mikeharder:
@jkotalik, @pakrym: Please add a link to a build log if you see this fail again.  I'm not sure if it will help, but currently there's not much information in this issue.

pakrym:
Not much info on this issue can be gained from build log, it always looks like this:

```
  
  Starting test execution, please wait...
  [xUnit.net 00:00:00.48]     Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReaderThrowsCancelledException [SKIP]
  [xUnit.net 00:00:00.50]     Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.WriterThrowsCancelledException [SKIP]
  [xUnit.net 00:00:00.50]     Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.WritesSucceedAfterClientDisconnect [SKIP]
  [xUnit.net 00:00:00.50]     Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReadThrowsAfterClientDisconnect [SKIP]
[...]

PR #1145, opened 2018-08-02
Fix incrementality checks for test projects

pakrym:
When I did https://github.com/aspnet/IISIntegration/pull/1129 I've broken a scenario where native asset is being rebuilt while managed (test project/test site is not).

So I've added explicit project references from csprojes to vcxprojes that would force a rebuild. Unfortunately I had to set `DisableFastUpToDateCheck` to make them work with `<ReferenceOutputAssembly>False</ReferenceOutputAssembly>`

Additionally, I switched to always copying request handler bin local without platform subdirectory. It allows us to copy the handler only once and it would work for both build and publish scenario.


pakrym:
VSTS failure: #1049

PR #1177, opened 2018-08-08
Split handler resolution from application info

pakrym:
Move handler resolution logic and error handling into its own class and add tests for multiple app scenarios.

### DIFF
--- a/src/AspNetCoreModuleV2/AspNetCore/dllmain.cpp
+++ b/src/AspNetCoreModuleV2/AspNetCore/dllmain.cpp
@@ -16,6 +16,7 @@ DECLARE_DEBUG_PRINT_OBJECT("aspnetcorev2.dll");
 HANDLE              g_hEventLog = NULL;
 BOOL                g_fRecycleProcessCalled = FALSE;
 BOOL                g_fInShutdown = FALSE;
+HINSTANCE           g_hServerModule;
 
 VOID
 StaticCleanup()
@@ -40,6 +41,7 @@ BOOL WINAPI DllMain(HMODULE hModule,
     switch (ul_reason_for_call)
     {
     case DLL_PROCESS_ATTACH:
+        g_hServerModule = hModule;
         DisableThreadLibraryCalls(hModule);
         DebugInitialize(hModule);
         break;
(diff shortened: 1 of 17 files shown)


pakrym:
VSTS Failure: https://github.com/aspnet/IISIntegration/issues/1049

pakrym:
PR that addresses Jenkins failure: https://github.com/aspnet/IISIntegration/pull/1179

jkotalik:
Any functional changes?

pakrym:
Validation logic moved but rules didn't change.

src/AspNetCoreModuleV2/AspNetCore/dllmain.cpp
```diff
@@ -40,6 +41,7 @@ BOOL WINAPI DllMain(HMODULE hModule,
     switch (ul_reason_for_call)
     {
     case DLL_PROCESS_ATTACH:
+        g_hServerModule = hModule;
```
jkotalik: Quite a long way to propagate the HINSTANCE. Any other way to get the module?

test/Common.FunctionalTests/MultiApplicationTests.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        [ConditionalFact]
+        public async Task RunsTwoOutOfProcessApps()
+        {
+            var parameters = _fixture.GetBaseDeploymentParameters(HostingModel.OutOfProcess, publish: true);
+            parameters.ServerConfigActionList.Add(DuplicateApplication);
+            var result = await DeployAsync(parameters);
+            var id1 = await result.HttpClient.GetStringAsync("/app1/ProcessId");
+            var id2 = await result.HttpClient.GetStringAsync("/app2/ProcessId");
+            Assert.NotEqual(id2, id1);
```
jkotalik: Assert both of these are 200.

pakrym: `GetStringAsync` would throw otherwise.

src/AspNetCoreModuleV2/AspNetCore/dllmain.cpp
pakrym: This is the most reliable way to do it.

PR #1187, opened 2018-08-09
Fix environmentVariable usage in tests

pakrym:
Verified on our CI


pakrym:
Merging to unblock things.

PR #1197, opened 2018-08-09
Start W3SVC if it is stopped

pakrym:


pakrym:
I think I've figured it out and it might be a fix for all our `api-ms-win-service-core-l1-1-1.dll` related issues: `System.ServiceProcess.ServiceController` that `Web.Administration` references was old 4.1 instead of the latest 4.5.

PR #1212, opened 2018-08-11
Start adding configuration reload tests

pakrym:


test/Common.FunctionalTests/ConfigurationChangeTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public ConfigurationChangeTests(PublishedSitesFixture fixture)
+        {
+            _fixture = fixture;
+        }
+
+        [ConditionalFact]
+        [RequiresIIS(IISCapability.ShutdownToken)]
+        public async Task ConfigurationChangeStopsInProcess()
+        {
```
jkotalik: We should verify that in full IIS, a new worker process is created automatically here.

pakrym: Done, even enabled app_offline tests

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployer.cs
```diff
@@ -339,6 +339,12 @@ private void Stop()
                         throw new InvalidOperationException("Site not stopped yet");
                     }
 
+                    if (appPool.WorkerProcesses.Any(wp => wp.State == WorkerProcessState.Running ||
```
pakrym: This is because we sometimes force a recycle so HostProcess is not always the running worker process.

test/Common.FunctionalTests/Utilities/Helpers.cs
```diff
@@ -73,5 +74,64 @@ public static void CopyFiles(DirectoryInfo source, DirectoryInfo target, ILogger
...
+        }
+
+        public static void AssertWorkerProcessStop(this IISDeploymentResult deploymentResult)
+        {
+            var hostProcess = deploymentResult.HostProcess;
+            Assert.True(hostProcess.WaitForExit((int)TimeoutExtensions.DefaultTimeout.TotalMilliseconds));
+
+            if (deploymentResult.DeploymentParameters.ServerType == ServerType.IISExpress)
+            {
+                Assert.Equal(0, hostProcess.ExitCode);
```
jkotalik: Why can't this check be done for w3wp?

test/Common.FunctionalTests/ConfigurationChangeTests.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private readonly PublishedSitesFixture _fixture;
+
+        public ConfigurationChangeTests(PublishedSitesFixture fixture)
+        {
+            _fixture = fixture;
+        }
+
+        [ConditionalFact]
+        public async Task ConfigurationChangeStopsInProcess()
```
jkotalik: Add a test that modifies apphost.config?

test/Common.FunctionalTests/Utilities/Helpers.cs
pakrym: When you don't start process you can't observe it's exit code.

test/Common.FunctionalTests/ConfigurationChangeTests.cs
pakrym: Feels like this tests IIS not ANCM.

jkotalik: ANCM should reload on apphost.config changes too. Difference is global vs local configurations.

pakrym: https://github.com/aspnet/IISIntegration/issues/1222

PR #1214, opened 2018-08-12
Enable any app (including Docker) to run under ANCM with access to the IIS user identity

nil4:
ANCM is architected as a generic reverse proxy (owing to its HttpPlatformHandler origins), but a couple of implementation choices limit its full potential to ASP.NET Core apps. 

This PR enables a few new options (all opt-in) that enable a wider range of apps to use IIS and ANCM (v2 out-of-process) **with** access the IIS authenticated user identity. 

A more detailed write-up is available at https://github.com/nil4/ANCM-AnyApp/blob/master/ancmv2_extra/Readme.md. 

The [same repository](https://github.com/nil4/ANCM-AnyApp) includes a few demo apps that use the features in this PR; the [repo Readme](https://github.com/nil4/ANCM-AnyApp#readme) file contains a walk-through of the demo apps.

#### Background

When IIS is configured to use Windows authentication, ANCM passes the identity of the user to the backend application, as a [forwarded request header](https://github.com/aspnet/IISIntegration/blob/8cf39d35924ad875b2af32d8caaddcdfc4b6693b/src/AspNetCoreModuleV2/OutOfProcessRequestHandler/forwardinghandler.cpp#L925) (`MS-ASPNETCORE-WINAUTHTOKEN`). The header value is a hex number (e.g. `39E`), representing the Windows handle for the user identity. 

The backend app reads the user token/handle value on every request, and [creates a `WindowsIdentity`](https://github.com/aspnet/IISIntegration/blob/8cf39d35924ad875b2af32d8caaddcdfc4b6693b/src/Microsoft.AspNetCore.Server.IISIntegration/AuthenticationHandler.cs#L41) from the handle value, which is used for the duration of the request.

This method of passing the IIS authentication information has a few downsides:

-  it only works for handles of [Windows-authentication users](https://github.com/aspnet/IISIntegration/blob/8cf39d35924ad875b2af32d8caaddcdfc4b6693b/src/AspNetCoreModuleV2/OutOfProcessRequestHandler/forwardinghandler.cpp#L926), while IIS supports multiple authentication schemes (see the [*Remarks* section](https://msdn.microsoft.com/en-us/library/ms689371(v=vs.90).aspx));

-   it does not work for applications that cannot (or are not willing to) call Win32 APIs to recreate a Windows identity from the handle value; examples include apps running in Docker containers, or cross-platform applications that *run* on Windows, but are designed to also run on other web servers, e.g. expecting a `REMOTE_USER` header (like many NodeJS, Java, Ruby or Python apps do.)

#### PR description

The first commit adds new ANCM settings to allow passing the IIS authenticated user **name** to the backend application. These settings are complementary to the existing `forwardWindowsAuthToken` setting, which is enabled by default today, but applicable to only ASP.NET Core apps.

The **`forwardUserName`** setting (default value: `false`), when enabled, instructs ANCM to pass the IIS authenticated user **name** in a request header to the backend app, regardless of the IIS authentication method. If IIS anonymous authentication is used, an empty value will be forwarded. When this option is enabled, `forwardWindowsAuthToken` should be disabled, as the backend app cannot be expected to close the Windows user handle; this is not currently enforced. 

The **`forwardDomainName`** setting (default value: `true`) specifies whether the domain name (if any) of the authenticated IIS user is forwarded to the backend app. When set to `true`, the forwarded user name is `MyDomain\MyUserName`; when `false`, only `MyUserName` will be sent. 

The **`forwardUserNameHeader`** setting instructs ANCM which header to use when forwarding the IIS user name to the backend app. The default value, `MS-ASPNETCORE-USER`, is used for consistency with other forwarded headers, and to ensure the value cannot be tampered by a malicious user (see [security considerations](https://github.com/nil4/ANCM-AnyApp/blob/master/ancmv2_extra/Readme.md#security)). It can be set to another value (e.g. `REMOTE_USER`) if the implications are well-understood.

The second commit adds one setting (**`disableProcessIdCheck`**, default value: `false`) specifically for backend applications running in Docker containers. 

ANCM v2 currently expects the process it launches (or one of *its* child processes) to own the listening port configured by the `ASPNETCORE_PORT` environment variable. That does not match what Docker does: the user-mode `docker.exe` program *triggers the container launch*, but the Docker system/daemon service (`com.docker.service`) actually hosts the running container and the ports it listens to. 

Since there is no parent-child relationship between the two Docker processes in this case, ANCM times-out trying to match the port to the `docker.exe` process it launched. The `disableProcessIdCheck` setting makes the ANCM test less strict: it still verifies that *some* process listens on the expected port, but it no longer requires that process to be directly related to the one it starts.

The third commit re-enables replacing `%ASPNETCORE_PORT%` placeholders in the ANCM process arguments. This feature used to be available in earlier ANCM versions, but [regressed during a refactoring](https://github.com/aspnet/AspNetCoreModule/issues/117#issuecomment-311748366).

This restores feature parity with HttpPlatformHandler, which is documented to [performs a similar replacement](https://technet.microsoft.com/en-us/library/mt125371.aspx):

> %HTTP_PLATFORM_PORT% is a special placeholder that must be specified as part of the arguments attribute or the httpPlatformEnvironmentVariables list.
> HttpPlatformHandler dynamically replaces this variable with a port so that the HTTP listener can listen on this port.

#### Alternatives

IIS [Application Request Routing](https://www.iis.net/downloads/microsoft/application-request-routing) and [HttpPlatformHandler](https://www.iis.net/downloads/microsoft/httpplatformhandler) enable a variety of web applications to be hosted under IIS on Windows. 

However there is still no native IIS solution for providing the proxied application with the *identity of the IIS user*, despite clear user demand ([1](https://forums.iis.net/t/1163866.aspx), [2](https://forums.iis.net/t/1213161.aspx), [3](https://forums.iis.net/t/1155169.aspx), [4](https://stackoverflow.com/questions/37074887/issues-getting-remote-user-passed-from-shibboleth-when-using-iis), [5](https://stackoverflow.com/questions/24032014/forwarding-ntlm-credentials-from-iis-with-arr-and-url-rewrite), [6](https://stackoverflow.com/questions/33347412/rails-app-remote-user-attribute-in-iis-8-5-with-windows-authentication), [7](https://stackoverflow.com/questions/7364736/http-server-behind-iis-pass-authentication-headers)). 

HttpPlaformHandler, in particular, would arguably be a better fit for the options added in this PR. But it is not open source; and while support for the `forwardWindowsAuthToken` setting was added in v1.2 for DNX, requests to [enable `REMOTE_USER` header](https://forums.iis.net/t/1230254.aspx?HttpPlatformHandler+and+REMOTE_USER) remain unanswered to this day.

#### Final remarks

With version 2.2, ANCM will gain a fantastic new capability: running ASP.NET Core apps in-process. This ensures ASP.NET Core apps will have an edge over anything else running on Windows/IIS in terms of performance. 

To quote @shanselman from his [blog post](https://www.hanselman.com/blog/AnnouncingRunningRubyOnRailsOnIIS8OrAnythingElseReallyWithTheNewHttpPlatformHandler.aspx) a few years back:

> Why is HttpPlatformHandler interesting? Check this from the docs...it means IIS can host anything that runs on Windows now, easily. These things were possible before, but with all kinds of hacks and FastCGI this and that. What's great about HttpPlatformHandler is that it isn't about Rails. It's about any process that's listening on a port. You get all the value of IIS *and* total control of your self-hosting scenario.

The same is true of ANCM today. And with this PR, it would enable any out-of-process app (regardless of underlying technology) to really leverage the full value of IIS, including access to authenticated user information.

Another example would be an application running in a *Windows* Docker container. Such an app can be configured to run under a [Group Managed Service Account identity](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/manage-serviceaccounts#how-it-works), enabling it to use integrated Windows authentication against SQL Server and other networked services and resources. 

However, to my knowledge, there is currently no integrated auth solution for the *users accessing this application*. If the containerized application were proxied by IIS, then integrated authentication can be used, and ANCM could forward the IIS user name to the app. 

What do you think?


nil4:
Added a [third commit](https://github.com/aspnet/IISIntegration/pull/1214/commits/6659ab4e203875e878fc94f967cd62519536b642) that re-enables replacing `%ASPNETCORE_PORT%` placeholder in the ANCM process arguments.

This restores feature parity with HttpPlatformHandler, which performs a [similar replacement](https://technet.microsoft.com/en-us/library/mt125371.aspx):

> %HTTP_PLATFORM_PORT% is a special placeholder that must be specified as part of the arguments attribute or the httpPlatformEnvironmentVariables list.
> HttpPlatformHandler dynamically replaces this variable with a port so that the HTTP listener can listen on this port.

The argument replacement used to be available in earlier ANCM versions, but apparently 
[regressed during a refactoring](https://github.com/aspnet/AspNetCoreModule/issues/117#issuecomment-311748366). 

Hopefully fixes aspnet/AspNetCoreModule#117 and SuaveIO/Suave.IIS#6

nil4:
@Tratcher I presume that is because HttpSys can already do many of the same things that ANCM is used for: URL reservation, HTTPS certificate binding, Windows authentication etc. Other than easier administration, I'm not sure ANCM brings much value to that scenario. 

The use cases I see are centered around hosting non-native Windows apps (and lately, Dockerized apps) under IIS. Intranet users expect integrated authentication everywhere, and there are no good solutions today for those use cases. 

I hope this PR improves things a bit at least in *that* direction. And I'm glad if it helps with other scenarios I haven't thought of 👍. It goes both ways:

[Any Developer. Any App. Any Platform.](https://developer.microsoft.com/en-us/)

/cc @guardrex

nil4:
Rebased on top of the latest `release/2.2` changes (and for once all builds succeeded).

shirhatti:
@nil4 We appreciate the contribution but we are unable to accept this PR at this moment. This stems from the fact that we don't have enough hands on deck to deal with the long tail of issues from folks attempting to use this in a myriad of ways. The biggest burden here would be ensuring compatibility as we presently only test with ASP.NET Core.

> ANCM is architected as a generic reverse proxy (owing to its HttpPlatformHandler origins), but a couple of implementation choices limit its full potential to ASP.NET Core apps.

This is intentional decision to not build a generic reverse proxy, but limit ourselves to just ASP.NET Core workloads. 

That being said, we'd happy to help in whatever limited capacity we have to answer any questions you have that may arise from your fork.

Issue #1119, opened 2018-07-30
ANCM v1: App fails to start if ASPNETCORE_PORT env var set


mikeharder:
In ANCM v1, the app fails to start if the `ASPNETCORE_PORT` env var is set.

# Repro steps
1. Set `ASPNETCORE_PORT` in `web.config`, like so:
```
<aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="true">
  <environmentVariables>
    <!-- Uncomment to run on fixed port instead of random port -->
    <environmentVariable name="ASPNETCORE_PORT" value="10000" />
  </environmentVariables>
</aspNetCore>
```
2. Attempt to run app in ANCM v1
3. App fails with following in stdout log:
```
Unhandled Exception: System.ArgumentException: Item has already been added. Key in dictionary: 'ASPNETCORE_APPL_PATH'  Key being added: 'ASPNETCORE_APPL_PATH'
   at System.Collections.Hashtable.Insert(Object key, Object nvalue, Boolean add)
   at System.Environment.ToHashtable(IEnumerable`1 pairs)
   at System.Environment.GetEnvironmentVariables()
   at Microsoft.Extensions.Configuration.EnvironmentVariables.EnvironmentVariablesConfigurationProvider.Load()
   at Microsoft.Extensions.Configuration.ConfigurationRoot..ctor(IList`1 providers)
   at Microsoft.Extensions.Configuration.ConfigurationBuilder.Build()
   at Microsoft.AspNetCore.Hosting.WebHostBuilder..ctor()
   at Microsoft.AspNetCore.WebHost.CreateDefaultBuilder(String[] args)
```
4. If you dump the env vars at app startup (example: https://github.com/mikeharder/ancm-docker/blob/82cf503a9bc7c905df4e9ed0cf7e5caccbf73f1d/app/2.1/Program.cs#L20), you will see that `ASPNETCORE_PORT` is never set, and instead `ASPNETCORE_APPL_PATH` is set twice, which causes the failure:
```
ASPNETCORE_IIS_HTTPAUTH=anonymous;
ASPNETCORE_APPL_PATH=/2.1/OUTPROC
ASPNETCORE_HOSTINGSTARTUPASSEMBLIES=Microsoft.AspNetCore.Server.IISIntegration
ASPNETCORE_TOKEN=71b6d69c-7786-4c04-b613-5a6a6919d074
[...]

Issue #1223, opened 2018-08-13
Populate IServerAddresses feature for the in process IIS integration


javiercn:
@jkotalik

Tratcher:
What do you plan on using it for?

muratg:
@shirhatti please check with @javiercn & @BillHiebert to see what their scenarios are.

Tratcher:
The HttpsRedirect middleware would benifit from it. Not sure if they had other scenarios in mind.

pakrym:
Dup: https://github.com/aspnet/IISIntegration/issues/485

PR #1225, opened 2018-08-14
Remove early dereference of ASPNETCORE_PORT env var

mikeharder:
- Addresses #1119 
- Includes workaround for #1237


mikeharder:
Verified fix by manual testing.  Will add automated tests tomorrow.

test/Common.FunctionalTests/OutOfProcess/AspNetCorePortTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        private static readonly Random _random = new Random();
+
+        private readonly PublishedSitesFixture _fixture;
+
+        public AspNetCorePortTests(PublishedSitesFixture fixture)
+        {
+            _fixture = fixture;
+        }
+
+        public static TestMatrix TestVariants
```
pakrym: Would this return inprocess variants?

test/Common.FunctionalTests/OutOfProcess/AspNetCorePortTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            // Workaround for GetBaseDeploymentParameters hardcoding ANCM v2
+            deploymentParameters.AncmVersion = variant.AncmVersion;
+
+            var port = GetUnusedRandomPort().ToString();
+            deploymentParameters.WebConfigBasedEnvironmentVariables["ASPNETCORE_PORT"] = port;
+
+            var deploymentResult = await DeployAsync(deploymentParameters);
+
+            var response = await deploymentResult.HttpClient.GetAsync("/AspNetCorePortEnvVar");
```
pakrym: I would prefer asserting that we are actually listening on the specified port.

mikeharder: OK, this should work for outofprocess, but there's currently an issue claiming it doesn't work for inprocess: https://github.com/aspnet/IISIntegration/issues/1223

pakrym: We don't care about inprocess in this PR.

test/Common.FunctionalTests/OutOfProcess/AspNetCorePortTests.cs
mikeharder: Actually `OutOfProcess\HelloWorldTest.cs` doesn't set this explicitly either.  Do you think it should?

pakrym: I don't have a strong opinion on this. I think it's better to be explicit but your test would fail if ran with inproc anyway.

test/Common.FunctionalTests/OutOfProcess/AspNetCorePortTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        public async Task EnvVarInWebConfig(TestVariant variant)
+        {
+            // Must publish to set env vars in web.config
+            var deploymentParameters = _fixture.GetBaseDeploymentParameters(variant, publish: true);
+
+            // Workaround for GetBaseDeploymentParameters hardcoding ANCM v2
+            deploymentParameters.AncmVersion = variant.AncmVersion;
+
+            var port = GetUnusedRandomPort().ToString();
+            deploymentParameters.WebConfigBasedEnvironmentVariables["ASPNETCORE_PORT"] = port;
```
jkotalik: Consider adding a test that sends more than 1 environment variable.

mikeharder:
@pakrym: Changed test to use `IServerAddressesFeature`: https://github.com/aspnet/IISIntegration/pull/1225/files#diff-60889ab35232ca308d944d68ae0e2c84R53

PR #1231, opened 2018-08-14
Throw if web config env vars set without publish

mikeharder:


src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISExpressDeployer.cs
```diff
@@ -295,6 +295,12 @@ private void PrepareConfig(string contentRoot, int port)
             }
             else
             {
+                if (IISDeploymentParameters.WebConfigBasedEnvironmentVariables.Count > 0)
```
pakrym: I would prefer fixing logic inside RunWebConfigActions and calling it outside the if.

Issue #1237, opened 2018-08-14
PublishedSitesFixture.GetBaseDeploymentParameters() overrides too many properties of baseParameters


mikeharder:
https://github.com/aspnet/IISIntegration/blob/4b5a32d3b95299445163137b3443fb8f89389e27/test/Common.FunctionalTests/PublishedSitesFixture.cs#L57-L68

`PublishedSitesFixture.GetBaseDeploymentParameters()` overrides too many properties of `baseParameters`.  In particular, even if the test requests ANCM v1, this method silently forces it to ANCM v2, which means the wrong version is being tested.

As part of this change, we should also have at least one test which verifies the host process is actually using the correct version of ANCM at runtime.

Also, please revert the workaround added in https://github.com/aspnet/IISIntegration/pull/1225/files#diff-60889ab35232ca308d944d68ae0e2c84R41.

PR #1241, opened 2018-08-15
Fix deployment parameters override

pakrym:
https://github.com/aspnet/IISIntegration/issues/1237

In addition:

1. Make more logic common in IIS deployers
2. Get rid of token replacement in IISExpressDeployer
3. Remove our copy of IISExpress.config and use one that ships with integration testing package


test/Common.FunctionalTests/OutOfProcess/HelloWorldTest.cs
```diff
@@ -30,9 +33,16 @@ public static TestMatrix TestVariants
         [MemberData(nameof(TestVariants))]
         public async Task HelloWorld(TestVariant variant)
         {
-            // The default in hosting sets windows auth to true.
-            // Set it to the IISExpress.config file
             var deploymentParameters = _fixture.GetBaseDeploymentParameters(variant);
+            deploymentParameters.ServerConfigActionList.Add(
+                (element, _) => {
+                    element
+                        .RequiredElement("system.webServer")
```
pakrym: This was done in FunctionalTests copy of IISExpress config

Issue #1100, opened 2018-07-26
Flaky test: CheckShutdownEventLogMessage - StdOut Logging


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=515544&buildTypeId=Releases_22xPublic_Win2012) occasionally with the following error:
```
Assert.Contains() Failure
Not found: (filter expression)
In value:  ConcurrentQueue<WriteContext> [WriteContext { EventId = 0, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Server.IntegrationTesting.IIS"..., LogLevel = Information, ... }, WriteContext { EventId = 0, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Server.IntegrationTesting.IIS"..., LogLevel = Information, ... }, WriteContext { EventId = 0, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Server.IntegrationTesting.IIS"..., LogLevel = Information, ... }, WriteContext { EventId = 0, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Server.IntegrationTesting.IIS"..., LogLevel = Information, ... }, WriteContext { EventId = 0, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Server.IntegrationTesting.IIS"..., LogLevel = Information, ... }, ...]
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.EventLogHelpers.VerifyEventLogEvent(ITestSink testSink, String expectedRegexMatchString) in /_/test/Common.FunctionalTests/Utilities/EventLogHelpers.cs:line 16
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.EventLogTests.CheckShutdownEventLogMessage() in /_/test/Common.FunctionalTests/Inprocess/EventLogTests.cs:line 34
--- End of stack trace from previous location where exception was thrown ---
------- Stdout: -------
| [0.001s] TestLifetime Information: Starting test CheckShutdownEventLogMessage at 2018-07-25T22:43:55
[...]

PR #1242, opened 2018-08-15
Call StopServer before checking all event logs

pakrym:
Fixes https://github.com/aspnet/IISIntegration/issues/1100


pakrym:
https://github.com/aspnet/IISIntegration/issues/1049

PR #1244, opened 2018-08-15
[WIP] Collect crash dumps on VSTS

pakrym:


pakrym:
This makes VSTS build slightly longer because of AppVerif it's now ~22m

Issue #1196, opened 2018-08-09
Switching hostingModel from out-of-process to inprocess does not work


jhkimnew:
I can switch from in-process to out-of-process with simply changing the hostingModel value while application is running with some stress requests generated by stress test tool.

However I can not switch from out-of-process to inprocess with changing the hostingModel value to inprocess. 
The app is still running in the out-of-process mode even though the configuration is changed.

This is a regression issue.

Workaround is to force to recycle IIS worker process.

BTW, this issue seems to be reproducible without stress test. I found the similar issue without using stress testing. Here is the repro step.

STEPS TO REPRO:

1.	(Optional) Kill w3wp.exe to clean up
2.	Set hostingModel with empty value to set out-of-process
3.	Send a request
4.	Open TaskManager and you will see a new dotnet.exe process is running
5.	Set hostingModel with “inprocess” and save
6.	You will see the dotnet.exe process is gone
7.	Send a request

EXPECTED:

No dotnet.exe is created because hostingModel is now inprocess.

ACTUAL:

New dotnet.exe process starts even though the hostingModel is currently inprocess mode.

mikeharder:
@jhkimnew : Why do you say this is a regression?  Is it a regression from a previous build of ANCM v2?

jhkimnew:
@mikeharder Yes, I think so. I did not see this kind of issue when I used the old version before.

PR #1245, opened 2018-08-15
Add hosting model switch tests

pakrym:
New test doesn't repro https://github.com/aspnet/IISIntegration/issues/1196

Also:

1. Made NativeSample support outofproc if we need to change hosting model for debugging
2. Fixed throwing and catching of exceptions in GlobalVerstionUtility 
3. Cleaned-up eventlog usages

### DIFF
--- a/build/launchSettings.json
+++ b/build/launchSettings.json
@@ -15,8 +15,8 @@
       "nativeDebugging": true,
       "environmentVariables": {
         "IIS_SITE_PATH": "$(MSBuildThisFileDirectory)",
-        "ANCM_PATH": "$(AncmPath)",
-        "ANCMV2_PATH": "$(AncmV2Path)",
+        "ANCM_PATH": "$(TargetDir)\\$(AncmPath)",
+        "ANCMV2_PATH": "$(TargetDir)\\$(AncmV2Path)",
         "LAUNCHER_ARGS": "$(TargetPath)",
         "ASPNETCORE_ENVIRONMENT": "Development",
         "LAUNCHER_PATH": "$(DotNetPath)",
--- a/samples/NativeIISSample/Startup.cs
+++ b/samples/NativeIISSample/Startup.cs
@@ -14,8 +14,15 @@ namespace NativeIISSample
 {
     public class Startup
     {
+        private readonly IAuthenticationSchemeProvider _authSchemeProvider;
+
+        public Startup(IAuthenticationSchemeProvider authSchemeProvider = null)
+        {
+            _authSchemeProvider = authSchemeProvider;
+        }
+
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IHostingEnvironment env, IAuthenticationSchemeProvider authSchemeProvider)
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
             app.Run(async (context) =>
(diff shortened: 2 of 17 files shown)


build/launchSettings.json
```diff
@@ -15,8 +15,8 @@
       "nativeDebugging": true,
       "environmentVariables": {
         "IIS_SITE_PATH": "$(MSBuildThisFileDirectory)",
-        "ANCM_PATH": "$(AncmPath)",
-        "ANCMV2_PATH": "$(AncmV2Path)",
+        "ANCM_PATH": "$(TargetDir)\\$(AncmPath)",
```
pakrym: This change is to use applocal aspnetcorev2.dll and outofproc handler next to it.

samples/NativeIISSample/Startup.cs
```diff
@@ -14,8 +14,15 @@ namespace NativeIISSample
 {
     public class Startup
     {
+        private readonly IAuthenticationSchemeProvider _authSchemeProvider;
+
+        public Startup(IAuthenticationSchemeProvider authSchemeProvider = null)
+        {
+            _authSchemeProvider = authSchemeProvider;
```
jkotalik: Nit: why move the AuthSchemeProvider out of Configure?

src/AspNetCoreModuleV2/AspNetCore/applicationmanager.cpp
```diff
@@ -189,12 +189,6 @@ APPLICATION_MANAGER::RecycleApplicationFromManager(
 
         // Removed the applications which are impacted by the configurtion change
         m_pApplicationInfoHash->DeleteIf(FindConfigChangedApplication, (PVOID)&context);
-
-        if (m_pApplicationInfoHash->Count() == 0 && m_hostingModel == HOSTING_OUT_PROCESS)
-        {
-            // reuse current process
-            m_hostingModel = HOSTING_UNKNOWN;
```
jkotalik: Why remove this check? What happens when we run an out of process app, shut it down, then start an inprocess app?

pakrym: m_hostingModel  value never used. The logic you are talking about is in HandlerResolver

test/Common.FunctionalTests/ConfigurationChangeTests.cs
```diff
@@ -54,5 +55,24 @@ public async Task ConfigurationChangeForcesChildProcessRestart(AncmVersion versi
             // Verify that worker process gets restarted with new process id
             await deploymentResult.HttpClient.RetryRequestAsync("/ProcessId", async r => await r.Content.ReadAsStringAsync() != processBefore);
         }
+
+        [ConditionalFact]
+        public async Task OutOfProcessToInProcessHostingModelSwitchWorks()
```
jkotalik: What about inprocess to out of process and expect a new process to start.

samples/NativeIISSample/Startup.cs
pakrym: It's not available in outofproc so I've made it optional

test/Common.FunctionalTests/ConfigurationChangeTests.cs
pakrym: It won't start our testsites do not support switching hosting mode.

PR #1249, opened 2018-08-17
Upgrade build tools to use the 2.1.400 SDK

natemcmaster:
Required a workaround for a regression in MSBuild 15.8.


pakrym:
It is a similar failure but not the same issue. And it's definitely caused by SDK upgrade other PRs are passing fine.

pakrym:
Bug in new websdk?

jkotalik:
I'll grab this branch and try building it locally. WebSDK issue seems the most likely.

natemcmaster:
cc @vijayrkn - what changed between 2.1.302 and 2.1.400 in the SDK which might cause issues in web.config generation?

vijayrkn:
Could it be related to this - https://github.com/aspnet/websdk/pull/377
We add the location tag by default for all new web.configs

vijayrkn:
Location tag is only added for generated web.configs though. So, if the app already have a web.config this should not be applicable.

jkotalik:
For out of process, the web config looks like this:

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <location path="." inheritInChildApplications="false">
    <system.webServer>
      <handlers>
      </handlers>
      <aspNetCore processPath="dotnet" arguments=".\OutOfProcessWebSite.dll" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" />
    </system.webServer>
  </location>
  <system.webServer>
    <aspNetCore hostingModel="" />
    <handlers>
      <add modules="AspNetCoreModuleV2" />
    </handlers>
  </system.webServer>
</configuration>
<!--ProjectGuid: 42E60F88-E23F-417A-8143-0CCEC05E1D02-->
```

so it does seem to be an issue with the location.

jkotalik:
@pakrym looks like we will need to update tests that modify the web.config.

vijayrkn:
Is this the generated web.config ? Websdk should not be modifying it this way.

natemcmaster:
After thought - would this also affect tests in aspnet/Hosting, ServerTests, etc? cc @Tratcher

jkotalik:
@natemcmaster are feature branches currently active? I don't think it will because none of the tests in server tests modify the web.config after publish (only the apphost.config).

natemcmaster:
> are feature branches currently active?

Not for more repos. We disabled by default, selectively kept ones that matter

jkotalik:
All changes except one are in the deployers, so changes should be picked up in other repos.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/WebConfigHelpers.cs
```diff
@@ -12,7 +13,7 @@ public static class WebConfigHelpers
         {
             return (element, _) => {
                 element
-                    .GetOrAdd("system.webServer")
+                    .Descendants("system.webServer").Single()
```
pakrym: formatting

natemcmaster:
Thanks @jkotalik. Tests pass now. Okay to merge?

jkotalik:
Missed some nits, should be fine to merge. Wait for one test to pass for sanity.

PR #1250, opened 2018-08-17
Cleanup application manager

pakrym:
1. Remove manual ref-counting from APPPLICATION_INFO
2. Added config change stress test
3. Modified the old test to assert the right thing


src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
```diff
@@ -17,129 +17,140 @@ const PCWSTR HandlerResolver::s_pwzAspnetcoreOutOfProcessRequestHandlerName = L"
...
+    auto pfnAspNetCoreCreateApplication = reinterpret_cast<PFN_ASPNETCORE_CREATE_APPLICATION>(GetProcAddress(hRequestHandlerDll, "CreateApplication"));
+    RETURN_LAST_ERROR_IF_NULL(pfnAspNetCoreCreateApplication);
 
+    pApplicationFactory = std::make_unique<ApplicationFactory>(hRequestHandlerDll.release(), location, pfnAspNetCoreCreateApplication);
     return S_OK;
 }
 
 HRESULT
-HandlerResolver::GetApplicationFactory(IHttpApplication &pApplication, STRU& location, PFN_ASPNETCORE_CREATE_APPLICATION * pfnCreateApplication)
+HandlerResolver::GetApplicationFactory(IHttpApplication &pApplication, std::unique_ptr<ApplicationFactory>& pApplicationFactory)
```
jkotalik: Function name is confusing, you aren't getting the factory at all.

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
```diff
@@ -177,33 +186,29 @@ HRESULT
...
-    STRU        struNativeSearchPaths;
-    STRU        struNativeDllLocation;
-    INT         intIndex = -1;
-    INT         intPrevIndex = 0;
-    BOOL        fFound = FALSE;
-    DWORD       dwBufferSize = 1024 * 10;
-    DWORD       dwRequiredBufferSize = 0;
-
-    DBG_ASSERT(struFilename != NULL);
+    std::wstring   struNativeSearchPaths;
```
jkotalik: Where is the closest try/catch for this function? May want to wrap try/catch around this call?

jkotalik: Also, make this return void, HRESULT is never used.

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
pakrym: You are getting ApplicationFactory object

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
pakrym: in proxymodule, I'll look at it

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
jkotalik: I see now, my mistake. This is more of an initialize method though. with a call to LoadRequestHandlerAssembly. Can you move the initialize into the constructor?

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
```diff
@@ -177,33 +186,29 @@ HRESULT
...
+    size_t         intIndex = 0;
+    size_t         intPrevIndex = 0;
+    bool           fFound = FALSE;
+    DWORD          dwBufferSize = 1024 * 10;
```
jkotalik: Some of these can probably be const?

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
```diff
@@ -225,26 +236,26 @@ HandlerResolver::FindNativeAssemblyFromHostfxr(
...
     // The native search directories are semicolon delimited.
     // Split on semicolons, append aspnetcorerh.dll, and check if the file exists.
-    while ((intIndex = struNativeSearchPaths.IndexOf(L";", intPrevIndex)) != -1)
+    while ((intIndex = struNativeSearchPaths.find(L';', intPrevIndex)) != std::wstring::npos)
     {
-        RETURN_IF_FAILED(struNativeDllLocation.Copy(&struNativeSearchPaths.QueryStr()[intPrevIndex], intIndex - intPrevIndex));
+        auto path = struNativeSearchPaths.substr(intPrevIndex, intIndex - intPrevIndex);
 
-        if (!struNativeDllLocation.EndsWith(L"\\"))
+        if (!path.empty() && !(path[path.length()] == L'\\'))
```
jkotalik: Shouldn't this be path.length() -1? Consider using std::filesystem::path here too.

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
pakrym: It's used for logging and eventlogs

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
pakrym: Why is it initialize? It's called per application_info creation

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
```diff
@@ -225,26 +236,26 @@ HandlerResolver::FindNativeAssemblyFromHostfxr(
         }
     }
 
-    RETURN_IF_FAILED(struNativeSearchPaths.SyncWithBuffer());
+    struNativeSearchPaths.resize(struNativeSearchPaths.find(L'\0'));
```
jkotalik: Seems to be repeated operations between this line and the one prior.

pakrym: `            struNativeSearchPaths.resize(dwBufferSize);` only kicks in when there is enough space. In successfull case dwRequiredBufferSize would be 0 unfortunately.

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
jkotalik: I see now, the HandlerResolver object is part of the application manager, so you never call new HandlerResolver.

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.h
```diff
@@ -19,30 +21,53 @@ HRESULT
     _Out_ IAPPLICATION         **pApplication
     );
 
+class ApplicationFactory
```
jkotalik: Separate file; too much stuff in here to have in a shared header file.

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.h
```diff
@@ -19,30 +21,53 @@ HRESULT
...
     HandlerResolver(HMODULE hModule, IHttpServer &pServer);
-    HRESULT GetApplicationFactory(IHttpApplication &pApplication, STRU& location, PFN_ASPNETCORE_CREATE_APPLICATION *pfnCreateApplication);
+    HRESULT GetApplicationFactory(IHttpApplication &pApplication, std::unique_ptr<ApplicationFactory>& pApplicationFactory);
+    void ResetHostingModel();
 
 private:
-    HRESULT LoadRequestHandlerAssembly(IHttpApplication &pApplication,STRU& location, ASPNETCORE_SHIM_CONFIG * pConfiguration);
-    HRESULT FindNativeAssemblyFromGlobalLocation(ASPNETCORE_SHIM_CONFIG * pConfiguration, PCWSTR libraryName, STRU* location);
-    HRESULT FindNativeAssemblyFromHostfxr(HOSTFXR_OPTIONS* hostfxrOptions, PCWSTR libraryName, STRU* location);
+    HRESULT LoadRequestHandlerAssembly(IHttpApplication &pApplication, ASPNETCORE_SHIM_CONFIG * pConfiguration, std::unique_ptr<ApplicationFactory>& pApplicationFactory);
```
jkotalik: ASPNETCORE_SHIM_CONFIG * pConfiguration -> ASPNETCORE_SHIM_CONFIG  &pConfiguration?

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
pakrym: Yep

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -3,36 +3,21 @@
...
 #include "exceptions.h"
 #include "EventLog.h"
-#include "HandleWrapper.h"
 #include "ServerErrorApplication.h"
 #include "AppOfflineApplication.h"
 
 APPLICATION_INFO::~APPLICATION_INFO()
 {
-    ShutDownApplication();
+    ShutDownApplication(false);
```
jkotalik: nit comment on what the parameter is.

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -54,6 +39,7 @@ APPLICATION_INFO::GetOrCreateApplication(
             // Call to wait for application to complete stopping
             m_pApplication->Stop(/* fServerInitiated */ false);
             m_pApplication = nullptr;
+            m_pApplicationFactory = nullptr;
```
jkotalik: Nit: helper method to call these 3 lines (repeated in shutdown application).

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.h
```diff
@@ -19,150 +17,66 @@ class APPLICATION_INFO
...
+    APPLICATION_INFO(
+        IHttpServer &pServer,
+        IHttpApplication    &pApplication,
+        HandlerResolver     *pHandlerResolver
+    ) :
         m_pServer(pServer),
-        m_cRefs(1),
-        m_handlerResolver(nullptr)
+        m_handlerResolver(pHandlerResolver),
+        m_struConfigPath(pApplication.GetAppConfigPath()),
```
jkotalik: nit, remove stru from names?

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.h
```diff
@@ -19,150 +17,66 @@ class APPLICATION_INFO
 {
 public:
 
-    APPLICATION_INFO(IHttpServer &pServer) :
+    APPLICATION_INFO(
+        IHttpServer &pServer,
+        IHttpApplication    &pApplication,
+        HandlerResolver     *pHandlerResolver
```
jkotalik: nit: why can't HandlerResolver be a ref?

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.h
```diff
@@ -19,150 +17,66 @@ class APPLICATION_INFO
...
-    VOID
-    DereferenceApplicationInfo() const
+    bool ConfigurationPathApplies(const std::wstring& path)
     {
-        if (InterlockedDecrement(&m_cRefs) == 0)
+        // We need to check that the last character of the config path
+        // is either a null terminator or a slash.
+        // This checks the case where the config path was
+        // MACHINE/WEBROOT/site and your site path is MACHINE/WEBROOT/siteTest
+        auto const changed = m_struConfigPath._Starts_with(path);
```
jkotalik: where does `_Starts_with` come from?

pakrym: microsoft implementation of std::wstring

src/AspNetCoreModuleV2/AspNetCore/applicationmanager.cpp
```diff
@@ -148,101 +87,82 @@ APPLICATION_MANAGER::RecycleApplicationFromManager(
...
+                else
+                {
+                    ++itr;
+                }
+            }
+
+            // All applications were unloaded reset handler resolver validation logic
+            if (m_pApplicationInfoHash.empty())
+            {
+                m_handlerResolver.ResetHostingModel();
```
jkotalik: Idea: Instead of resetting the handler resolver, we could just create it on creating the first application info and delete it when empty?

src/AspNetCoreModuleV2/AspNetCore/aspnetcore_shim_config.h
```diff
@@ -63,9 +62,9 @@ class ASPNETCORE_SHIM_CONFIG
 
 private:
 
-    STRU                   m_struArguments;
-    STRU                   m_struProcessPath;
-    APP_HOSTING_MODEL      m_hostingModel;
-    STRU                   m_struHandlerVersion;
+    std::wstring                   m_struArguments;
```
jkotalik: nit: remove stru from names.

src/AspNetCoreModuleV2/CommonLib/HandleWrapper.h
```diff
@@ -51,6 +51,13 @@ class HandleWrapper
 
     HandleType* operator&() { return &m_handle; }
 
+    HandleType release() noexcept
+    {
+        auto value = m_handle;
```
jkotalik: Reason for this change?

src/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
```diff
@@ -1283,7 +1283,7 @@ SERVER_PROCESS::SendSignal(
     // Do it only for the case that debugger is attached during process creation
     // as IsDebuggerIsAttached call is too heavy
     //
-    if (WaitForSingleObject(m_hShutdownHandle, m_fDebuggerAttached ? INFINITE : m_dwShutdownTimeLimitInMS) != WAIT_OBJECT_0)
+    if (WaitForSingleObject(m_hShutdownHandle, INFINITE) != WAIT_OBJECT_0)
```
jkotalik: ?

test/Common.FunctionalTests/ConfigurationChangeTests.cs
```diff
@@ -71,8 +77,48 @@ public async Task OutOfProcessToInProcessHostingModelSwitchWorks()
...
+
+            await deploymentResult.AssertStarts();
+            var load = Helpers.StressLoad(deploymentResult.HttpClient, "/HelloWorld", response => {
+                var statusCode = (int)response.StatusCode;
+                Assert.True(statusCode == 200 || statusCode == 503, "Status code was " + statusCode);
+            });
+
+            for (int i = 0; i < 100; i++)
+            {
+                // AddAppOffline might fail if app_offline is being read by ANCM and deleted at the same time
```
jkotalik: I mentioned this before, but having an equivalent check with apphost.config would be useful.

src/AspNetCoreModuleV2/CommonLib/HandleWrapper.h
pakrym: Because we use it now to transfer handle ownership from LoadRequestHandlerAssembly to ApplicationFactory.

src/AspNetCoreModuleV2/AspNetCore/applicationmanager.cpp
pakrym: I'll have to dynamically allocate it and add lazy loading logic. Not sure it's better.

src/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
pakrym: Debugging leftovers

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
jkotalik: I see.

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
```diff
@@ -17,129 +17,146 @@ const PCWSTR HandlerResolver::s_pwzAspnetcoreOutOfProcessRequestHandlerName = L"
...
-        LOG_INFOF("Loading request handler:  %S", struFileName.QueryStr());
+        LOG_INFOF("Loading request handler:  %S", handlerDllPath.c_str());
 
-        m_hRequestHandlerDll = LoadLibraryW(struFileName.QueryStr());
-        RETURN_LAST_ERROR_IF_NULL(m_hRequestHandlerDll);
+        hRequestHandlerDll = LoadLibrary(handlerDllPath.c_str());
+        if (preventUnload)
+        {
+            // Pin module in memory
+            GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_PIN, pstrHandlerDllName, &hRequestHandlerDll);
```
jkotalik: What happens when you call FreeModule on hRequestHandlerDll when it is pinned?

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
jkotalik: k

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.h
```diff
@@ -19,150 +17,66 @@ class APPLICATION_INFO
...
-        if (InterlockedDecrement(&m_cRefs) == 0)
+        // We need to check that the last character of the config path
+        // is either a null terminator or a slash.
+        // This checks the case where the config path was
+        // MACHINE/WEBROOT/site and your site path is MACHINE/WEBROOT/siteTest
+        auto const changed = m_strConfigPath._Starts_with(path);
+        if (changed)
         {
-            delete this;
+            const auto lastChar = m_strConfigPath[m_strConfigPath.length()];
```
jkotalik: bruh.

jkotalik: Actually this may be correct. Will indexing into the null character cause an exception?

pakrym: Nope

src/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
pakrym: Nothing

src/AspNetCoreModuleV2/AspNetCore/applicationinfo.h
jkotalik: https://stackoverflow.com/questions/17939323/is-wstring-null-terminated Yeah you seem to ge good.

jkotalik:
ConfigurationTouchedStress failed.

PR #1254, opened 2018-08-17
Free environment variable table in inprocess application

jkotalik:
Found using appverifier. Test will be enabled through adding app verifier to VSTS (#1244 attempted to do it, will enable once all failures are resolved).


src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -403,18 +403,19 @@ IN_PROCESS_APPLICATION::SetEnvironementVariablesOnWorkerProcess(
...
-    pHashTable->Apply(ENVIRONMENT_VAR_HELPERS::AppendEnvironmentVariables, &hr);
-    if (FAILED(hr))
-    {
-        goto Finished;
-    }
-    pHashTable->Apply(ENVIRONMENT_VAR_HELPERS::SetEnvironmentVariables, &hr);
+    table->Apply(ENVIRONMENT_VAR_HELPERS::AppendEnvironmentVariables, &hr);
     if (FAILED(hr))
     {
         goto Finished;
```
pakrym: macrosify

src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -403,18 +403,19 @@ IN_PROCESS_APPLICATION::SetEnvironementVariablesOnWorkerProcess(
...
     }
+
+    table->Apply(ENVIRONMENT_VAR_HELPERS::SetEnvironmentVariables, &hr);
+
 Finished:
+    table->Clear();
```
pakrym: You can add a deleter that would clear table too.

jkotalik: Yeah I'll do that.

pakrym:
I don't like appveyor failure

jkotalik:
Yeah that one seems concerning. I doubt it's related though?

pakrym:
I've never seen it, and it might be caused by a  shutdown AV.

jkotalik:
`AppOfflineDropped_CanRemoveAppOfflineAfterAddingAndSiteWorks` seems to be unrelated to this change. I'll file an issue.

pakrym:
Rerun in too

PR #1266, opened 2018-08-18
Fix typo in process start error message

mikeharder:


pakrym:
Want to add a test? 😀

pakrym:
I was joking earlier but now think we really should do it, I'm sure we have some test that logs this event log message, would be nice to add a verification to it while we do all these fixes.

mikeharder:
@pakrym, @jkotalik : I suggest we use this issue to track adding tests for EventLog messages: https://github.com/aspnet/IISIntegration/issues/1270

Issue #1261, opened 2018-08-17
ANCM v1: Event ID 1000 (ASPNETCORE_EVENT_PROCESS_START_ERROR) has no message


mikeharder:
In ANCM v1, if an app fails to start, an event with ID 1000 (`ASPNETCORE_EVENT_PROCESS_START_ERROR`) is raised.  However, the event has no message.

* Repros with the last public release of ANCM v1.
* Does not repro with ANCM v2.
* Note this is a different issue than https://github.com/aspnet/IISIntegration/issues/1128, which only impacts our private builds.

# Repro Steps
1. Create an app which fails to start.  For example, set an invalid value for `ASPNETCORE_PORT` in web.config:
```
<aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="true">
  <environmentVariables>
    <environmentVariable name="ASPNETCORE_PORT" value="foo" />
  </environmentVariables>
</aspNetCore>
```
2. Try to start app in IIS.
3. Check event log, and notice that event id 1000 has no message.
```
Message      :
Id           : 1000
ProviderName : IIS AspNetCore Module
TimeCreated  : 8/17/2018 3:37:04 PM

Message      : Sent shutdown HTTP message to process '2324' and received http status '202'.
Id           : 1006
ProviderName : IIS AspNetCore Module
TimeCreated  : 8/17/2018 3:36:05 PM

Message      : Application 'MACHINE/WEBROOT/APPHOST/DEFAULT WEB SITE/2.1/OUTPROC' started process '2324' successfully and is listening on port '10064'.
Id           : 1001
ProviderName : IIS AspNetCore Module
TimeCreated  : 8/17/2018 3:28:59 PM
```

# Root Cause
Printf parameters passed in wrong order (format string must be first parameter):

https://github.com/aspnet/IISIntegration/blob/13281620521e5bab9313fbc04ae05f6bcf790345/src/AspNetCoreModuleV1/AspNetCore/src/serverprocess.cxx#L1070-L1073

PR #1267, opened 2018-08-18
Fix process start error message

mikeharder:
- Addresses #1261


src/AspNetCoreModuleV1/AspNetCore/src/serverprocess.cxx
```diff
@@ -1068,8 +1068,8 @@ SERVER_PROCESS::StartProcess(
         {
             if (!fDonePrepareCommandLine)
                 strEventMsg.SafeSnwprintf(
-                m_struAppFullPath.QueryStr(),
```
mikeharder: Parameters were passed in wrong order (format string must be first parameter)

src/AspNetCoreModuleV1/AspNetCore/src/serverprocess.cxx
```diff
@@ -1068,8 +1068,8 @@ SERVER_PROCESS::StartProcess(
...
                 ASPNETCORE_EVENT_PROCESS_START_INTERNAL_ERROR_MSG,
+                m_struAppFullPath.QueryStr(),
                 hr);
```
mikeharder: Parameters were passed in wrong order (format string must be first)

PR #1276, opened 2018-08-20
Update schema for ANCMv1

jkotalik:
We would like to keep the ANCM schemas in V1 and V2 the same. As ANCMV1 will be shipped before ANCMV1 + ANCMV2, we would like to put the new schema in now for V1.


vivmishra:
Approved for 2.1.5

PR #1288, opened 2018-08-21
Disable tests that require environment variables.

jkotalik:
Fixing the build.


jkotalik:
Approved offline. Will wait for builds to finish.

PR #1291, opened 2018-08-21
Enable app verifier on VSTS

jkotalik:


pakrym:
There are crash dumps in VSTS artifacts

jkotalik:
Will make separate PRs for failures if VSTS runs pass.

jkotalik:
Test failures: #1049

PR #1286, opened 2018-08-21
Update ANCM version to 12.1

jkotalik:
For https://github.com/aspnet/IISIntegration/issues/1258

PR #1320, opened 2018-08-23
Update branding for 2.1.5

natemcmaster:
Changes the branding for the 2.1.5 release because this repo is taking patches

### DIFF
--- a/version.props
+++ b/version.props
@@ -1,6 +1,6 @@
 ﻿<Project>
   <PropertyGroup>
-    <VersionPrefix>2.1.1</VersionPrefix>
+    <VersionPrefix>2.1.2</VersionPrefix>
     <VersionSuffix>rtm</VersionSuffix>
     <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' == 'rtm' ">$(VersionPrefix)</PackageVersion>
     <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' != 'rtm' ">$(VersionPrefix)-$(VersionSuffix)-final</PackageVersion>


natemcmaster:
FYI @shirhatti @pakrym @jkotalik - when this merges, you are free to add other approved 2.1.5 work to release/2.1.

jkotalik:
Actually, if there are no code changes in Microsoft.AspNetCore.Server.IISIntegration, do we need to update the version? All changes are in ANCM.

natemcmaster:
Do you use the version.props file to set versions in aspnetcoremodulev1.dll?

jkotalik:
We use version.props, but other versions: https://github.com/aspnet/IISIntegration/pull/1320/files#diff-2a92b4d7f8df251ffd3a0aa63e97aad5L11

natemcmaster:
I see. My patching scripts didn't account for these variables. Presumably these should change too, right?

jkotalik:
They will: https://github.com/aspnet/IISIntegration/pull/1286

natemcmaster:
I see.

Generally, our philosophy has been "if one thing in a repo patches, everything in the repo patches." Our current build system is not flexible enough to partially re-build a repo. Even though patching the managed bits is unnecessary, but at the Universe layer all we have is an on/off switch. See https://github.com/aspnet/Universe/pull/1328/files#diff-98c23780ba4fa4d6a335ead2dc489fe0

jkotalik:
> if one thing in a repo patches, everything in the repo patches

Didn't realize that was the policy; I assumed versions only increase if an assembly or one of the assembly's dependencies change, not at a repo level.

jkotalik:
So I guess it's fine to increase the version. I'll modify the other PR to use the `VerisonPrefix` value.

jkotalik:
@natemcmaster can we merge this so I can finalize all 2.1.5 work?

natemcmaster:
Not yet. The release/2.1 branch isn't open yet. I'll notify the team when they are.

Issue #1319, opened 2018-08-23
hostfxr is being searched in application root and not next to the .exe


pakrym:
Should be fixed by recent changes. Add test for the scenario.

PR #1331, opened 2018-08-24
Add tests for processpath and arguments

pakrym:
https://github.com/aspnet/IISIntegration/issues/1319

1. Make all logging use wide strings to prevent failures when using unicode chars and `%S` format
2. Fix bugs found while adding tests


samples/NativeIISSample/web.config
```diff
@@ -4,6 +4,6 @@
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified" />
     </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" forwardWindowsAuthToken="false" stdoutLogEnabled="false" hostingModel="inprocess"/>
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS% &quot;extra argument&quot;" forwardWindowsAuthToken="false" stdoutLogEnabled="false" hostingModel="inprocess"/>
```
jkotalik: Cleanup debugging stuff.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -116,7 +116,7 @@ bool CreateDebugLogFile(const std::wstring &debugOutputFile)
         {
             if (g_logFile != INVALID_HANDLE_VALUE)
             {
-                LOG_INFOF("Switching debug log files to %S", debugOutputFile.c_str());
+                LOG_INFOF(L"Switching debug log files to '%ls'%", debugOutputFile.c_str());
```
jkotalik: nit.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -266,42 +266,99 @@ IsEnabled(
     return ( dwFlag & DEBUG_FLAGS_VAR );
 }
 
+void WriteFileEncoded(UINT codePage, HANDLE hFile, const LPCWSTR  szString)
+{
+    DWORD nBytesWritten = 0;
+    auto const encodedByteCount = WideCharToMultiByte(codePage, 0, szString, -1, nullptr, 0, nullptr, nullptr);
```
jkotalik: Pass in `WC_ERR_INVALID_CHARS` for flags? See: https://docs.microsoft.com/en-us/windows/desktop/api/stringapiset/nf-stringapiset-widechartomultibyte

pakrym: We want to see debug logs even if they contain invalid chars.

src/AspNetCoreModuleV2/CommonLib/hostfxr_utility.cpp
```diff
@@ -137,29 +138,49 @@ HOSTFXR_UTILITY::IsDotnetExecutable(const std::filesystem::path & dotnetPath)
...
     std::vector<std::wstring> &arguments,
     bool expandDllPaths
 )
 {
-    LOG_INFOF("Resolving hostfxr_main arguments application: '%S' arguments: '%S' path: %S", applicationExePath.c_str(), applicationArguments.c_str(), applicationPhysicalPath.c_str());
-
-    arguments = std::vector<std::wstring>();
-    arguments.push_back(applicationExePath);
+    // don't throw while trying to expand arguments
+    std::error_code ec;
```
jkotalik: Do we really need to? What errors will is_regular_file throw that are considered non fatal?

pakrym: arguments might contain invalid path characters. Tests fail without this.

test/Common.FunctionalTests/Utilities/Helpers.cs
```diff
@@ -64,12 +65,15 @@ public static void CopyFiles(DirectoryInfo source, DirectoryInfo target, ILogger
         {
             foreach (DirectoryInfo directoryInfo in source.GetDirectories())
             {
-                CopyFiles(directoryInfo, target.CreateSubdirectory(directoryInfo.Name), logger);
+                if (directoryInfo.FullName != target.FullName)
```
jkotalik: Why was this check needed?

pakrym: To avoid infinite recursion when directory it's copied into it's subdirectory.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -266,42 +266,98 @@ IsEnabled(
...
 VOID
-DebugPrint(
+DebugPrintW(
     DWORD   dwFlag,
-    const LPCSTR  szString
+    const LPCWSTR  szString
     )
 {
-    STACK_STRA (strOutput, 256);
+    STACK_STRU (strOutput, 256);
```
jkotalik: Were these restored functions that were present before?

pakrym: Nope, before they went the other direction wide strings were narrowed before being logged.

PR #1336, opened 2018-08-27
Re-enable --no-restore

pakrym:
Restoring is what caused mass test failures

And it seems it's not strictly required.


pakrym:
@Tratcher approved offline. Merging to fix tests.

PR #1337, opened 2018-08-27
Add Trace level for request events, allow logging to eventlog

pakrym:
Out of proc has many logs that are written during request processing that might flood debug log, add another level for them.

Allow eventlog as debug log target while I'm at it.


src/AspNetCoreModuleV2/CommonLib/EventLog.cpp
```diff
@@ -5,43 +5,56 @@
...
-    static const auto versionInfoString = GetVersionInfoString();
-
-    std::array<LPCWSTR, 3> eventLogDataStrings
-    {
-        pstrMsg,
-        processIdString.c_str(),
-        versionInfoString.c_str()
-    };
-
     if (g_hEventLog != NULL)
```
jkotalik: nit: invert if.

src/AspNetCoreModuleV2/CommonLib/debugutil.cpp
```diff
@@ -103,6 +105,7 @@ void SetDebugFlags(const std::wstring &debugValue)
             if (value >= 3) DEBUG_FLAGS_VAR |= ASPNETCORE_DEBUG_FLAG_INFO;
             if (value >= 4) DEBUG_FLAGS_VAR |= ASPNETCORE_DEBUG_FLAG_CONSOLE;
             if (value >= 5) DEBUG_FLAGS_VAR |= ASPNETCORE_DEBUG_FLAG_FILE;
+            if (value >= 6) DEBUG_FLAGS_VAR |= ASPNETCORE_DEBUG_FLAG_EVENTLOG;
```
jkotalik: I think EVENTLOG should be above console and file.

pakrym: Why?

jkotalik: Event logs are more accessible for debugging on most platforms (azure). No other big reason besides that.

pakrym: My thinking was that eventlog is the most inappropriate place to dump debug logs. Also, I would like to deprecate numbers anyway.

jkotalik: Yeah let's drop the numbers.

test/Common.FunctionalTests/Utilities/EventLogHelpers.cs
```diff
@@ -1,66 +1,65 @@
...
 
-            // Perf: only get the last 20 events from the event log.
+            // Perf: only get the last 20 events from the event log by default.
             // Eventlog is already sorted based on time of event in ascending time.
             // Add results in reverse order.
-            var expectedRegexEventLog = new Regex(expectedRegexMatchString);
             var processIdString = $"Process Id: {deploymentResult.HostProcess.Id}.";
 
-            for (var i = eventLog.Entries.Count - 1; i >= eventLog.Entries.Count - 20; i--)
+            for (var i = eventLog.Entries.Count - 1; i >= eventLog.Entries.Count - messageCount; i--)
```
jkotalik: Offline feedback: remove message count: https://github.com/aspnet/IISIntegration/pull/1338

Issue #1128, opened 2018-07-31
Event logs broken in ANCM v1 private builds


mikeharder:
Event logs are broken when using a private build of ANCM v1, in either of these scenarios:

1. Install [v1 hosting bundle](https://www.microsoft.com/net/download/dotnet-core/2.1), build ANCM v1 aspnetcore.dll locally, copy local build to `C:\windows\system32\inetsrv`
2. Install `aspnetcoremodule_x64_en.msi` built from aspnetci

Here's the broken event log:

```
> get-winevent -ProviderName 'IIS AspNetCore Module' | format-list -property *
get-winevent : The specified resource type cannot be found in the image file
At line:1 char:1
+ get-winevent -ProviderName 'IIS AspNetCore Module' | format-list -pro ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Get-WinEvent], EventLogException
    + FullyQualifiedErrorId : The specified resource type cannot be found in the image file,Microsoft.PowerShell.Commands.GetWinEventCommand

Message              : Cannot retrieve event message text.
Id                   : 1001
Version              :
Qualifiers           : 0
Level                : 4
Task                 : 0
Opcode               :
Keywords             : 36028797018963968
RecordId             : 107
ProviderName         : IIS AspNetCore Module
```

muratg:
The PR got approved for 2.1.5.

PR #1344, opened 2018-08-28
Add rc file as an include for ANCMV1 (#1284)

jkotalik:
## Description
 
Include .rc message file in the aspnetcoremodule.rc file.
 
## Customer Impact
 
Event messages will not work at all in ANCM. This will be a large regression.
 
## Regression from?
 
Regression from when we've moved the source code from SD to GitHub.
 
## Risk
 
Very low. We are just including a .rc file to get event messages.

cc @leecow @muratg


shirhatti:
https://github.com/aspnet/IISIntegration/issues/1128

src/AspNetCoreModuleV1/AspNetCore/aspnetcoremodule.rc
```diff
@@ -3,6 +3,7 @@
 #include <windows.h>
 #include "version.h"
 #include "resource.h"
+#include "Aspnetcore_msg.rc"
```
muratg: Nit: The actual filename starts with lowercase 'a'.  (Not a huge deal in Windows, but let's better be correct.)

jamshedd:
Approved for 2.1.5

PR #1355, opened 2018-08-30
Remove redundant rc files

jkotalik:
aspnetcore_msg.rc is a generated file, some were accidentally committed.


jkotalik:
cc @muratg

PR #1359, opened 2018-08-31
Merge QueryStatus and CreateHandler calls to make them atomic

pakrym:
Requires fewer lock acquisitions and provides stronger guarantees because we check status before creating handler.


src/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -19,76 +19,116 @@ APPLICATION_INFO::~APPLICATION_INFO()
...
+            return S_OK;
         }
-        else
+    }
+    
+    {
+        SRWExclusiveLock lock(m_applicationLock);
+        
+        // check if other thread created application
+        RETURN_IF_FAILED(hr = TryCreateHandler(pHttpContext, pHandler));
```
jkotalik: This section is a bit vague, but I think I get it. Can you expand/elaborate about the order of these calls as comments?

jkotalik:
Seems like in the debug build, tests exited preemptively.

Issue #1206, opened 2018-08-10
Verify behaviour in case of invalid message body


pakrym:
For example badly formatted chunks.

Tratcher:
On the request? For bad request chunks you shouldn't even receive them, Http.Sys should fail.

pakrym:
Is there any scenario where we get to start reading the body and while the client is sending it realize it's invalid?

pakrym:
> perhaps, but it would be reported as a read error or disconnect.

I want to verify that we get an exception that we expect.

PR #1364, opened 2018-09-04
Use ConnectionResetException with 'The client disconnected' message

pakrym:
https://github.com/aspnet/IISIntegration/issues/1206


test/IIS.Tests/ClientDisconnectTests.cs
```diff
@@ -17,7 +18,7 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
     public class ClientDisconnectTests : LoggedTest
     {
          
-        [ConditionalFact(Skip = "See: https://github.com/aspnet/IISIntegration/issues/1075")]
+        [ConditionalFact]
```
pakrym: I re-enabled some tests that should be unaffected by pipelines bug.

src/Microsoft.AspNetCore.Server.IIS/Core/IO/AsyncIOOperation.cs
```diff
@@ -103,7 +104,8 @@ public AsyncContinuation Complete(int hr, int bytes)
                 _result = bytes;
                 if (hr != NativeMethods.HR_OK)
                 {
-                    _exception = new IOException("Native IO operation failed", Marshal.GetExceptionForHR(hr));
+                    // Treat all errors as the client disconnect
+                    _exception = new ConnectionResetException("The client disconnected", Marshal.GetExceptionForHR(hr));
```
jkotalik: nit: The client has disconnected.

test/IIS.Tests/ClientDisconnectTests.cs
```diff
@@ -163,6 +163,51 @@ public async Task ReaderThrowsCancelledException()
             }
         }
 
+
+
```
jkotalik: nit new lines.

src/Microsoft.AspNetCore.Server.IIS/Core/IO/AsyncIOOperation.cs
pakrym: In my defence - I copied it from System.Web.

PR #1366, opened 2018-09-04
Add script to revert changes to redirection.config

jkotalik:
Commonly when debugging full IIS tests, if you stop the debugger, changes to the redirection.config file in inetsrv/config will not be reverted. After that, if you try to build/run test again, it will fail because the aspnetcore[v2].dll will be locked by IIS. This adds a simple script to undo changes to the redirection.config.

### DIFF
--- a/tools/clean_redirection_config.ps1
+++ b/tools/clean_redirection_config.ps1
@@ -1,0 +1,38 @@
+<#
+.DESCRIPTION
+Sets redirection.config enabled flag to false. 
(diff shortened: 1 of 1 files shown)


pakrym:
Why not just ` Stop-Service w3svc`?

jkotalik:
You would need to run build/test separately as tests would fail due to w3svc being stopped.

pakrym:
Tests restart w3svc

Issue #779, opened 2018-04-16
Harden callbacks agains exception leaking from manged to native


pakrym:
https://github.com/aspnet/IISIntegration/pull/767 happened to fix a bug where throwing in create context would crash the server.
We need to check and wrap all other places in try-catch-log

muratg:
@pakrym have you started on this already?

PR #1369, opened 2018-09-05
Catch all exception in native callbacks

pakrym:
https://github.com/aspnet/IISIntegration/issues/779


src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpServer.cs
```diff
@@ -167,7 +176,7 @@ private static async Task HandleRequest(IISHttpContext context)
             }
             catch (Exception ex)
             {
-                context.Server._logger.LogError("Exception in ProcessRequestAsync", ex);
+                context.Server._logger.LogError(0, ex, $"Unexpected exception in {nameof(IISHttpServer)}.{nameof(HandleRequest)}.");
```
jkotalik: Anyway to force an Unexpected exception for a test?

Issue #1279, opened 2018-08-21
Attempt to kill IIS and IIS Express when stop debugging unit tests


mikeharder:
If you stop debugging a unit test, the IIS or IIS Express process will not be killed, which causes future builds and tests to fail.  We might be able to workaround this by attaching to `AppDomain.ProcessExit` or `AppDomain.DomainUnload`, and either killing all IIS / IIS Express processes, or keeping a static list of processes we need to kill.

Another option is to use Windows Job Objects: https://stackoverflow.com/questions/3342941/kill-child-process-when-parent-process-is-killed

PR #1371, opened 2018-09-05
Ensure external processes are killed when test process exits

mikeharder:
- Prevents locked files when stop debugging unit test
- Addresses #1279


src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployer.cs
```diff
@@ -254,6 +254,11 @@ private void WaitUntilSiteStarted()
                 }
 
                 HostProcess = Process.GetProcessById(workerProcess.ProcessId);
+
+                // Ensure w3wp.exe is killed if test process termination is non-graceful.
+                // Prevents locked files when stop debugging unit test.
+                ProcessTracker.Add(HostProcess);
```
jkotalik: How can this work for w3wp? The worker process would be immediately recycled and the dll would be locked again when killed.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/ProcessTracker.cs
```diff
@@ -0,0 +1,125 @@
+using System;
```
jkotalik: nit copyright.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/ProcessTracker.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Runtime.InteropServices;
+
+namespace Microsoft.AspNetCore.Server.IntegrationTesting.IIS
+{
+    // Uses Windows Job Objects to ensure external processes are killed if the current process is terminated non-gracefully.
+    internal static class ProcessTracker
+    {
+        private static readonly IntPtr _jobHandle;
+
+        static ProcessTracker()
```
jkotalik: Where did you get this class?

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployer.cs
mikeharder: When the test process exits, `w3wp.exe` (and the `dotnet.exe` child process of `w3wp.exe`) are killed.  It's true that `w3wp.exe` is immediately restarted, but the `dotnet.exe` child process will not.  So, while some dlls may still be locked, it seems better than doing nothing.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/ProcessTracker.cs
mikeharder: Slightly modified version of https://stackoverflow.com/a/37034966/102052

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeployer.cs
jkotalik: The file that is locked is always aspnetcore[v2].dll, so this really doesn't help at all.

mikeharder: I would argue its still slightly better to kill `dotnet.exe` than do nothing.  However, I'm also fine limiting this change to IIS Express, if we think it's misleading or noisy to try killing `w3wp.exe`.

jkotalik: Anyway to register a callback? Doesn't look like it, but maybe we can add a callback to undo the redirection.config changes on exit.

mikeharder: I believe the best we can do with Windows Job Objects is to kill processes.  If we wanted to undo the `redirection.config` changes (or anything else), we could create a watchdog process which detects when the test process terminates (e.g. via anonymous pipes) and then performs cleanup.

Issue #1213, opened 2018-08-11
Add EventSource/EventCounter tracing and metrics for IISIntegration


shirhatti:
Kestrel emits EventSource events for
- ConnectionStart
- ConnectionStop
- RequestStart
- RequestStop

https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/src/Kestrel.Core/Internal/Infrastructure/KestrelEventSource.cs

We should add event to our server implementation to cover these cases

Tratcher:
You can't do connection start, http.sys doesn't surface that.

muratg:
We'll try to do this in 2.2.0-preview3.

cc @shirhatti

pakrym:
Closing as we decided that having hosting events is enough for now.

PR #1374, opened 2018-09-06
Add event source request start and stop

jkotalik:
https://github.com/aspnet/IISIntegration/issues/1213

Only does request event sources for now.


test/IIS.Tests/EventSourceTests.cs
```diff
@@ -0,0 +1,108 @@
...
+                Assert.Equal(helloWorld, await result.Content.ReadAsStringAsync());
+                Assert.Equal(expectedPath, path);
+            }
+
+            // capture list here as other tests executing in parallel may log events
+            Assert.NotNull(requestId);
+
+            var events = _listener.EventData.Where(e => e != null && GetProperty(e, "requestId") == requestId).ToList();
+            {
+                var requestStart = Assert.Single(events, e => e.EventName == "RequestStart");
```
pakrym: Extract a method.

test/IIS.Tests/EventSourceTests.cs
```diff
@@ -0,0 +1,108 @@
...
+                requestId = ctx.TraceIdentifier;
+                return ctx.Response.WriteAsync(helloWorld);
+            }, LoggerFactory))
+            {
+                var result = await testServer.HttpClient.GetAsync(expectedPath);
+                Assert.Equal(helloWorld, await result.Content.ReadAsStringAsync());
+                Assert.Equal(expectedPath, path);
+            }
+
+            // capture list here as other tests executing in parallel may log events
```
pakrym: Is this comment applicable?

test/IIS.Tests/EventSourceTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        public async Task EmitsRequestStartAndStop()
+        {
+            var helloWorld = "Hello World";
+            var expectedPath = "/Path";
+            string requestId = null;
+            string path = null;
+
+            using (var testServer = await TestServer.Create(ctx =>
+            {
+                path = ctx.Request.Path.ToString();
```
pakrym: Delete/

test/IIS.Tests/TestServerTest.cs
```diff
@@ -19,17 +19,17 @@ public async Task SingleProcessTestServer_HelloWorld()
...
-                {
-                    path = ctx.Request.Path.ToString();
-                    return ctx.Response.WriteAsync(helloWorld);
-                }, LoggerFactory))
-                {
-                    var result = await testServer.HttpClient.GetAsync(expectedPath);
-                    Assert.Equal(helloWorld, await result.Content.ReadAsStringAsync());
-                    Assert.Equal(expectedPath, path);
-                }
+            string path = null;
```
pakrym: Undo, I fixed this in other PR.

pakrym:
Crossplat tests failed.

src/Microsoft.AspNetCore.Server.IIS/Core/IISEventSource.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+        }
+
+        [NonEvent]
+        public void RequestStart(IISHttpContext httpContext)
+        {
+            // avoid allocating the trace identifier unless logging is enabled
+            if (IsEnabled())
+            {
+                RequestStart(httpContext.ConnectionId.ToString(), httpContext.TraceIdentifier);
```
pakrym: We can log much more useful information here, path, query, app name etc. Why only log these two?

jkotalik: @shirhatti any reason kestrel only logged connection Id and request Id?

src/Microsoft.AspNetCore.Server.IIS/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
mikeharder: I don't like `InternalsVisibleTo`, is it really necessary here?

jkotalik: Right now, the [IISHttpContext is internal](https://github.com/aspnet/IISIntegration/blob/release/2.2/src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs#L25). We would need to expose a lot to remove the InternalsVisibleTo.

pakrym: Why do you need  IISHttpContext in test?

jkotalik: The tests needs IISEventSource, which uses IISHttpContext https://github.com/aspnet/IISIntegration/pull/1374/files#diff-9e79b380478e1d8cf32c8b5944147c9eR19

pakrym: I think it's fine to use IVT here considering it's exactly what we do in kestrel https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/src/Kestrel.Core/Properties/AssemblyInfo.cs

src/Microsoft.AspNetCore.Server.IIS/Core/IISEventSource.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(3, Level = EventLevel.Verbose)]
+        private void RequestStart(string connectionId, string requestId)
+        {
+            WriteEvent(3, connectionId, requestId);
+        }
+
+        [NonEvent]
+        public void RequestStop(IISHttpContext httpContext)
```
anurse: You could consider updating some EventCounters here as well so that users can get the low-granularity data too. See [SignalR's counters](https://github.com/aspnet/SignalR/blob/release/2.2/src/Microsoft.AspNetCore.Http.Connections/Internal/HttpConnectionsEventSource.cs) for an example that's low-allocating and relatively performance-friendly.

src/Microsoft.AspNetCore.Server.IIS/Core/IISEventSource.cs
anurse: I think that it's reasonable to keep this somewhat limited to the things only the server knows since Hosting *also* has a RequestStart event, but I'm also fine logging everything you've got so that people don't have to grab both events to get the full set of data.

jkotalik:
@shirhatti @pakrym are we still wanting to merge this? I've heard mixed things all day about this 😅

shirhatti:
Nope. Kill it

Issue #692, opened 2018-03-17
ANCM: Globalization: Log file should be saved in unicode encoding


jhkimnew:
Currently the log file is saved in ANSI, not Unicode encoding. Because of the problem, the Unicode characters which is used as Uri path can't be logged correctly.

STEPS TO REPRO:
1. Create a new WebApplication and deploy aspnetcore app in inprocess
2. Put 𠀀𠀁𠀃𠀄𠀅.htm file on the wwwroot directory of the app
3. Enable stdoutLogEnabled true and give the full permission to the worker process
4. Send a request to the 𠀀𠀁𠀃𠀄𠀅.htm
5. Open the created log file

EXPECTED:

The log file shows the correct URI path for the request

ACTAUL

The file name is broken with ??????????.htm as the following example:

```
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost/aspnetcoreapp/??????????.htm  
info: Microsoft.AspNetCore.StaticFiles.StaticFileMiddleware[2]
      Sending file. Request path: '/??????????.htm'. Physical path: 'C:\inetpub\wwwroot\aspnetcoreapp\wwwroot\??????????.htm'
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 30.3956ms 200 text/html
```

pakrym:
@jkotalik  can you confirm we have tests for this both in in-proc and out-of-proc and close the issue?

jkotalik:
Won't fix, will add out of proc stdout logging tests in another PR.

PR #1382, opened 2018-09-07
Add test for log file with unicode

jkotalik:
For #692


jkotalik:
Having `/??????????.htm` in the std output is expected as the encoding is not set to unicode/utf8. Decided to add a test for the log file being in unicode.

pakrym:
This is not what bug was talking about

pakrym:
Change the existing test to use unicode name. Not much point in having both.

jkotalik:
Closing as redundant/won't fix.

Issue #1381, opened 2018-09-07
Add TlsConnectionFeature for client certs


shirhatti:
https://github.com/aspnet/HttpAbstractions/blob/master/src/Microsoft.AspNetCore.Http.Features/ITlsConnectionFeature.cs

PR #1390, opened 2018-09-12
Implement ITlsConnectionFeature

jkotalik:
For #1381


src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
```diff
@@ -165,6 +158,30 @@ protected void InitializeContext()
...
+
+            if (_options.ForwardClientCertificate)
+            {
+                // Only set the ITlsConnectionFeature if there is a cert available.
+                if (NativeRequest->pSslInfo != null &&
+                    NativeRequest->pSslInfo->pClientCertInfo != null &&
+                    NativeRequest->pSslInfo->pClientCertInfo->pCertEncoded != null &&
+                    NativeRequest->pSslInfo->pClientCertInfo->CertEncodedSize != 0)
+                {
+                    _currentITlsConnectionFeature = new IISTlsConnectionFeature(
```
pakrym: Why change the pattern and not implement the feature on context?

test/IISExpress.FunctionalTests/HttpsTests.cs
```diff
@@ -10,28 +10,25 @@
...
 
 namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
 {
-    // IIS Express preregisteres 44300-44399 ports with SSL bindings.
-    // So these tests always have to use ports in this range, and we can't rely on OS-allocated ports without a whole lot of ceremony around
-    // creating self-signed certificates and registering SSL bindings with HTTP.sys
-    // Test specific to IISExpress
     [Collection(PublishedSitesCollection.Name)]
-    public class HttpsTest : IISFunctionalTestBase
+    public class HttpsTests : IISFunctionalTestBase
```
pakrym: Why aren't we skipping tests anymore? What changed?

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
jkotalik: We need to check that the cert exists before adding the feature. This matches the out of proc behavior.

pakrym: We do this for websockets and null-out the feature if it's disabled.

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
```diff
@@ -276,6 +279,17 @@ async Task<Stream> IHttpUpgradeFeature.UpgradeAsync()
...
         }
 
+        Task<X509Certificate2> ITlsConnectionFeature.GetClientCertificateAsync(CancellationToken cancellationToken)
+        {
+            return Task.FromResult(ClientCertificate);
+        }
+
+        unsafe X509Certificate2 ITlsConnectionFeature.ClientCertificate
+        {
+            get => ClientCertificate;
```
pakrym: Why have this and separate ClientCertificate property?

test/IISExpress.FunctionalTests/HttpsTests.cs
pakrym: hm?

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
```diff
@@ -191,6 +197,38 @@ public string ReasonPhrase
...
+            get
+            {
+                if (_options.ForwardClientCertificate &&
+                    _certificate == null &&
+                    NativeRequest->pSslInfo != null &&
+                    NativeRequest->pSslInfo->pClientCertInfo != null &&
+                    NativeRequest->pSslInfo->pClientCertInfo->pCertEncoded != null &&
+                    NativeRequest->pSslInfo->pClientCertInfo->CertEncodedSize != 0)
+                {
+                    try
```
pakrym: Just let it throw

test/IISExpress.FunctionalTests/HttpsTests.cs
jkotalik: We are still skipping client cert tests. HTTPS tests aren't skipped and can be run inproc.

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
jkotalik: Pattern that all other features follow, like RequestHeaders, Request/Response Body, etc.: https://github.com/aspnet/IISIntegration/blob/release/2.2/src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs#L105-L109. Main reason is to prevent app code from overwriting critical features.

pakrym: I don't like it in other places too. How does it prevent overwriting?

jkotalik: If someone does ` context.Features.Get<IHttpRequestFeature>.Body` and replaces the body, bad things will happen with the request/repsonse. 

cc/ @Tratcher as this was copied from HttpSys.

pakrym: But that's the reason it's writable and can see scenarios where ClientCertificate can be replaced.

Tratcher: There are some fields where the server needs to keep a reference to the original value like headers or body, and those must be preserved in and referenced from a private field. ClientCertificate is not a field that the server has any need for (unless it needs to dispose it?).

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
```diff
@@ -165,6 +161,16 @@ protected void InitializeContext()
             _bodyOutput = new OutputProducer(pipe);
         }
 
+        private unsafe void ModifyFeatureCollection()
```
Tratcher: Modify? Initialize?

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
Tratcher: No, Getters shouldn't throw.

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
```diff
@@ -191,6 +197,38 @@ public string ReasonPhrase
             }
         }
 
+        public unsafe X509Certificate2 ClientCertificate
+        {
+            get
+            {
+                if (_options.ForwardClientCertificate &&
```
Tratcher: I think we can remove ForwardClientCertificate since ClientCertificate itself is lazy.

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
```diff
@@ -191,6 +197,38 @@ public string ReasonPhrase
...
+                    NativeRequest->pSslInfo->pClientCertInfo != null &&
+                    NativeRequest->pSslInfo->pClientCertInfo->pCertEncoded != null &&
+                    NativeRequest->pSslInfo->pClientCertInfo->CertEncodedSize != 0)
+                {
+                    try
+                    {
+                        // Based off of from https://referencesource.microsoft.com/#system/net/system/Net/HttpListenerRequest.cs
+                        var rawCertificateCopy = new byte[NativeRequest->pSslInfo->pClientCertInfo->CertEncodedSize];
+                        Marshal.Copy((IntPtr)NativeRequest->pSslInfo->pClientCertInfo->pCertEncoded, rawCertificateCopy, 0, rawCertificateCopy.Length);
+                        _certificate = new X509Certificate2(rawCertificateCopy);
```
Tratcher: Does this need to be disposed at the end of the request?

test/IISExpress.FunctionalTests/HttpsTests.cs
```diff
@@ -10,28 +10,25 @@
...
+        public HttpsTests(PublishedSitesFixture fixture)
         {
             _fixture = fixture;
         }
 
         public static TestMatrix TestVariants
             => TestMatrix.ForServers(DeployerSelector.ServerType)
-                .WithTfms(Tfm.NetCoreApp22, Tfm.Net461)
-                .WithAllAncmVersions();
+                .WithTfms(Tfm.NetCoreApp22)
```
Tratcher: Why no Net461 for out of proc? The TestMatrix should skip invalid combinations.

test/WebSites/InProcessWebSite/Startup.cs
```diff
@@ -644,5 +644,15 @@ private async Task CommandLineArgs(HttpContext ctx)
         {
             await ctx.Response.WriteAsync(string.Join("|", Environment.GetCommandLineArgs().Skip(1)));
         }
+
+        private async Task GetClientCert(HttpContext context)
+        {
+            var clientCert = context.Features.Get<ITlsConnectionFeature>();
+
+            await context.Response.WriteAsync(clientCert?.ClientCertificate != null ? "Enabled" : "Disabled");
```
Tratcher: HttpContext.Connection.ClientCertificate?

src/Microsoft.AspNetCore.Server.IIS/IISServerOptions.cs
```diff
@@ -22,5 +22,10 @@ public class IISServerOptions
         /// IIS has a non-anonymous authentication enabled, or for back compat with ANCMs that did not provide this information.
         /// </summary>
         internal bool ForwardWindowsAuthentication { get; set; } = true;
+
+        /// <summary>
+        /// Populates the ITLSConnectionFeature
```
davidfowl: This comment isn't very good.

src/Microsoft.AspNetCore.Server.IIS/IISServerOptions.cs
```diff
@@ -22,5 +22,10 @@ public class IISServerOptions
...
+        /// </summary>
+        public bool ForwardClientCertificate { get; set; } = true;
```
davidfowl: Why do we need this?

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
davidfowl: Getters can totally throw. That's not against the rules. What sort of errors can this throw? It's kinda nasty catching all the exceptions.

We also have the async call to get the cert and that should throw, but the async call uses the sync API then it would also swallow the error no?

src/Microsoft.AspNetCore.Server.IIS/IISServerOptions.cs
Tratcher: We don't, it was copied from out-of-proc. Remove it.

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
Tratcher: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property
"X AVOID throwing exceptions from property getters.

Property getters should be simple operations and should not have any preconditions. If a getter can throw an exception, it should probably be redesigned to be a method. Notice that this rule does not apply to indexers, where we do expect exceptions as a result of validating the arguments."

Tratcher: Also, only HttpSys has a meaningful implementation of the async API, it can dynamically renegotiate the connection to request a cert on demand. Kestrel and IIS both require the cert up front.

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
jkotalik: Doesn't look like we need to dispose it.

jkotalik:
Don't want to spend any more time getting automated client certificates working. Tests work locally when provided a valid cert. Will create a work item for it.

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
```diff
@@ -290,6 +294,42 @@ async Task<Stream> IHttpUpgradeFeature.UpgradeAsync()
...
+                    NativeRequest->pSslInfo->pClientCertInfo->CertEncodedSize != 0)
+                {
+                    try
+                    {
+                        // Based off of from https://referencesource.microsoft.com/#system/net/System/Net/HttpListenerRequest.cs,1037c8ec82879ba0,references
+                        var rawCertificateCopy = new byte[NativeRequest->pSslInfo->pClientCertInfo->CertEncodedSize];
+                        Marshal.Copy((IntPtr)NativeRequest->pSslInfo->pClientCertInfo->pCertEncoded, rawCertificateCopy, 0, rawCertificateCopy.Length);
+                        _certificate = new X509Certificate2(rawCertificateCopy);
+                    }
+                    catch(Exception)
```
davidfowl: Why?

Tratcher: https://github.com/aspnet/IISIntegration/pull/1390#discussion_r217917718

Properties are expected to act like simple public fields. Putting try/catch around every property would be a nuisance.

pakrym: You don't put try/catch around every method call.

pakrym: It also throws in HttpSys: https://github.com/aspnet/HttpSysServer/blob/eba2d7e380fef9f75203bb9a1f4d30827451d512/src/Microsoft.AspNetCore.Server.HttpSys/RequestProcessing/Request.cs#L269

Tratcher: You put try/catch around the ones that say they can.
IISIntegration catches client cert errors. https://github.com/aspnet/IISIntegration/blob/5e896ca50678547ce5a71a23edd7dcc5bab45ba7/src/Microsoft.AspNetCore.Server.IISIntegration/ForwardedTlsConnectionFeature.cs#L32-L40 That's a better precident for you here.

jkotalik: Will remove try/catch. Creating a new x509 cert will probably not fail as it already passed through IIS cert validation.

test/IISExpress.FunctionalTests/HttpsTests.cs
```diff
@@ -117,39 +120,76 @@ private async Task HttpsHelloWorldCerts(TestVariant variant, bool sendClientCert
...
+                new X509SubjectKeyIdentifierExtension(parentRequest.PublicKey, false));
+
+            var notBefore = DateTimeOffset.Now.AddDays(-1);
+            var notAfter = DateTimeOffset.Now.AddYears(5);
+
+            var parentCert = parentRequest.CreateSelfSigned(notBefore, notAfter);
+
+            // Need to export/import the certificate to associate the private key with the cert.
+            var imported = parentCert;
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
Tratcher: You don't test on OSX. Where did you copy this from?

test/IISExpress.FunctionalTests/HttpsTests.cs
```diff
@@ -117,39 +120,76 @@ private async Task HttpsHelloWorldCerts(TestVariant variant, bool sendClientCert
...
+            {
+                // On non OSX systems we need to export the certificate and import it so that the transient
+                // key that we generated gets persisted.
+                var export = parentCert.Export(X509ContentType.Pkcs12, "");
+                imported = new X509Certificate2(export, "", X509KeyStorageFlags.PersistKeySet | X509KeyStorageFlags.Exportable);
+                Array.Clear(export, 0, export.Length);
+            }
+
+            // Add the cert to the cert store
+            using (var store = new X509Store(StoreName.Root, StoreLocation.LocalMachine))
```
Tratcher: Admin only?

This kind of state change shouldn't be made on dev machines.

Tratcher: We might be better off pre-provisioning a test root cert on the CI agents.

jkotalik: Added a check for running as admin.

Tratcher: Adding a trusted cert is not something that should be automated if a 3rd party runs build.cmd on this repo. @blowdart

test/IISExpress.FunctionalTests/RequiresIISAttribute.cs
```diff
@@ -10,15 +11,24 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
...
 
-        public string SkipReason => IISExpressAncmSchema.SkipReason;
+        public string SkipReason { get; set; } = IISExpressAncmSchema.SkipReason;
 
         public RequiresIISAttribute() { }
 
         public RequiresIISAttribute(IISCapability capabilities)
         {
-            // IISCapabilities aren't pretinent to IISExpress
+            if (capabilities.HasFlag(IISCapability.Admin))
```
pakrym: It's not an IIS capability.

jkotalik: I would have to create a new enum otherwise.

test/WebSites/OutOfProcessWebSite/Startup.cs
```diff
@@ -105,5 +101,12 @@ private async Task ServerAddresses(HttpContext context)
         {
             await context.Response.WriteAsync(string.Join(",", _serverAddresses.Addresses));
         }
+
+        private async Task GetClientCert(HttpContext context)
+        {
+            var clientCert = context.Features.Get<ITlsConnectionFeature>();
+
+            await context.Response.WriteAsync(clientCert?.ClientCertificate != null ? "Enabled" : "Disabled");
```
pakrym: Return some property of the cert and assert it (issuer, subject, etc.).

test/IISExpress.FunctionalTests/HttpsTests.cs
```diff
@@ -78,69 +88,105 @@ private async Task HttpsHelloWorldCerts(TestVariant variant, bool sendClientCert
...
-                Assert.Equal("Scheme:https; Original:http; has cert? False", responseText);
+                Assert.Equal("Disabled", responseText);
+            }
+        }
+
+        private X509Certificate2 CreateRootCert()
+        {
+            var parentKey = CreateKeyMaterial(2048);
+
+            // On first run of the test, creates the certificate in the trusted root certificate authorties.
```
pakrym: Are we reusing the cert? Who is going to remove it?

test/IISExpress.FunctionalTests/HttpsTests.cs
jkotalik: We can add a check for running on only VSTS, jenkins, and appveyor.

test/IISExpress.FunctionalTests/HttpsTests.cs
```diff
@@ -78,69 +88,105 @@ private async Task HttpsHelloWorldCerts(TestVariant variant, bool sendClientCert
...
             {
                 ServerCertificateCustomValidationCallback = (a, b, c, d) => true,
-                ClientCertificateOptions = ClientCertificateOption.Manual
+                ClientCertificateOptions = ClientCertificateOption.Manual,
             };
 
             if (sendClientCert)
             {
-                X509Certificate2 clientCert = FindClientCert();
+                var clientCert = FindClientCert();
```
pakrym: Put this check into `FindClientCert`

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
```diff
@@ -144,17 +144,11 @@ protected void InitializeContext()
                 }
             }
 
-            ResetFeatureCollection();
-
-            if (!_server.IsWebSocketAvailable(_pInProcessHandler))
-            {
-                _currentIHttpUpgradeFeature = null;
-            }
+            InitializeFeatureCollection();
```
pakrym: Revert this file

test/IISExpress.FunctionalTests/RequiresIISAttribute.cs
pakrym: You can create a new attribute.

test/IISExpress.FunctionalTests/HttpsTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                    critical: true));
+
+            parentRequest.CertificateExtensions.Add(
+                new X509KeyUsageExtension(X509KeyUsageFlags.DigitalSignature | X509KeyUsageFlags.NonRepudiation, critical: true));
+
+            parentRequest.CertificateExtensions.Add(
+                new X509SubjectKeyIdentifierExtension(parentRequest.PublicKey, false));
+
+            var notBefore = DateTimeOffset.Now.AddDays(-1);
+            var notAfter = DateTimeOffset.Now.AddYears(5);
```
pakrym: `AddDays(1)` ?

jkotalik: Why?

jkotalik:
🆙 📅

test/IISExpress.FunctionalTests/HttpsTests.cs
Tratcher: Because you expect the cert to be cleaned up at the end of the test.

test/IISExpress.FunctionalTests/HttpsTests.cs
blowdart: A fixed test root would be better. Spin up a Win Server CA and issue the certs from there, import the root CA via powershell (this is what I do for the cert auth). This could also exercise CRL checking.

Issue #1386, opened 2018-09-10
Implement IHttpBufferingFeature


Tratcher:
We need to evaluate if DisableRequestBuffering and DisableResponseBuffering are relevant to ANCM in-proc. 

IIS response compression may be one scenario, but that may have been addressed for the signalr scenario by the installer change to remove text/event-stream. @anurse

anurse:
We still use the `Content-Encoding: identity` hack so I don't know for sure that it's working now with that web.config change.

PR #1391, opened 2018-09-12
Implement IHttpBufferingFeature

pakrym:
https://github.com/aspnet/IISIntegration/issues/1386


src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
```diff
@@ -284,5 +298,16 @@ void IHttpRequestLifetimeFeature.Abort()
...
+
+        public void DisableRequestBuffering()
+        {
+        }
+
+        public void DisableResponseBuffering()
+        {
+            NativeMethods.HttpDisableBuffering(_pInProcessHandler);
+            var serverVariableFeature = (IServerVariablesFeature)this;
+            serverVariableFeature["IIS_EnableDynamicCompression"] = "0";
```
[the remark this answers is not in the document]
pakrym: Disables dynamic compression.

pakrym:
> Are responses buffered by default?

Yes

src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.FeatureCollection.cs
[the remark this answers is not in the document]
pakrym: https://referencesource.microsoft.com/#System.Web/HttpRequest.cs,7594c3f7d09f40ee,references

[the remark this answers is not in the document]
jkotalik: Based on earlier discussion, make the method declaration: `void IHttpBufferingFeature.DisableRequestBuffering()`

src/Microsoft.AspNetCore.Server.IIS/HttpContextExtensions.cs
```diff
@@ -34,5 +34,17 @@ public static string GetIISServerVariable(this HttpContext context, string varia
...
+
+        /// <summary>
+        /// Sets the value of a server variable for the current request.
+        /// </summary>
+        /// <param name="context">The http context for the request.</param>
+        /// <param name="variableName">The name of the variable.</param>
+        /// <param name="variableValue">The new value for the variable.</param>
+        public static void SetIISServerVariable(this HttpContext context, string variableName, string variableValue)
+        {
+            var feature = context.Features.Get<IServerVariablesFeature>();
```
Tratcher: Null check and throw InvalidOperationException, the feature could have been removed.

pakrym: I was thinking about removing these methods all together.

Tratcher: Get is common enough, but I agree Set is uncommon.

pakrym: What about removing both and changing IServerVariableFeature to have methods instead of indexer?

Tratcher: What's wrong with the indexer?

pakrym: It's a bit strange seeing an interface with only member being string indexer.

Tratcher: *shrug* it fits the requirements.

src/Microsoft.AspNetCore.Server.IIS/HttpContextExtensions.cs
```diff
@@ -1,50 +0,0 @@
...
-        /// <param name="context">The http context for the request.</param>
-        /// <param name="variableName">The name of the variable.</param>
-        /// <returns>
-        /// <c>null</c> if the feature does not support the <see cref="IServerVariablesFeature"/> feature.
-        /// May return null or empty if the variable does not exist or is not set.
-        /// </returns>
-        /// <remarks>
-        /// For a list of common server variables available in IIS, see http://go.microsoft.com/fwlink/?LinkId=52471.
-        /// </remarks>
-        public static string GetIISServerVariable(this HttpContext context, string variableName)
```
Tratcher: @shirhatti are you cool with this? I was under the impression you wanted these to be more visible.

davidfowl: I’d hate for this to show up by default in intellisense. These are referenced by default.

Tratcher: That's why the namespace was `Microsoft.AspNetCore.Server.IIS`, that won't show up in Startup.

davidfowl: Then that’s fine imo, no need to delete.  Though it kinda forces user code to check for null.

pakrym: Why have extra API that almost no-one uses? How is it different from other feature we don't write extension methods for?

Tratcher: We went over this when we added the method the first time. Nothing has changed, I don't see any reason to rehash the discussion now.

src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/IISDeploymentParameterExtensions.cs
```diff
@@ -80,5 +80,31 @@ public static void TransformArguments(this IISDeploymentParameters parameters, F
...
+                        .RequiredElement("globalModules")
+                        .GetOrAdd("add", "name", moduleName)
+                        .SetAttributeValue("image", modulePath);
+
+                    (webServerElement.Element("modules") ??
+                     element
+                         .Element("location")
+                         .RequiredElement("system.webServer")
+                         .RequiredElement("modules"))
+                        .GetOrAdd("add", "name", moduleName);
```
jkotalik: nit space.

test/Common.FunctionalTests/Inprocess/CompressionTests.cs
```diff
@@ -0,0 +1,96 @@
...
+        [RequiresIIS(IISCapability.DynamicCompression)]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task BufferingDisabled(bool compression)
+        {
+            using (var connection = _fixture.CreateTestConnection())
+            {
+                var requestLength = 0;
+                var messages = new List<string>();
+                for (var i = 1; i < 100; i++)
```
jkotalik: nit why not start at i = 0?

test/IISExpress.FunctionalTests/InProcess/WebSocketTests.cs
```diff
@@ -20,7 +20,7 @@ public class WebSocketsTests
 
         public WebSocketsTests(IISTestSiteFixture fixture)
         {
-            _webSocketUri = fixture.BaseUri.Replace("http:", "ws:");
+            _webSocketUri = fixture.DeploymentResult.ApplicationBaseUri.Replace("http:", "ws:");
```
jkotalik: Seems like a random fix.

pakrym: I removed extra properties from fixture to make it lazy.

Issue #933, opened 2018-06-18
Process crash in hostpolicy.dll when throwing exception in main.


jkotalik:
Repro steps:
1. Run an inprocess application.
2. In program main, add `throw new Exception();`
3. Run the application.
4. Send http request, receive response, but w3wp crashes in hostpolicy with the following stack trace.

```c++
 	hostpolicy.dll!std::basic_string<wchar_t,struct std::char_traits<wchar_t>,class std::allocator<wchar_t> >::assign(class std::basic_string<wchar_t,struct std::char_traits<wchar_t>,class std::allocator<wchar_t> > const &,unsigned __int64,unsigned __int64)	C++
 	hostpolicy.dll!breadcrumb_writer_t::write_callback(void)	C++
 	hostpolicy.dll!breadcrumb_writer_t::write_worker_callback(class breadcrumb_writer_t *)	C++
 	hostpolicy.dll!std::_LaunchPad<class std::unique_ptr<class std::tuple<void (*)(class breadcrumb_writer_t *),class breadcrumb_writer_t *>,struct std::default_delete<class std::tuple<void (*)(class breadcrumb_writer_t *),class breadcrumb_writer_t *> > > >::_Run(class std::_LaunchPad<class std::unique_ptr<class std::tuple<void (*)(class breadcrumb_writer_t *),class breadcrumb_writer_t *>,struct std::default_delete<class std::tuple<void (*)(class breadcrumb_writer_t *),class breadcrumb_writer_t *> > > > *)	C++
>	hostpolicy.dll!std::_Pad::_Call_func(void *)	C++
```

pakrym:
Filed https://github.com/dotnet/core-setup/issues/4315

jkotalik:
No work for us to do on our side. Let's keep this issue open for tracking.

pakrym:
Not crashing anymore but still causes AppVerifier assert.

pakrym:
Tracking https://github.com/dotnet/core-setup/issues/4646

PR #1395, opened 2018-09-13
Skip AppOfflineDroppedWhi… test when AppVerif is running

pakrym:
https://github.com/aspnet/IISIntegration/issues/933


pakrym:
It's been failing on aspnetci too, I'll skip it completely.

Issue #1393, opened 2018-09-13
Test failure: Https tests all failing on test runs


jkotalik:
https://ci3.dot.net/job/aspnet_IISIntegration/job/release_2.2/job/windows-Configuration_Release_prtest/752/
```
Failed   Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.HttpsTest.HttpsHelloWorld_NoClientCert(variant: Server: IISExpress, TFM: netcoreapp2.2, Type: Portable, Arch: x64, ANCM: V2, Host: OutOfProcess)
18:29:05   Error Message:
18:29:05    System.Net.Http.HttpRequestException : The SSL connection could not be established, see inner exception.
18:29:05   ---- System.IO.IOException : Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.
18:29:05   -------- System.Net.Sockets.SocketException : An existing connection was forcibly closed by the remote host
18:29:05   Stack Trace:
18:29:05      at System.Net.Http.ConnectHelper.EstablishSslConnectionAsyncCore(Stream stream, SslClientAuthenticationOptions sslOptions, CancellationToken cancellationToken)
18:29:05      at System.Net.Http.HttpConnectionPool.CreateConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
18:29:05      at System.Net.Http.HttpConnectionPool.WaitForCreatedConnectionAsync(ValueTask`1 creationTask)
18:29:05      at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
18:29:05      at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
18:29:05      at Microsoft.AspNetCore.Server.IntegrationTesting.IIS.LoggingHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken) in /_/src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS/LoggingHandler.cs:line 29
18:29:05      at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
[...]

PR #1397, opened 2018-09-13
Update certs on Jenkins

pakrym:
https://github.com/aspnet/IISIntegration/issues/1393


pakrym:
IIS Express tests passed in both Jenkins runs.

pakrym:
Merging to unblock other PR.

Issue #1404, opened 2018-09-19
ANCM V2 InProcess - Server Header Issue


rockerinthelocker:
Instead of replacing the 'Server' header, the custom 'Server' header is just added when using hostingModel="InProcess".

```
    public class Startup
    {
        // Replacing the Server header does not work with InProcess.
        private const string serverHeader = "MyServerHeader";

        public void Configure(IApplicationBuilder app)
        {
            app.Use((context, next) =>
            {
                context.Response.Headers["Server"] = serverHeader;
                return next();
            });
        }
    }
```

Maybe it's the same for 'OutOfProcess' (not tested), but anyway, the header should be replaced.

PR #1409, opened 2018-09-19
Override IIS headers in inproc

pakrym:
https://github.com/aspnet/IISIntegration/issues/1404


src/Microsoft.AspNetCore.Server.IIS/Core/IISHttpContext.cs
```diff
@@ -323,12 +323,13 @@ public unsafe void SetResponseHeaders()
                     var headerNameBytes = Encoding.UTF8.GetBytes(headerPair.Key);
                     for (var i = 0; i < headerValues.Count; i++)
                     {
+                        var isFirst = i == 0;
```
jkotalik: nit: consider moving for loop outside of if check, like:
```c#
 for (var i = 0; i < headerValues.Count; i++)
 {
     if (knownHeaderIndex == -1)
     {
     }
     else
     {
     }
```

pakrym:
https://github.com/aspnet/IISIntegration/issues/1418

Issue #1418, opened 2018-09-20
Test failure: ReaderThrowsResetExceptionOnInvalidBody


pakrym:
```
Failed   Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReaderThrowsResetExceptionOnInvalidBody
2018-09-20T00:20:40.6059769Z   Error Message:
2018-09-20T00:20:40.6060265Z    System.TimeoutException : The operation at /_/test/IIS.Tests/ClientDisconnectTests.cs:238 timed out after reaching the limit of 300000ms.
2018-09-20T00:20:40.6061786Z   Stack Trace:
2018-09-20T00:20:40.6062110Z      at Microsoft.AspNetCore.Testing.TaskExtensions.TimeoutAfter[T](Task`1 task, TimeSpan timeout, String filePath, Int32 lineNumber)
2018-09-20T00:20:40.6062502Z      at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReaderThrowsResetExceptionOnInvalidBody() in /_/test/IIS.Tests/ClientDisconnectTests.cs:line 238
2018-09-20T00:20:40.6063800Z   --- End of stack trace from previous location where exception was thrown ---
2018-09-20T00:20:40.6064284Z   Standard Output Messages:
2018-09-20T00:20:40.6064521Z    | [0.001s] TestLifetime Information: Starting test ReaderThrowsResetExceptionOnInvalidBody at 2018-09-20T00:15:33
2018-09-20T00:20:40.6065150Z    | [4.814s] Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.TestServer Debug: Method: GET, RequestUri: 'http://localhost:1777/start', Version: 2.0, Content: <null>, Headers:
2018-09-20T00:20:40.6065678Z    |                                                                                       {
2018-09-20T00:20:40.6065996Z    |                                                                                       }
2018-09-20T00:20:40.6066260Z    | [4.820s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting starting
2018-09-20T00:20:40.6066591Z    | [4.821s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting started
[...]

PR #1420, opened 2018-09-20
Fix ReaderThrowsResetExceptionOnInvalidBody

pakrym:
https://github.com/aspnet/IISIntegration/issues/1418

If IIS body read goes async it would not abort until the client disconnects.


test/IIS.Tests/ClientDisconnectTests.cs
```diff
@@ -227,16 +229,25 @@ public async Task ReaderThrowsResetExceptionOnInvalidBody()
             {
                 using (var connection = testServer.CreateConnection())
                 {
-                    await connection.Send(
+                   await connection.Send(
```
jkotalik: nit space.

PR #1437, opened 2018-09-25
Set the current working directory to the app directory

jkotalik:
This can be considered a risky change, but from what I can tell this works.

### DIFF
--- a/src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
+++ b/src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
@@ -220,6 +220,10 @@ IN_PROCESS_APPLICATION::ExecuteApplication()
         // set the callbacks
         s_Application = this;
 
+        // Set the current working directory to the application path.
+        // This keeps the current directory behavior consistent with out of process
+        SetCurrentDirectory(QueryApplicationPhysicalPath().c_str());
+
         //Start CLR thread
         m_clrThread = std::thread(ClrThreadEntryPoint, context);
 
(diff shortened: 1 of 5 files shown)


src/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -220,6 +220,10 @@ IN_PROCESS_APPLICATION::ExecuteApplication()
         // set the callbacks
         s_Application = this;
 
+        // Set the current working directory to the application path.
+        // This keeps the current directory behavior consistent with out of process
+        SetCurrentDirectory(QueryApplicationPhysicalPath().c_str());
```
mikeharder: Should current directory be reset after CLR is started, to preserve compat for other IIS modules?

pakrym: I'm not sure if we should even take this change. We can't reset current directory after CLR started because we don't know which user code relies on it and it would even more confusing if it changes during runtime.

jkotalik: Also, the core clr calls into native to query the current directory. We can't set it after startup. It's the same issue that console redirection had.

pakrym:
What's the status of this? Did we decide if we want to do it and accept risks?

jkotalik:
Undecided.

pakrym:
Is there a tracking issue?

jkotalik:
https://github.com/aspnet/IISIntegration/issues/1469

muratg:
We have decided not to do this.

Issue #1435, opened 2018-09-25
Investigate w3wp failing to shutdown stress issues


jkotalik:
Many tests have been failing on IIS due to w3wp failing to shut down. Once one test fails to shut down w3wp, all other tests will fail saying the server manager cannot accept command messages at this time.

For example https://ci.appveyor.com/project/aspnetci/iisintegration/build/1.0.2514?fullLog=true
```
Failed   Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.CommonStartupTests.StartupStress(variant: Server: IIS, TFM: netcoreapp2.2, Type: Portable, Arch: x64, ANCM: V2, Host: InProcess)
  Error Message:
[...]

Issue #1428, opened 2018-09-21
Test failure: OutOfProcessToInProcessHostingModelSwitchWorks


aspnet-hello:
This test [fails](http://aspnetci/viewLog.html?buildId=558508&buildTypeId=Lite_UniverseTest) occasionally with the following error:
```
System.InvalidOperationException : Didn't get response that satisfies predicate after 5 retries
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.Helpers.RetryRequestAsync(HttpClient client, String uri, Func`2 predicate) in /_/test/Common.FunctionalTests/Utilities/Helpers.cs:line 112
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ConfigurationChangeTests.OutOfProcessToInProcessHostingModelSwitchWorks() in /_/test/Common.FunctionalTests/ConfigurationChangeTests.cs:line 85
--- End of stack trace from previous location where exception was thrown ---
------- Stdout: -------
| [0.001s] TestLifetime Information: Starting test OutOfProcessToInProcessHostingModelSwitchWorks at 2018-09-21T19:27:07
| [0.003s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Information: Deploying [Variation] :: ServerType=IIS, Runtime=CoreClr, Arch=x64, BaseUrlHint=, Publish=True
| [0.004s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Debug: Processing C:\b\t\buildTmp\67235146143944c0af6821fe081bd5fa\runtimes\unix\lib\netcoreapp2.0
| [0.005s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Debug:   Copying System.Security.Principal.Windows.dll
| [0.006s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Debug: Processing C:\b\t\buildTmp\67235146143944c0af6821fe081bd5fa\runtimes\unix\lib
| [0.006s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Debug: Processing C:\b\t\buildTmp\67235146143944c0af6821fe081bd5fa\runtimes\unix
[...]

PR #1449, opened 2018-09-27
Increase retry request count.

jkotalik:
For https://github.com/aspnet/IISIntegration/issues/1428.

From what I can tell, the server didn't shut down with 500ms. This could be related to #1435, but this should allow us to diagnose whether this is due shutdown being slow or an actual product issue.


jkotalik:
Test failure: https://github.com/aspnet/IISIntegration/issues/1450

PR #1452, opened 2018-09-28
Stop calling StopCallsIntoManaged after first call

jkotalik:
[There were two test failures on shutdown on this PR. ](https://dnceng.visualstudio.com/public/_build/results?buildId=25048&view=ms.vss-test-web.test-result-details) One of them seems to be when setting `pInProcessApplication->StopCallsIntoManaged();` after managed has shut down. The actual exception seems to be in std::atomic_bool, but my guess is the native application is disposed and the IISNativeApplication finalizer is being called: https://github.com/aspnet/IISIntegration/blob/release/2.2/src/Microsoft.AspNetCore.Server.IIS/Core/IISNativeApplication.cs#L54. I otherwise can't tell what the error is.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IIS/Core/IISNativeApplication.cs
+++ b/src/Microsoft.AspNetCore.Server.IIS/Core/IISNativeApplication.cs
@@ -8,7 +8,7 @@ namespace Microsoft.AspNetCore.Server.IIS.Core
 {
     internal class IISNativeApplication
     {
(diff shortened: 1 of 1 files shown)


pakrym:
Native application should wait for CLR to finish, the only time it may be deleted before CLR exits is in case of CLR shutdown timeout which I don't think is the case for the tests that failed.

If application was deleted before CLR exits the fix should be somewhere in native.

PR #1453, opened 2018-09-28
Move Log call before starting application

jkotalik:
Couldn't think of a better way to avoid this race. For https://github.com/aspnet/IISIntegration/issues/1440


pakrym:
Move std out redirection initialization to clr thread?

pakrym:
On the other hand it's not worth doing just for debug logs. LGTM

PR #1457, opened 2018-09-28
Fix mof file

jkotalik:
The mof file was invalid for ANCM (there was a top level comment that caused a parsing issue. The entire function should be commented out (see: https://github.com/aspnet/IISIntegration/compare/jkotalik/badMof?expand=1#diff-cac310b2dee11a7804cee343275a1582L56)


test/IIS.Tests/MofFileTests.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
```
pan-wang: this will only ensure the mof file is well formatted but does not guarantee the content is correctly. Ideally, we should have some e2e test using logon/ETW API to retrieve the event.

jkotalik: Follow up PR. The events to day are not great, so I'll add some more and then add E2E tests for them.

test/IIS.FunctionalTests/MofFileTests.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public class MofFileTests
+    {
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
+        [RequiresIIS(IISCapability.TracingModule)]
+        public void CheckMofFile()
+        {
+            var path = Path.Combine(TestPathUtilities.GetSolutionRootDirectory("IISIntegration"), "src", "aspnetcoremodulev2", "aspnetcore", "ancm.mof");
+            var process = Process.Start("mofcomp.exe", path);
```
pakrym: Why not do it as part of the build?

jkotalik: Will add it as part of build when verifying freb/etw events. Calling mofcomp multiple times is fine.

PR #1483, opened 2018-10-09
 Avoid double builds of dependencies in /t:PackageStressTestApp

natemcmaster:
This target was causing a double build of Microsoft.AspNetCore.Server.IISIntegration.csproj and resulted in a hash mismatch between the package and the final .dll on disk.


natemcmaster:
The RuntimeIdentifier property.

:-/

natemcmaster:
Also, it looks like BuildNumber didn't flow into the right build context, so the assembly attributes are slightly different.

image

Issue #1481, opened 2018-10-09
Test failure: ApplicationInitializationInitializedInProc


jkotalik:
```
[xUnit.net 00:00:34.38]     IIS.FunctionalTests.ApplicationInitializationTests.ApplicationInitializationInitializedInProc(hostingModel: OutOfProcess) [FAIL]
11:46:49   Failed   IIS.FunctionalTests.ApplicationInitializationTests.ApplicationInitializationInitializedInProc(hostingModel: OutOfProcess)
11:46:49   Error Message:
11:46:49    System.InvalidOperationException : Didn't get response that satisfies predicate after 10 retries
11:46:49   Stack Trace:
11:46:49      at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.Helpers.RetryRequestAsync(HttpClient client, String uri, Func`2 predicate) in /_/test/Common.FunctionalTests/Utilities/Helpers.cs:line 125
11:46:49      at IIS.FunctionalTests.ApplicationInitializationTests.ApplicationInitializationInitializedInProc(HostingModel hostingModel) in /_/test/IIS.FunctionalTests/ServicesTests.cs:line 43
11:46:49   --- End of stack trace from previous location where exception was thrown ---
11:46:49    | [4.447s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Information: Stopping site, state: Stopped
11:46:49    | [4.450s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Information: Stopping pool, state: Stopping
11:46:49    | [4.452s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Information: WorkerProcess not stopped yet
11:46:49    | [4.673s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Information: WorkerProcess not stopped yet
11:46:49    | [4.899s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Information: WorkerProcess not stopped yet
11:46:49    | [5.119s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Information: WorkerProcess not stopped yet
11:46:49    | [5.337s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISDeployer Information: WorkerProcess not stopped yet
[...]

Issue #1392, opened 2018-09-13
ANCM V2 - net stop was /y Issue


rockerinthelocker:
'net stop was /y' causes 'Failed to gracefully shutdown application' warnings being logged to the Application event log.

Is this a known issue?

jkotalik:
Is this with In-proc or out of proc ANCM? 

This isn't a known issue. Should be an easy repro. Thanks for noticing!

rockerinthelocker:
Sorry for not being specific. Yes, it's in-proc.

pakrym:
Can you share your `Program.Main` method?

rockerinthelocker:
```
    public class Program
    {
        public static void Main(string[] args)
        {
            CreateWebHostBuilder(args).Build().Run();
        }

        public static IWebHostBuilder CreateWebHostBuilder(string[] args) => WebHost.CreateDefaultBuilder(args).UseStartup<Startup>();
    }



    public class Startup
    {
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddMvcCore()
                .AddAuthorization()
                .AddRazorViewEngine()
                .SetCompatibilityVersion(CompatibilityVersion.Version_2_2);
        }

        public void Configure(IApplicationBuilder app)
        {
            app.UseStaticFiles();

            app.UseMvc(routes =>
            {
                routes.MapRoute(
                name: "default",
                template: "{controller=Home}/{action=Index}/{id?}");
            });
        }
    }

```

rockerinthelocker:
IIS application pool settings for the site running on Windows Server 2012 R2 with dotnet-hosting-2.2.0-preview1-35029-win installed:

```
            <add name="MySite" autoStart="true" enable32BitAppOnWin64="false" managedRuntimeVersion="" managedPipelineMode="Integrated" startMode="AlwaysRunning">
                <processModel identityType="SpecificUser" userName="User" password="Pass" idleTimeout="00:00:00" />
            </add>
```

rockerinthelocker:
PS: It's a test server and the test site (just serving static files and 'hello, word!') was idle for hours at the time net stop was executed.

rockerinthelocker:
Just tested with dotnet-hosting-2.2.0-preview2-35157-win and the same warning message got logged. Interestingly, shutting down the worker process takes even longer than with preview 1.

```
C:\Windows\system32>net stop was /y
The following services are dependent on the Windows Process Activation Service service.
[...]

PR #1484, opened 2018-10-09
Fix applicationInitialization tests and disconnect handler

pakrym:
Should fix https://github.com/aspnet/IISIntegration/issues/1481

Might solve https://github.com/aspnet/IISIntegration/issues/1392

`pHttpContext->GetConnection()` could return null in case of `applicationInitialization` request.

Also verify that expected things happen both when we have `applicationInitialization->add` and without it.


pakrym:
I think I'll merge this. VSTS failure is being fixed by https://github.com/aspnet/IISIntegration/pull/1489

PR #1485, opened 2018-10-10
Reenable client disconnect tests

pakrym:
Pipelines package was updated in https://github.com/aspnet/IISIntegration/pull/1483


pakrym:
VSTS https://github.com/aspnet/IISIntegration/issues/1463

Issue #1487, opened 2018-10-10
Flaky Test: CheckFrebDisconnect


jkotalik:
```
 Failed   Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.FrebTests.CheckFrebDisconnect
2018-10-09T23:44:57.5718697Z   Error Message:
2018-10-09T23:44:57.5719409Z    The collection contained 0 matching element(s) instead of 1.
2018-10-09T23:44:57.5728543Z   Stack Trace:
2018-10-09T23:44:57.5728777Z      at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.FrebTests.AssertFrebLogs(IISDeploymentResult result, String[] data) in /_/test/Common.FunctionalTests/Inprocess/FrebTests.cs:line 111
2018-10-09T23:44:57.5731665Z      at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.FrebTests.CheckFrebDisconnect() in /_/test/Common.FunctionalTests/Inprocess/FrebTests.cs:line 86
2018-10-09T23:44:57.5732398Z   --- End of stack trace from previous location where exception was thrown ---
2018-10-09T23:44:57.5733627Z   Standard Output Messages:
2018-10-09T23:44:57.5733961Z    | [0.000s] TestLifetime Information: Starting test CheckFrebDisconnect at 2018-10-09T23:44:49
2018-10-09T23:44:57.5734237Z    | [0.003s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISExpressDeployer Information: Deploying [Variation] :: ServerType=IISExpress, Runtime=CoreClr, Arch=x64, BaseUrlHint=, Publish=True

2018-10-09T23:44:57.5752100Z    | [0.040s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISExpressDeployer Information: Attempting to start IIS Express on port: 2634
2018-10-09T23:44:57.5752261Z    | [0.042s] Microsoft.AspNetCore.Server.IntegrationTesting.IIS.IISExpressDeployer Debug: Saving Config to C:\Users\VssAdministrator\AppData\Local\Temp\tmp78C4.tmp
[...]

Issue #1447, opened 2018-09-27
TestServer stopping dotnet test process.


jkotalik:
We have had this issue before. Will investigate.
```
 [Step 6/7]          IISIntegration        |   Starting test execution, please wait...
[18:57:10] :	 [Step 6/7]          IISIntegration        |   [xUnit.net 00:00:11.21]     Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReaderThrowsCancelledException [SKIP]
[18:57:10] :	 [Step 6/7]          IISIntegration        |   [xUnit.net 00:00:11.21]     Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.WriterThrowsCancelledException [SKIP]
[18:57:11] :	 [Step 6/7]          IISIntegration        |   Skipped  Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReaderThrowsCancelledException
[18:57:11] :	 [Step 6/7]          IISIntegration        |   Skipped  Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.WriterThrowsCancelledException
[18:57:26] :	 [Step 6/7]          IISIntegration        |   [xUnit.net 00:00:27.14]     Microsoft.AspNetCore.Server.IIS.FunctionalTests.AppOfflineTests.AppOfflineDroppedWhileSiteFailedToStartInRequestHandler_SiteStops_InProcess [SKIP]
[18:57:27] :	 [Step 6/7]          IISIntegration        |   Skipped  Microsoft.AspNetCore.Server.IIS.FunctionalTests.AppOfflineTests.AppOfflineDroppedWhileSiteFailedToStartInRequestHandler_SiteStops_InProcess
[18:57:40] :	 [Step 6/7]          IISIntegration        |   The active test run was aborted. Reason: 
[19:05:07] :	 [Step 6/7]          IISIntegration        |   Results File: C:\BuildAgent\work\33bdfc1cae7b2a38\modules\IISIntegration\artifacts\logs\UnitTests-netcoreapp2.2-10525.trx
[19:05:07] :	 [Step 6/7]          IISIntegration        |   
[19:05:07] :	 [Step 6/7]          IISIntegration        |   Total tests: Unknown. Passed: 298. Failed: 0. Skipped: 3.
[...]

PR #1488, opened 2018-10-10
Fix flaky ClientDisconnect freb tests.

jkotalik:
For https://github.com/aspnet/IISIntegration/issues/1487


test/Common.FunctionalTests/Inprocess/FrebTests.cs
```diff
@@ -96,23 +110,32 @@ private async Task<IISDeploymentResult> SetupFrebApp()
             return result;
         }
 
-        private void AssertFrebLogs(IISDeploymentResult result, params string[] data)
+        private void AssertSingleFrebLog(IISDeploymentResult result, params string[] data)
```
pakrym: Load all logs into an object structure and assert against it.

pakrym: Add a class FrebLog with FrebLogItem[] inside. Then you can use Assert.Contains and friends.

jkotalik: Create a FrebLogItem class, but didn't add a wrapper class.

jkotalik:
Test failure: #1447

pakrym (approved):

jkotalik:
Test failure #1447

PR #1496, opened 2018-10-11
public const string AuthenticationScheme

pakrym:
Related: https://github.com/aspnet/IISIntegration/issues/1493

Make `IISServerDefaults.AuthenticationScheme` const


src/Microsoft.AspNetCore.Server.IIS/IISServerDefaults.cs
```diff
@@ -5,7 +5,7 @@ namespace Microsoft.AspNetCore.Server.IIS
 {
     public class IISServerDefaults
     {
-        public static readonly string AuthenticationScheme = "Windows";
+        public const string AuthenticationScheme = "Windows";
         public const string Negotiate = "Negotiate";
```
Tratcher: These other values aren't used anywhere are they? Remove or make internal?

PR #1499, opened 2018-10-11
Include mof file in ANCMV2 package

jkotalik:
We are moving away from using a zip file to build ANCM and instead using the nuget packages. We realized with 2.1 that having an artifact dependency on ANCM breaks with patches. In 2.2 and 3.0, we will use an artifact dependency on ANCM if built from CI or pull from the dotnetfeed.

Therefore, we need to include the mof file 😄


natemcmaster:
I don't know what a .mof file is, but I think it needs to be code signed. It appears in the list of signable file types: https://microsoft.sharepoint.com/teams/codesigninfo/Wiki/Signable%20Files.aspx

jkotalik:
It wasn't signed before, good catch. @pan-wang is this something you do/need to do for other IIS-OOB modules?

jkotalik:
After investigation, mof files are plain text. We shouldn't need to sign them.

Issue #1439, opened 2018-09-25
ANCM V2 - Connection Abort Handling Issue


rockerinthelocker:
While ANCM V1 properly handles Http​Context.​Abort() and the likes..., ANCM V2 fails and connections are not aborted.

pakrym:
What's the behaviour you expect vs seeing?

rockerinthelocker:
Expected:
Connection abort (RST/ACK).

Actual:
In-process returns empty body, out-of-process throws.

pakrym:
Can you provide the app func you are using? In my tests I see resets or 502s.

rockerinthelocker:
Just to make sure we're on the same page. This is about the latest bits shipped, which is v2.2.0-preview2 at this time.

Nonetheless, ANCM v2 in-process/out-of-process must reset the connection on abort. ANCM v1 handles it properly for out-of-process and so should ANCM v2 for both, in-process and out-of-process.

```
public class Startup
{
	public Startup(IHostingEnvironment env)
	{
		HostingEnvironment = env;
	}

	public void ConfigureServices(IServiceCollection services)
	{
		services.AddMvcCore()
			.AddAuthorization()
			.AddRazorPages()
			.AddRazorViewEngine()
			.SetCompatibilityVersion(CompatibilityVersion.Version_2_2);
	}
	public void Configure(IApplicationBuilder app)
	{
		app.Use((context, next) =>
		{
			context.Abort();
			return next();
		});
		app.UseMvc();
	}
}
```

pakrym:
I can't repro the issue using the code you provided. Out-of-proc reliably serves `HTTP Error 502.3 - Bad Gateway`. Let's see if issue reproes for you after preview3 update.

rockerinthelocker:
With 2.2.0-preview3-35497, in-process works, out-of-process still fails to reset the connection and returns

```502 - Web server received an invalid response while acting as a gateway or proxy server.
There is a problem with the page you are looking for, and it cannot be displayed. When the Web server (while acting as a gateway or proxy) contacted the upstream content server, it received an invalid response from the content server.```

instead. Since a connection reset is expected (and works in ANCM v1), what's the point of breaking things and serving a 502 instead?

pakrym:
Serving 502 is not breaking, it's a valid response in the situation when the response wasn't started by IIS yet.

rockerinthelocker:
[...]

PR #1510, opened 2018-10-12
Add server abort tests

pakrym:
As part of https://github.com/aspnet/IISIntegration/issues/1439

I did a couple things in addition to abort tests
1. Added a pattern for writing tests that run on inproc/outofproc/outofprocv1 without deploying - we should switch to using it in tests that do not require server restarts (HelloWorld, etc.).
2. Fixed `launchSettings.json` for IIS Express
3. Added environment variable (ANCM_OUTOFPROCESS_HANDLER) for shortcircuiting out of process handler resolution and used it in launchSettings.json to point to projects build outputs and avoid copying and arranging dlls.


jkotalik:
> Added environment variable (ANCM_OUTOFPROCESS_HANDLER) for shortcircuiting out of process handler resolution and used it in launchSettings.json to point to projects build outputs and avoid copying and arranging dlls.

Confused why we should add this. Is this a feature addition or something to help with testing?

pakrym:
> Confused why we should add this. Is this a feature addition or something to help with testing?

Helps with testing, for F5 experience we don't need to copy files around and arrange them into versioned subfolders, they get used directly from project build output paths.

test/IISExpress.FunctionalTests/BasicAuthTests.cs
```diff
@@ -15,6 +15,50 @@
 
 namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
 {
+    public abstract class ServerAbortTests: FixtureLoggedTest
```
jkotalik: nit: file is named BasicAuthTests.

pakrym:
And by F5 experience I mean running our test sites from VS by pressing F5.

test/IISExpress.FunctionalTests/BasicAuthTests.cs
```diff
@@ -15,6 +15,50 @@
...
+            _fixture = fixture;
+        }
+
+        [ConditionalFact]
+        public async Task ClosesConnectionOnServerAbort()
+        {
+            try
+            {
+                var response = await _fixture.Client.GetAsync("/Abort").DefaultTimeout();
+                Assert.Equal(HttpStatusCode.BadGateway, response.StatusCode);
```
jkotalik: Isn't BadGateway wrong for inproc?

test/IISExpress.FunctionalTests/BasicAuthTests.cs
pakrym: Right, I'll move it out.

test/IISExpress.FunctionalTests/BasicAuthTests.cs
pakrym: I'll put it under if condition.

test/Common.FunctionalTests/ServerAbortTests.cs
```diff
@@ -0,0 +1,65 @@
...
+                    Assert.Equal(HttpStatusCode.BadGateway, response.StatusCode);
+                    // 0x80072f78 ERROR_HTTP_INVALID_SERVER_RESPONSE The server returned an invalid or unrecognized response
+                    Assert.Contains("0x80072f78", await response.Content.ReadAsStringAsync());
+                }
+                else
+                {
+                    Assert.True(false, "Should not reach here");
+                }
+            }
+            catch (HttpRequestException)
```
jkotalik: No way to Assert the status code for inproc? Is the status code set?

pakrym: No, it's connection reset, that's why you get an exception.

jkotalik: I see now. Personally I don't like this test as you don't know which variation will lead which response type. Ex Inproc and ANCMV1 will do connection reset while V2 will return 502.3. Consider splitting into two tests.

pakrym: Both connection reset and 502 are valid for V1 and V2 outofproc.

jkotalik: What do you get for inproc?

pakrym: Connection reset.

jkotalik: Can you add a comment about what are the expected results for each input?

pakrym: Will do

PR #1511, opened 2018-10-12
Enable auth tests on IIS

pakrym:
They were IIS Express only.


test/Common.FunctionalTests/Utilities/RequiresEnvironmentVariableAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+        private readonly string _name;
+
+        public RequiresEnvironmentVariableAttribute(string name)
+        {
+            _name = name;
+        }
+
+        public bool IsMet => !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(_name));
+
+        public string SkipReason => "Test cannot be run in Debug mode.";
```
jkotalik: Update skip reason you copy pasta demon.

Issue #1523, opened 2018-10-16
In some cases we compress/decompress/compress again, heating up CPUs


muratg:
We're doing unnecessary work with compression.

jkotalik:
This would be modifying WinHttp to not decompress when receiving the response.

pakrym:
Wasn't able to reproduce it. See #1525.

PR #1525, opened 2018-10-17
Add compression pass through tests

pakrym:
@Tratcher, do you think is this a valid test for https://github.com/aspnet/IISIntegration/issues/1523 ?


Tratcher:
No 
1. Pull in the actual response compression middleware (the repo layering allows it).
1. Disable response compression in IIS
1. Disable response decompression on the client, but set the Accept-encoding header manually
1. Set a compressible content-type (text/plain), write data
1. Read the response. Make sure the Content-Encoding header is set. Make sure the content is compressed.

pakrym:
> Also need to disable compression in IIS.

Disabled by default in deployers.

pakrym:
@Tratcher  ping

pakrym:
Failures are known and are being addressed

Issue #1458, opened 2018-09-29
Test failure: CheckUTF8File


aspnet-hello:
This issue was made automatically. If there is a problem contact ryanbrandenburg.

This test [failed](http://aspnetci/viewLog.html?buildId=564733&buildTypeId=XPlat_Windows_Win10Universe) with the following error:

```
System.AggregateException : One or more errors occurred. (The process cannot access the file 'C:\BuildAgent\temp\buildTmp\59dd3ec8-4640-4eee-ab37-8e87884b0a85\彡⾔\std_20180928221132_18624.log' because it is being used by another process.) (The process cannot access the file 'std_20180928221132_18624.log' because it is being used by another process.)
---- System.IO.IOException : The process cannot access the file 'C:\BuildAgent\temp\buildTmp\59dd3ec8-4640-4eee-ab37-8e87884b0a85\彡⾔\std_20180928221132_18624.log' because it is being used by another process.
---- System.IO.IOException : The process cannot access the file 'std_20180928221132_18624.log' because it is being used by another process.
----- Inner Stack Trace #1 (System.IO.IOException) -----
   at System.IO.FileStream.ValidateFileHandle(SafeFileHandle fileHandle)
   at System.IO.FileStream.CreateFileOpenHandle(FileMode mode, FileShare share, FileOptions options)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at System.IO.StreamReader..ctor(String path, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize)
   at System.IO.File.InternalReadAllText(String path, Encoding encoding)
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.LoggingTests.CheckUTF8File(TestVariant variant) in /_/test/Common.FunctionalTests/LogFileTests.cs:line 205
--- End of stack trace from previous location where exception was thrown ---
----- Inner Stack Trace #2 (System.IO.IOException) -----
   at System.IO.FileSystem.RemoveDirectoryRecursive(String fullPath, WIN32_FIND_DATA& findData, Boolean topLevel)
   at System.IO.FileSystem.RemoveDirectory(String fullPath, Boolean recursive)
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.LoggingTests.Dispose() in /_/test/Common.FunctionalTests/LogFileTests.cs:line 34
[...]

Issue #1526, opened 2018-10-17
Flaky test: 


jkotalik:
https://ci3.dot.net/job/aspnet_IISIntegration/job/release_2.2/job/windows-Configuration_Debug_prtest/1034/console

```
09:14:10   
09:14:10   Starting test execution, please wait...
09:14:13   [xUnit.net 00:00:00.55]     Microsoft.AspNetCore.Server.IIS.FunctionalTests.AppOfflineTests.AppOfflineDroppedWhileSiteFailedToStartInRequestHandler_SiteStops_InProcess [SKIP]
09:14:14   Skipped  Microsoft.AspNetCore.Server.IIS.FunctionalTests.AppOfflineTests.AppOfflineDroppedWhileSiteFailedToStartInRequestHandler_SiteStops_InProcess
09:14:26   [xUnit.net 00:00:13.20]     Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReaderThrowsCancelledException [FAIL]
09:14:26   Failed   Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReaderThrowsCancelledException
09:14:26   Error Message:
09:14:26    Assert.DoesNotContain() Failure
09:14:26   Found:    (filter expression)
09:14:26   In value: ConcurrentQueue<WriteContext> [WriteContext { EventId = 0, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Server.IISIntegration.Functio"..., LogLevel = Debug, ... }, WriteContext { EventId = 3, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Hosting.Internal.WebHost", LogLevel = Debug, ... }, WriteContext { EventId = 4, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Hosting.Internal.WebHost", LogLevel = Debug, ... }, WriteContext { EventId = 0, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Hosting.Internal.WebHost", LogLevel = Debug, ... }, WriteContext { EventId = 1, Exception = null, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Hosting.Internal.WebHost", LogLevel = Information, ... }, ...]
09:14:26   Stack Trace:
[...]

PR #1530, opened 2018-10-17
[WIP] Check what current processes are running to see if dotnet.exe/iisexpress.exe is being stopped.

jkotalik:
Trying to find the root cause of https://github.com/aspnet/IISIntegration/issues/1458. For now, when we can't read a file, we should try to what dotnet process was currently running. The test failure before was most likely due to a process locking the files from being deleted, we should make sure the process is actually dotnet.

I also added w3wp and iisexpress for completeness.


test/Common.FunctionalTests/Utilities/Helpers.cs
```diff
@@ -209,5 +210,29 @@ public static void AllowNoLogs(this IISDeploymentResult deploymentResult)
...
+        public static string ReadAllTextFromFile(string filename, ILogger logger)
+        {
+            try
+            {
+                return File.ReadAllText(filename);
+            }
+            catch (Exception ex)
+            {
+                // check if there is a dotnet.exe, iisexpress.exe, or w3wp.exe processes still running.
+                var hostingProcesses = Process.GetProcesses("dotnet").Where(proc =>
```
pakrym: This is a strange condition. `Process.GetProcesses("dotnet")` and then filtering by name.

jkotalik: I had that first, but I wanted to get w3wp/iisexpress too. It probably isn't necessary.

jkotalik:
Test failures:
ReaderThrowsCancelledException - https://github.com/aspnet/IISIntegration/issues/1526
Test run aborted - #1447

test/Common.FunctionalTests/Utilities/Helpers.cs
```diff
@@ -209,5 +210,26 @@ public static void AllowNoLogs(this IISDeploymentResult deploymentResult)
...
+        public static string ReadAllTextFromFile(string filename, ILogger logger)
+        {
+            try
+            {
+                return File.ReadAllText(filename);
+            }
+            catch (Exception ex)
+            {
+                // check if there is a dotnet.exe, iisexpress.exe, or w3wp.exe processes still running.
+                var hostingProcesses = Process.GetProcessesByName("dotnet");
```
pakrym: You use this helper in in-process tests, it needs to support w3wp and iisexpress.

test/Common.FunctionalTests/LogFileTests.cs
```diff
@@ -13,32 +13,21 @@
...
-            {
-                Directory.Delete(_logFolderPath, true);
-            }
-        }
-
         public static TestMatrix TestVariants
             => TestMatrix.ForServers(DeployerSelector.ServerType)
                 .WithTfms(Tfm.NetCoreApp22)
-                .WithAllApplicationTypes()
+                .WithApplicationTypes(ApplicationType.Standalone)
```
pakrym: Why narrow the matrix?

PR #1533, opened 2018-10-18
Fix flaky disconnect test

pakrym:
https://github.com/aspnet/IISIntegration/issues/1526


pakrym:
Failure is successful VSTS build timing out.

Issue #1522, opened 2018-10-16
Change 2.1-oob build to be feature flag of 2.2 branch


pakrym:
This is required for mono-repo efforts.

pakrym:
https://dnceng.visualstudio.com/internal/internal%20Team/_build?definitionId=190&_a=summary

PR #1534, opened 2018-10-18
Add support for building 2.1 based Server.IIS package

pakrym:
https://github.com/aspnet/IISIntegration/issues/1522

/cc @natemcmaster


jkotalik:
Do we care that tests are publishing for 2.2 instead of 2.1?

jkotalik:
Also is this using the 2.1 build in tests?

pakrym:
> Do we care that tests are publishing for 2.2 instead of 2.1?

We'll just check when giving the build away.

> Also is this using the 2.1 build in tests?

It uses the project reference so yes.

PR #1536, opened 2018-10-18
Shorten log file names

pakrym:
To avoid VSTS failures


pakrym:
Doesn't help VSTS isssue but nice to have anyway. Tracking https://dnceng.visualstudio.com/internal/_workitems/edit/100

PR #1539, opened 2018-10-19
Fix two string allocation issues

jkotalik:
Found when verifying.

- Resize char* when allocating std::string with returned size.
- Resize StringHelpers::format after allocating to make null character not part of size of string.


src/AspNetCoreModuleV2/CommonLib/ServerErrorHandler.h
```diff
@@ -63,13 +63,13 @@ class ServerErrorHandler : public REQUEST_HANDLER
...
             HRSRC rc = nullptr;
             HGLOBAL rcData = nullptr;
-            const char* data = nullptr;
+            std::string data;
 
             THROW_LAST_ERROR_IF_NULL(rc = FindResource(module, MAKEINTRESOURCE(page), RT_HTML));
             THROW_LAST_ERROR_IF_NULL(rcData = LoadResource(module, rc));
             auto const size = SizeofResource(module, rc);
             THROW_LAST_ERROR_IF(size == 0);
-            THROW_LAST_ERROR_IF_NULL(data = static_cast<const char*>(LockResource(rcData)));
```
pakrym: Bring the null check back.

src/AspNetCoreModuleV2/CommonLib/ServerErrorHandler.h
```diff
@@ -51,7 +51,7 @@ class ServerErrorHandler : public REQUEST_HANDLER
         dataChunk.DataChunkType = HttpDataChunkFromMemory;
 
         dataChunk.FromMemory.pBuffer = page.data();
-        dataChunk.FromMemory.BufferLength = static_cast<ULONG>(page.size());
+        dataChunk.FromMemory.BufferLength = static_cast<ULONG>(page.size());
```
pakrym: revert

src/AspNetCoreModuleV2/CommonLib/StringHelpers.h
```diff
@@ -27,6 +27,7 @@ std::wstring format(const std::wstring& format, Args ... args)
         {
             throw std::system_error(std::error_code(errno, std::system_category()));
         }
+        result.resize(size);
```
pakrym: Do you need the initial resize to `size + 1`?

test/Common.FunctionalTests/Inprocess/ErrorPagesTests.cs
```diff
@@ -24,37 +24,49 @@ public ErrorPagesTests(PublishedSitesFixture fixture)
...
             var deploymentParameters = _fixture.GetBaseDeploymentParameters(HostingModel.OutOfProcess, publish: true);
             var response = await DeployAppWithStartupFailure(deploymentParameters);
 
             Assert.Equal(HttpStatusCode.BadGateway, response.StatusCode);
 
             StopServer();
+            var responseString = await response.Content.ReadAsStringAsync();
+            Assert.Contains("HTTP Error 502.5 - ANCM Out-Of-Process Startup Failure", responseString);
+
+            if (!DeployerSelector.IsBackwardsCompatiblityTest)
```
pakrym: Extract a method.

test/IIS.FunctionalTests/DeployerSelector.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Server.IntegrationTesting;
+
+namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
+{
+    public static class DeployerSelector
+    {
+        public static ServerType ServerType => ServerType.IIS;
+        public static bool IsBackwardsCompatiblityTest => false;
```
pakrym: Why not in RequiresNewShimAttribute?

src/AspNetCoreModuleV2/CommonLib/StringHelpers.h
jkotalik: Yeah. The null terminator is still written to the string.

src/AspNetCoreModuleV2/CommonLib/ServerErrorHandler.h
```diff
@@ -63,13 +63,15 @@ class ServerErrorHandler : public REQUEST_HANDLER
...
+            std::string data;
+            const char* pTempData;
 
             THROW_LAST_ERROR_IF_NULL(rc = FindResource(module, MAKEINTRESOURCE(page), RT_HTML));
             THROW_LAST_ERROR_IF_NULL(rcData = LoadResource(module, rc));
             auto const size = SizeofResource(module, rc);
             THROW_LAST_ERROR_IF(size == 0);
-            THROW_LAST_ERROR_IF_NULL(data = static_cast<const char*>(LockResource(rcData)));
+            THROW_LAST_ERROR_IF_NULL(pTempData = static_cast<const char*>(LockResource(rcData)));
+            data = std::string(pTempData, size);
```
pakrym: Do you have to use size here?

test/IIS.FunctionalTests/DeployerSelector.cs
jkotalik: We already used DeployerSelector for specifying whether a test was IIS or IISExpress (for statics). Seemed better to use this instead.

src/AspNetCoreModuleV2/CommonLib/StringHelpers.h
jkotalik: Let's say we have the string `"Hello %s"` with `args="World"`. The size of the resulting string is 11 characters not including the null terminator. Next, we call `const size_t size = swprintf(nullptr, 0, format.c_str(), args ...);` which would return 11. We resize result to 11. Now, on the second call to snprintf, we need to pass in n+1 for the size of the new buffer because "The generated string has a length of at most n-1, leaving space for the additional terminating null character." This is fine to do as std::string will always have room for one more character.

So yes, you are right in that we can use size(), but we weren't passing the right thing into snprintf.

jkotalik:
Running locally to verify as webhooks are down.

PR #1553, opened 2018-10-23
Fix upgrade deps and run upgrade deps

jkotalik:
Packages need to be in the Package Version: Pinned section.


jkotalik:
Test failure: #1447

natemcmaster (approved):

Issue #1494, opened 2018-10-11
Test failure: AppOfflineDroppedWhileSiteStarting_SiteShutsDown_InProcess under App Verifier


pakrym:
I think the failure is caused by a race between stopping stdout redirect and debugutil logging code.

PR #1554, opened 2018-10-24
Fix test server issues.

jkotalik:
Seems like delaying in dispose for a bit stops issues in hostable web core. Ran 100 times locally without failures. Unfortunate 😢


jkotalik:
Closing and reopening for more coverage.

jkotalik:
Test failure: https://github.com/aspnet/IISIntegration/issues/1494

pakrym:
Does it work without sleep?

jkotalik:
From what I can tell no. I can try reducing the sleep timeout, but I couldn’t find any other combination that worked (without a different approach).

pakrym:
Rerun PR multiple times please.

pakrym:
Add a comment

Issue #1561, opened 2018-10-25
Publish to azure initial request returns "The service is unavailable"


jkotalik:
Repro:
1. Publish site to azure using inprocess
2. Initial request is sent to site automatically
3. Check response

Expected: Response is the home page of the site
Actual: ~50% of the time, it returns "The service is unavailable". Refreshing the page will return the home page.

rockerinthelocker:
Maybe servers are oversold? There are certainly more reliable Web hosting providers out there having ANCM 2.2 preview installed for testing.

rockerinthelocker:
What's the point of recycling the worker process before shutting down anyway? If publishing fails, it's most likely due to file locks and the site (not the worker process) must be restarted in IIS to remove file locks.

rockerinthelocker:
(But maybe I misread the issue.)

PR #1564, opened 2018-10-26
[WIP] Call recycle worker process significantly earlier

jkotalik:
Ideas for #1561 . Still testing validity.


natemcmaster:
@jkotalik if this is still necessary, reopen on aspnet/AspNetCore.

jkotalik:
Will do

